Issue #1, opened 2016-11-23
Non-negative tensor decompositions by NNLS


ahwillia:
I haven't had good luck with multiplicative update algorithms. I'd like to leverage the code here - https://github.com/kimjingu/nonnegfac-python - to fit non-negative CP by alternating nonnegative least squares.

I'm happy to open a PR if you'd like to include this. The only tricky bit is how to cite/repurpose that code. I think we can add it to the `requirements.txt` as follows:

```
-e https://github.com/kimjingu/nonnegfac-python.git#egg=nonnegfac
```

cc @kimjingu

PR #2, opened 2016-11-24
Add nonneg cp decomposition

ahwillia:
Here is a prototype for what I mentioned here: https://github.com/tensorly/tensorly/issues/1

It could probably use a bit more documentation - but would love feedback at this stage. I added some short tests/checks in `examples/decomposition/`

If you like it I'm happy to edit and clean it up as you see fit.

### DIFF
--- a/doc/templates/layout.html
+++ b/doc/templates/layout.html
@@ -3,6 +3,10 @@
 {% block footer %}
     {{ super() }}
 
+	{# https://github.com/blog/273-github-ribbons #}
+	<a href="https://github.com/tensorly/tensorly"><img style="position: absolute; top: 0; right: 0; border: 0;" src="https://camo.githubusercontent.com/a6677b08c955af8400f44c6298f40e7d19cc5b2d/68747470733a2f2f73332e616d617a6f6e6177732e636f6d2f6769746875622f726962626f6e732f666f726b6d655f72696768745f677261795f3664366436642e706e67" alt="Fork me on GitHub" data-canonical-src="https://s3.amazonaws.com/github/ribbons/forkme_right_gray_6d6d6d.png">
+	</a>
+
     <script>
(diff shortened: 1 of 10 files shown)


JeanKossaifi:
Thank you for your contribution, it would be interesting to have more decomposition methods.

When creating this project we wanted to keep dependencies to a minimum (NumPy and SciPy only, with a soft-dependency on Matplotlib) so we would prefer an alternative that doesn't involve extra-dependencies.

ahwillia:
I really only need this file - https://github.com/kimjingu/nonnegfac-python/blob/master/nonnegfac/nnls.py

I'm going to copy it over to tensorly and paste the [license](https://github.com/kimjingu/nonnegfac-python/blob/master/LICENSE) at the top. Is this okay @kimjingu?

kimjingu:
@ahwillia Fine with myself, but I'd leave the interpretation of the license up to you. A down side of copying is that the file can be left out from any future bug fixes. If you do, I'd recommend adding a comment with information of the original file and the citation info. 

Jingu Kim and Haesun Park. Fast Nonnegative Matrix Factorization: An Active-set-like Method And Comparisons. SIAM Journal on Scientific Computing (SISC), 33(6), pp. 3261-3281, 2011.

Thanks for your interest.

JeanKossaifi:
Whilst I see the point, we would rather add tailored contributions as opposed to copying raw code from other projects (code rot, not updated as mentioned by Kim, etc). In the latter case it seems more appropriate to directly call the other library. 

You can still easily use this in your work and we could potentially modify the code to allow adding other solvers to make this easier.

Feel free to propose ideas in the mailing list https://groups.google.com/forum/#!forum/tensorly.

nipunbatra:
In case using nnls.py from nonneg fac was a deal-breaker, then I wanted to chime in that scipy has an NNLS implementation. Here's a [small post](https://nipunbatra.github.io/blog/2017/nmf-nnls.html) I wrote for NNMF using NNLS.

Issue #8, opened 2017-08-30
Errors happened when resetting backend


JianboGuo:
I used the following official code to set backend for tensorly lib:
`tl.set_backend('numpy')`
and suffered from errors:
`AttributeError: 'module' object has no attribute 'reload'
`
Actually I find importlib doesn't have reload function. How can I solve this problem? I would appreciate you kind reply.

JianboGuo:
@animakumar  @JeanKossaifi I am using Python 2.7+ and I found it's because there was no 'reload' func in importlib module in this series of versions. However, I used another method mentioned in your docs
`# Set the default backend
default_backend = 'numpy'`
and suffered from
`Using numpy backend.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/Jianbo/Programs/anaconda2/lib/python2.7/site-packages/tensorly/__init__.py", line 28, in <module>
    from .backend import *
  File "/Users/Jianbo/Programs/anaconda2/lib/python2.7/site-packages/tensorly/backend/__init__.py", line 9, in <module>
    from .numpy_backend import *
  File "/Users/Jianbo/Programs/anaconda2/lib/python2.7/site-packages/tensorly/backend/numpy_backend.py", line 5, in <module>
    from numpy import testing
  File "/Users/Jianbo/Programs/anaconda2/lib/python2.7/site-packages/numpy/__init__.py", line 142, in <module>
    from . import add_newdocs
  File "/Users/Jianbo/Programs/anaconda2/lib/python2.7/site-packages/numpy/add_newdocs.py", line 13, in <module>
    from numpy.lib import add_newdoc
  File "/Users/Jianbo/Programs/anaconda2/lib/python2.7/site-packages/numpy/lib/__init__.py", line 8, in <module>
    from .type_check import *
  File "/Users/Jianbo/Programs/anaconda2/lib/python2.7/site-packages/numpy/lib/type_check.py", line 11, in <module>
    import numpy.core.numeric as _nx
  File "/Users/Jianbo/Programs/anaconda2/lib/python2.7/site-packages/numpy/core/__init__.py", line 26, in <module>
    raise ImportError(msg)
ImportError: 
Importing the multiarray numpy extension module failed.  Most
likely you are trying to import a failed build of numpy.
If you're working with a numpy git repo, try `git clean -xdf` (removes all
files not under version control).  Otherwise reinstall numpy.`

[...]

PR #9, opened 2017-08-30
Remove non-ascii characters from decomposition docstrings

cswiercz:
The presence of non-ascii characters causes a SyntaxError to be raised on
import. This minor fix removes said characters.

### Example

```
>>> import tensorly
>>> from tensorly.decomposition import parafac
  File "/Users/csw/tensorly/tensorly/decomposition/candecomp_parafac.py", line 126
    pp 792–799, ICML, 2005
                          ^
SyntaxError: Non-ASCII character '\xe2' in file /Users/csw/tensorly/tensorly/decomposition/candecomp_parafac.py on line 127, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

### DIFF
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -152,7 +152,7 @@ def non_negative_tucker(tensor, ranks, n_iter_max=10, init='svd', tol=10e-5,
     .. [2] Yong-Deok Kim and Seungjin Choi,
        "Nonnegative tucker decomposition",
        IEEE Conference on Computer Vision and Pattern Recognition s(CVPR),
-       pp 1–8, 2007
+       pp 1-8, 2007
     """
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
Thanks for the fix!

Not sure how these characters got in there or why they didn't raise an issue on my machines/the test server!

cswiercz:
No problem. My guess is that Python3 accepts unicode docstrings but Python2 does not.

JeanKossaifi:
As a side note, TensorLy is Python 3 only.

cswiercz:
That's interesting. I've been experimenting with your package in Python2 and found that, as long as I didn't change the backend after importing, (re: Issue #8 ) I'm able to run `parafac` and other components fine. 

Is there are particular part of your package that you think is Python3 only?

PR #14, opened 2017-09-11
Add utilities for kruskal tensors

ahwillia:
So far I added my function for normalizing the factor matrices of a kruskal tensor.

The other function that I find quite useful is [`align_kruskal`](https://github.com/ahwillia/tensortools/blob/master/tensortools/kruskal.py#L69), which is analogous to the `score` function in Kolda's TensorToolbox in MATLAB.

Also, what do you think of adding the [`_validate_kruskal`](https://github.com/ahwillia/tensortools/blob/master/tensortools/kruskal.py#L191) function which basically checks that the user gave an appropriate input?

### DIFF
--- a/tensorly/kruskal_tensor.py
+++ b/tensorly/kruskal_tensor.py
@@ -93,3 +93,41 @@ def kruskal_to_vec(factors):
         vectorised tensor
     """
     return tensor_to_vec(kruskal_to_tensor(factors))
+
+def normalize_kruskal(factors):
+    """Normalizes factors to unit length and returns factor magnitudes
+        turns  ``factors = [|U_1, ... U_n|]`` into ``[lam; |V_1, ... V_n|]``,
+        where the norm of the columns of all `U_k` are absorbed into the
+        weights `lam` and so that the columns of all `V_k` have unit
+        Euclidean norm. In the special case that `factors` represents a 
+        symmetric tensor the weights `lam` correspond to higher-order
+        eigenvalues and the normalized factors `V_k` correspond to
+        higher-order eigenvectors.
+    Parameters
+    ----------
+    factors : ndarray list
+        list of matrices, all with the same number of columns
+        i.e.::
+            for u in U:
+                u[i].shape == (s_i, R)
+                
+        where `R` is fixed while `s_i` can vary with `i`
+    Returns
+    -------
+    normalized_factors : ndarray list
+        list of matrices with the same shape as `factors`
+    lam : ndarray
+        vector of length `R` holding weights
+    """
+
+    # allocate variables for weights, and normalized factors
+    lam = np.ones(rank)
+    V = []
+
+    # normalize columns of factor matrices
+    for fact in factors:
+        s = np.linalg.norm(fact, axis=0)
+        lam *= s
+        V.append(fact / np.where(s==0, 1, s))
+
+    return V, lam
--- a/tensorly/kruskal_tensor.py
+++ b/tensorly/kruskal_tensor.py
@@ -96,13 +96,12 @@ def kruskal_to_vec(factors):
 
 def normalize_kruskal(factors):
     """Normalizes factors to unit length and returns factor magnitudes
-        turns  ``factors = [|U_1, ... U_n|]`` into ``[lam; |V_1, ... V_n|]``,
-        where the norm of the columns of all `U_k` are absorbed into the
-        weights `lam` and so that the columns of all `V_k` have unit
-        Euclidean norm. In the special case that `factors` represents a 
-        symmetric tensor the weights `lam` correspond to higher-order
-        eigenvalues and the normalized factors `V_k` correspond to
-        higher-order eigenvectors.
+        turns  ``factors = [|U_1, ... U_n|]`` into ``[weights; |V_1, ... V_n|]``,
+        where the columns of each `V_k` are normalized to unit Euclidean length
+        from the columns of `U_k` with the normalizing constants absorbed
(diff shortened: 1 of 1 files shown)


tensorly/kruskal_tensor.py
```diff
@@ -93,3 +93,41 @@ def kruskal_to_vec(factors):
...
+                
+        where `R` is fixed while `s_i` can vary with `i`
+    Returns
+    -------
+    normalized_factors : ndarray list
+        list of matrices with the same shape as `factors`
+    lam : ndarray
+        vector of length `R` holding weights
+    """
+
```
JeanKossaifi: Need to define `rank`:
```python
rank = factors[0].shape[0]
```

tensorly/kruskal_tensor.py
```diff
@@ -93,3 +93,41 @@ def kruskal_to_vec(factors):
...
+    # allocate variables for weights, and normalized factors
+    lam = np.ones(rank)
```
JeanKossaifi: maybe we could call it `weight`or `scale`instead of lamb as we were discussing with @cswiercz on the other thread?

tensorly/kruskal_tensor.py
```diff
@@ -93,3 +93,41 @@ def kruskal_to_vec(factors):
...
+    V = []
+
+    # normalize columns of factor matrices
+    for fact in factors:
+        s = np.linalg.norm(fact, axis=0)
```
JeanKossaifi: Since the norm was moved to the backend:
```python
s = tl.norm(fact, axis=0)
```
though this requires the patch by @cswiercz

tensorly/kruskal_tensor.py
```diff
@@ -93,3 +93,41 @@ def kruskal_to_vec(factors):
...
+        lam *= s
+        V.append(fact / np.where(s==0, 1, s))
```
JeanKossaifi: Same, we can't use np.where here since the backend could be mxnet or pytorch..

JeanKossaifi:
The align_function seems rather useful! Curious what use cases you have for it?
As to _validate_kruskal it is useful as well, but since we mainly form kruskal tensor by decomposing a full tensor, not sure where a user would get a non-valide kruskal tensor? Let me know what you think :)

ahwillia:
I provisionally fixed those issues, though we'll have to implement `T.where` for the various backends (or prevent divide-by-zero a different way)

The align function is nice to see if you recover the same CP decomposition from different initializations (since optimization can get stuck in local min). It is somewhat analogous to the idea of [clustering stability](https://arxiv.org/abs/1007.1075), but for tensor decompositions.

I will wait on the norm function provided by #10 as you suggest.

tensorly/kruskal_tensor.py
```diff
@@ -93,3 +93,41 @@ def kruskal_to_vec(factors):
...
+    weights = T.ones(rank)
+    V = []
+
+    # normalize columns of factor matrices
+    for fact in factors:
+        s = T.norm(fact, axis=0)
+        weights *= s
+        V.append(fact / T.where(s==0, 1, s))
+
+    return V, lam
```
cswiercz: `lam` is not defined.

Also, for clarity I would recommend renaming `V` to `normalized_factors`. That way the code is self-documenting.

Nitpicking: rename `fact` to `factor`.

tensorly/kruskal_tensor.py
```diff
@@ -93,3 +93,41 @@ def kruskal_to_vec(factors):
...
+    # allocate variables for weights, and normalized factors
+    rank = factors[0].shape[0]
+    weights = T.ones(rank)
+    V = []
+
+    # normalize columns of factor matrices
+    for fact in factors:
+        s = T.norm(fact, axis=0)
+        weights *= s
+        V.append(fact / T.where(s==0, 1, s))
```
cswiercz: Maybe I have a different version of MXNet but shouldn't the second argument be `T.ones(s.size)`?

ahwillia: Don't know about MXNet but I believe that should work for numpy

tensorly/kruskal_tensor.py
ahwillia: Good catch. I need to add a test function, but am being lazy until #10 is resolved.

tensorly/kruskal_tensor.py
ahwillia: Edited to your version as that also works for numpy -- I am not testing against MXNet

ahwillia:
I have a bit of time now. Should I clean this up and merge? Or are we giving up on this?

JeanKossaifi:
Let's get this merged!

ahwillia:
What needs to be done?

JeanKossaifi:
Mainly we just need a test function.

Also, the `norm` function currently has no `axis` parameter (added by #10) so this will fail..
@cswiercz can you push your changes you made in #10 to `norm` and the added corresponding test as a new PR that we can merge quickly? If not we can simply add these changes here.

JeanKossaifi:
@ahwillia do you want to add these or should I do it?

ahwillia:
Is #10 merged yet? I'm a bit busy so if you want to add the test function
that would be great. Otherwise I'll get to it in the next few days.

ahwillia:
I believe this PR depends on #10, so I'm closing this as well due to inactivity. But let me know if you want to re-open.

cswiercz:
@ahwillia Some of the things I wanted to do in #10 (way back when) are being done in #29 's sub-prs. I unfortunately created #10 just as other things got really busy.

PR #24, opened 2018-01-13
Grab the version from the tensorly/__init__.py file

### LINKED ISSUE #25: Tensorly cannot be installed if the dependencies are not met beforehand.
## Problem
Tensorly cannot be installed if the dependencies are not met beforehand.
The `setup.py` script is importing tensorly to grab the version; this is problematic, since importing tensorly requires that all the dependencies are already met, and we are importing it before `setup(**config)` gets run.

This defeats the purpose of using `'install_requires': ['numpy', 'scipy']`

jesuscast:
## Problem
Tensorly cannot be installed if the dependencies are not met beforehand.
The `setup.py` script is importing tensorly to grab the version; this is problematic, since importing tensorly requires that all the dependencies are already met, and we are importing it before `setup(**config)` gets run.

This defeats the purpose of using `'install_requires': ['numpy', 'scipy']`
## Solution
Grab the version by parsing through `tensorly/__version__.py` using regex
This is the method used by other popular python libraries (e.g flask).

closes #25 

--------------
**Side note**: While fixing this I discovered we currently do not support numpy v.1.14.0, so I made issue #26 , and changed requirements.txt accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,8 @@ install:
     # Useful for debugging any issues with conda
     - conda info -a
     # Replace dep1 dep2 ... with your dependencies
-    - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy scipy
+    # Hardcoding numpy version temporarily, refer to issue #26
+    - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy==1.13.3 scipy
     - source activate test-environment
     - pip install coveralls pytest pytest-cov nose
     - if [[ "$BACKEND" == "mxnet" ]]; then
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-numpy
+numpy<1.14.0
 scipy
 pytest-cov
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,17 @@
+import re
+import ast
+
 try:
     from setuptools import setup, find_packages
 except ImportError:
     from distutils.core import setup, find_packages
 
-import tensorly
-version = tensorly.__version__
+# Grab the version
+_version_re = re.compile(r'__version__\s+=\s+(.*)')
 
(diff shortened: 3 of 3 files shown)


JeanKossaifi:
Thanks for opening the pull-request!

As mentioned in the issues, I'd rather not have a hard coded version for NumPy: can we keep only the changes to the `setup.py` file in this pull-request?

Perhaps we should just use the version mentioned in the [Python Packaging User Guide](https://packaging.python.org/guides/single-sourcing-package-version/): 

```python
here = os.path.abspath(os.path.dirname(__file__))

def read(*parts):
    with codecs.open(os.path.join(here, *parts), 'r') as fp:
        return fp.read()

def find_version(*file_paths):
    version_file = read(*file_paths)
    version_match = re.search(r"^__version__ = ['\"]([^'\"]*)['\"]",
                              version_file, re.M)
    if version_match:
        return version_match.group(1)
    raise RuntimeError("Unable to find version string.")

setup(
   ...
   version=find_version("package", "__init__.py")
   ...
)
```

JeanKossaifi:
@jesuscast want to make the changes?

jesuscast:
yup, sorry about my inactivtiy, I was busy with a couple of deadlines at work.

jesuscast:
If we do not hard code the version, the tests will just fail.

Travis CI will fetch the latest numpy version, and then the tests will fail because of #26

JeanKossaifi:
@jesuscast the issue is fixed by 3dbf6bdc8595a8fdc8cf5f3c08178f9c4b223e2f, we can remove the version :)
Can we also use the code mentioned in the Python Packaging User Guide to extract the version? Or do you have a specific reason to use AST?

JeanKossaifi:
Ping @jesuscast  :)

JeanKossaifi:
Fixed by 9f87b657392d28ac15fa5ed7421269bc5af66856.

PR #31, opened 2018-01-31
Add functionality to backend

cswiercz:
We add the QR decomposition function `qr()` to the backends as well as update
`norm()` to accept different axes. MXNet's `dot()` is updated to return a scalar
when the dimension of the output is 1x1 to match the behavior of the Numpy and
Pytorch backends.

### DIFF
--- a/tensorly/backend/mxnet_backend.py
+++ b/tensorly/backend/mxnet_backend.py
@@ -2,20 +2,21 @@
 Core tensor operations with MXnet.
 """
 
+import warnings
+
+import mxnet as mx
 import numpy
 import scipy.linalg
 import scipy.sparse.linalg
-from numpy import testing
-import mxnet as mx
-from mxnet import nd as nd
-from . import numpy_backend
 
+from numpy import testing
+from mxnet import nd as nd
 from mxnet.ndarray import arange, zeros, zeros_like, ones
-from mxnet.ndarray import reshape, moveaxis, dot, transpose
+from mxnet.ndarray import reshape, moveaxis, transpose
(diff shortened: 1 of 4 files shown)


cswiercz:
Handling the inconsistencies between how the different backends handle 1x1 tensors and scalars is tricky. For example,

```
>>> import numpy as np
>>> import mxnet as mx
>>> import torch
>>> np.reshape(1.0, (1,))
array([ 1.])
>>> mx.nd.reshape(1.0, (1,))
AssertionError
```

This came up because of the way the backends handle vector dot products:

```
>>> x = np.array([1,2,3])
>>> np.dot(x,x)
14
>>> y = mx.nd.array(x)
>>> mx.nd.dot(y,y)
[ 14. ]
<NDArray 1 @cpu(0)>
>>> z = torch.Tensor(x)
>>> torch.dot(z,z)
14.0
```

I'm int he process of wrapping the MXNet backend `dot()` function as well as the MXNet and Pytorch `reshape()` functions to handle scalars in a consistent manner. Wanted to hear your thoughts before I get too deep into the weeds.

*Side note - there are performance considerations on the MXNet side. I believe that `mx.ndarray.NDArray.asscalar()` is a blocking call. (That is, the MXNet engine needs to "pause" to convert the array into a Python object.) However, since casting to scalar is used often in conditionals...*

cswiercz:
My current approach will be to *not* cast 1x1 MXNet `dot()` output into scalars. However, I'll just have to find a different way to test if the inner product of two vectors is a given value. (Convert to Numpy is probably the best bet.)

cswiercz:
Accidentally deleted remote branch instead of overwriting it...

PR #29, opened 2018-01-24
[Tracking] Major Parafac Improvements

cswiercz:
> This is a tracking PR for a collection of improvements to `parafac()`. I wanted to get @JeanKossaifi 's input before I dump a bunch of code into this PR.

# Sub-PRs

- [x] #32 Add functionality to backend
- [x] #33 Add where() to Pytorch backend
- [x] #34 Allow optional weights in kruskal_to_tensor
- [x]  #37 Refactor `parafac()` initialization
- [x] *Add Orthogonalized ALS*
- [ ] *Add Line Search*

# Description

In a personal, private branch of Tensorly I have implemented several improvements to `parafac()`. Primarily,

* Orthogonalized ALS [1] - QR-orthogonalizing the factor matrices in the first few iterations can lead to better local minima in fewer iterations, especially if the input tensor has orthogonal components.
* Enhanched Line Search [2] - technically, I only implemented a basic line search algorithm. Experiments show that line search improves convergence rates as well as helps the algorithm escape local minima more quickly.

I would like to add this private work to Tensorly. This work requires a bit of a refactor of the existing `parafac()` code so that it is easy to turn these options on and off as well as makes the existing code more easily extendible. Below I included some arguments to the effectiveness of these "augmentations". Please let me know if this sounds good or if you have any concerns. I will be posting the actual code (and many many tests) in this PR soon.

# Experimental Data

The two plots below are typical examples of the reconstruction error improvement on random noisly orthogonal order three tensors. (The following code shows how these tensors are generated with `noisy = True`.) The figures below compare iteration count vs. reconstruction error between "vanilla" alternating least squares (ALS) and adding combinations of line search (LS) and QR orthogonalization (QR). The dashed black line indicates the L2-norm of the added noise which is the absolute best we can do in estimating the tensor as a sum of orthogonal components.

```python
def generate_random_tensor(dimensions=(10,10,10), rank=10, seed=0,
                           orthogonal=False, near_orthogonal=False, noisy=False):
    order = len(dimensions)
    dim = min(dimensions)
   if (rank > dim) and (orthogonal or near_orthogonal):
       raise ValueError('Can only construct orthogonal and near-orthogonal '
                        'tensors when rank <= min(dimensions)')

    np.random.seed(seed)
    weights = 4*tensorly.arange(1, rank+1).astype('float32')
    factors = map(tensorly.tensor, (np.random.randn(dim,rank) for dim in dimensions))

   if orthogonal:
[...]

PR #32, opened 2018-01-31
Add functionality to backend

cswiercz:
We add the QR decomposition function `qr()` and `prod()` function to the backends as well as update `norm()` to accept different axes. This PR is a pre-requisite to #29 .

# Tests

```bash
$ TENSORLY_BACKEND=numpy pytest -v ./tensorly
$ TENSORLY_BACKEND=mxnet pytest -v ./tensorly
$ TENSORLY_BACKEND=pytorch pytest -v ./tensorly
```

### DIFF
--- a/tensorly/backend/mxnet_backend.py
+++ b/tensorly/backend/mxnet_backend.py
@@ -2,20 +2,21 @@
 Core tensor operations with MXnet.
 """
 
+import warnings
+
+import mxnet as mx
 import numpy
 import scipy.linalg
 import scipy.sparse.linalg
-from numpy import testing
-import mxnet as mx
-from mxnet import nd as nd
-from . import numpy_backend
 
+from numpy import testing
+from mxnet import nd as nd
 from mxnet.ndarray import arange, zeros, zeros_like, ones
-from mxnet.ndarray import reshape, moveaxis, dot, transpose
+from mxnet.ndarray import moveaxis, dot, transpose, reshape
 from mxnet.ndarray import sqrt, abs, where, maximum, sign
 
+from . import numpy_backend
 
-#import numpy as nd
 # Author: Jean Kossaifi
 
 # License: BSD 3 clause
@@ -85,26 +86,36 @@ def min(tensor, *args, **kwargs):
 def max(tensor, *args, **kwargs):
     return nd.min(tensor, *args, **kwargs).asscalar()
 
-def norm(tensor, order):
+def norm(tensor, order=2, axis=None):
     """Computes the l-`order` norm of tensor
+
     Parameters
     ----------
     tensor : ndarray
     order : int
+    axis : int or tuple
+
     Returns
     -------
-    float
-        l-`order` norm of tensor
+    float or tensor
+        If `axis` is provided returns a tensor.
     """
+    # handle difference in default axis notation
+    if axis is None:
+        axis = ()
+
     if order == 'inf':
-        return nd.max(nd.abs(tensor)).asscalar()
+        return nd.max(nd.abs(tensor), axis=axis).asscalar()
     if order == 1:
-        res =  nd.sum(nd.abs(tensor))
+        res =  nd.sum(nd.abs(tensor), axis=axis)
     elif order == 2:
-        res = nd.sqrt(nd.sum(tensor**2))
+        res = nd.sqrt(nd.sum(tensor**2, axis=axis))
     else:
-        res = nd.sum(nd.abs(tensor)**order)**(1/order)
-    return res.asscalar()
+        res = nd.sum(nd.abs(tensor)**order, axis=axis)**(1/order)
+
+    if res.shape == (1,):
+        return res.asscalar()
+    return res
 
 def kr(matrices):
     """Khatri-Rao product of a list of matrices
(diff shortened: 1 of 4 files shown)


tensorly/backend/pytorch_backend.py
```diff
@@ -135,21 +138,29 @@ def solve(matrix1, matrix2):
...
-    float
-        l-`order` norm of tensor
+    float or tensor
+        If `axis` is provided returns a tensor.
     """
+    # pytorch does not accept `None` for any keyword arguments
+    if order and order != 'inf':
+        kwds['p'] = order
+    if axis is not None:
+        kwds['dim'] = axis
```
JeanKossaifi: I'm thinking for clarity it might be best to explicitly have `p=order` and `dim=axis` in the function calls rather than add them to the dict, especially since we call then call different functions with different parameters. Or at least add the values in the dict only in the relevant part of conditional branching. We could even leave completely the **kwargs out for now, since it is never actually used (if we need other params we can add them then),

tensorly/backend/mxnet_backend.py
```diff
@@ -207,6 +218,17 @@ def partial_svd(matrix, n_eigenvecs=None):
         U, S, V = U[:, ::-1], S[::-1], V[:, ::-1]
         return tensor(U, **ctx), tensor(S, **ctx), tensor(V.T, **ctx)
 
+def qr(matrix, **kwds):
+    # NOTE - should be replaced with geqrf when available
+    try:
+        Q, R = map(nd.transpose, nd.linalg.gelqf(matrix.T))
```
JeanKossaifi: Personally I'd rather have 
```
Q, R = nd.linalg.gelqf(matrix.T)
return Q.T R.T
```
I find it makes what the code does more evident to the reader (generally I tend to not use map unless I have to)..

tensorly/backend/mxnet_backend.py
```diff
@@ -207,6 +218,17 @@ def partial_svd(matrix, n_eigenvecs=None):
...
+    except AttributeError:
+        warnings.warn('This version of MXNet does not include the linear '
+                      'algebra function gelqf(). Substituting with numpy.')
+        Q, R = map(tensor, numpy_backend.qr(to_numpy(matrix), **kwds))
```
JeanKossaifi: Similarly, I would just write 
```
Q, R = numpy_backend.qr(to_numpy(matrix))
return tensor(Q), tensor(R)
```
I'd also skip the **kwds here.

tensorly/tests/test_backend.py
```diff
@@ -308,3 +308,75 @@ def test_partial_svd():
...
+    A = T.arange(3*4*5)
+    T.assert_equal(T.ndim(A), 1)
+
+    shape1 = (3*4,5)
+    A1 = T.reshape(A, shape1)
+    T.assert_equal(T.ndim(A1), 2)
+
+    shape2 = (3,4,5)
+    A2 = T.reshape(A, shape2)
+    T.assert_equal(T.ndim(A2), 3)
```
JeanKossaifi: Not sure tests are needed for `shape` and `ndim` since we don't actually implement these ourselves but I guess it doens't hurt..

tensorly/backend/mxnet_backend.py
cswiercz: Sure. Good point.

tensorly/backend/mxnet_backend.py
cswiercz: How come? I suppose the `qr()` keywords differ between backends...

JeanKossaifi: Since i) it is only used in the NumPy case so leads to potentially inconsistent behaviour and ii) it's not really needed :)

tensorly/backend/pytorch_backend.py
cswiercz: The reason why I wrote it this way is because of the way Pytorch handles function arguments. (Poorly, in my opinion.) Take a look at the following:

```python
>>> A = torch.arange(9).resize_(3,3)
>>> torch.norm(A, p=1)  # this is fine
36.0
>>> torch.norm(A, p=1, dim=0)  # also fine
  9
 12
 15
[torch.FloatTensor of size 3]
>>> torch.norm(A, dim=0)  # uh oh
TypeError: torch.norm received an invalid combination of arguments - got (torch.FloatTensor, dim=int)
>>> torch.norm(A, p=1, dim=None)  # also wrong. this is the key difference
TypeError: torch.norm received an invalid combination of arguments - got (torch.FloatTensor, dim=NoneType, p=int)
```

So Pytorch essentially doesn't allow keyword arguments. But more key to this issue is that `dim=None` is not allowed.

tensorly/tests/test_backend.py
cswiercz: This might be left over from when I was modifying the behavior of `reshape()`. I can remove.

tensorly/backend/mxnet_backend.py
cswiercz: Sure.

tensorly/backend/pytorch_backend.py
JeanKossaifi: Good point.

cswiercz: I will add a comment mentioning something about this

cswiercz:
I also noticed that the infinity norm was broken in MXNet and Pytorch so I went ahead and fixed that. (That's what happens when you forget to implement a test.)

tensorly/backend/numpy_backend.py
```diff
@@ -188,5 +188,5 @@ def partial_svd(matrix, n_eigenvecs=None):
         U, S, V = U[:, ::-1], S[::-1], V[:, ::-1]
         return U, S, V.T
 
-def qr(matrix, **kwds):
-    return np.linalg.qr(matrix, **kwds)
+def qr(matrix):
+    return np.linalg.qr(matrix)
```
JeanKossaifi: Should be simply an import.

tensorly/backend/pytorch_backend.py
```diff
@@ -240,5 +245,5 @@ def partial_svd(matrix, n_eigenvecs=None):
     return U, S, V
 
 
-def qr(tensor, **kwds):
-    return torch.qr(tensor, **kwds)
+def qr(tensor):
+    return torch.qr(tensor)
```
JeanKossaifi: Same, could be just an import

PR #33, opened 2018-02-01
Add where() to Pytorch backend

cswiercz:
Pytorch does not seem to include the `where()` function so we implement our own.

Note that this only provides the three-argument version. `np.where(condition)`, where the other two arguments are `None`, is equivalent to `condition.nonzero()`. MXNet, however, only supports the three argument version.

This function will be used in #29 .


cswiercz:
Oops. I forgot to commit the change I made that fixed the error produced. (Working to quickly...)

JeanKossaifi:
The looping here would be quite slow.

The following should give the same results while avoiding the loop:
```python
def where(condition, x, y):
    res = zeros(condition.shape, **context(x))
    res[condition] = x
    res[~condition] = y
    return res
```

PR #37, opened 2018-02-14
Refactor parafac initialization

cswiercz:
We factor out the factor initialization code from the main parafac algorithm. We
do this in anticipation of upcoming changes to the main algorithm mentioned in
PR #29. Otherwise, `parafac()` will be too large of a function and, therefore,
difficult to debug.

Added a collection of randomized parafac tests, as well.

### DIFF
--- a/tensorly/decomposition/candecomp_parafac.py
+++ b/tensorly/decomposition/candecomp_parafac.py
@@ -1,5 +1,3 @@
-import logging
-
 import numpy as np
 
 from .. import backend as T
@@ -7,8 +5,6 @@ from ..random import check_random_state
 from ..base import unfold
 from ..kruskal_tensor import kruskal_to_tensor
 from ..tenalg import khatri_rao
-
-logger = logging.getLogger(__name__)
 
 # Author: Jean Kossaifi <jean.kossaifi+tensors@gmail.com>
 # Author: Chris Swierczewski <csw@amazon.com>
@@ -81,23 +77,21 @@ def initialize_factors(tensor, rank, init='svd', random_state=None):
     rng = check_random_state(random_state)
 
     if init is 'random':
-        logger.info('[CPDecomp] Using random initialization')
         factors = [T.tensor(rng.random_sample((tensor.shape[i], rank))) for i in range(T.ndim(tensor))]
         factors, _ = normalize_factors(factors)
         return factors
 
     if init is 'svd':
-        logger.info('[CPDecomp] Using svd initialization')
         factors = []
         for mode in range(T.ndim(tensor)):
             U, _, _ = T.partial_svd(unfold(tensor, mode), n_eigenvecs=rank)
 
             if tensor.shape[mode] < rank:
                 # TODO: this is a hack but it seems to do the job for now
-                factor = T.tensor(np.zeros((U.shape[0], rank)))
-                factor[:, tensor.shape[mode]:] = T.tensor(rng.random_sample((U.shape[0], rank - tensor.shape[mode])))
+                factor = T.tensor(np.zeros((U.shape[0], rank)), **T.context(tensor))
+                factor[:, tensor.shape[mode]:] = T.tensor(rng.random_sample((U.shape[0], rank - tensor.shape[mode])), **T.context(tensor))
                 factor[:, :tensor.shape[mode]] = U
-                U = T.tensor(factor)
+                U = factor
             factors.append(U[:, :rank])
         return factors
 
--- a/tensorly/decomposition/candecomp_parafac.py
+++ b/tensorly/decomposition/candecomp_parafac.py
@@ -77,11 +77,10 @@ def initialize_factors(tensor, rank, init='svd', random_state=None):
     rng = check_random_state(random_state)
 
     if init is 'random':
-        factors = [T.tensor(rng.random_sample((tensor.shape[i], rank))) for i in range(T.ndim(tensor))]
+        factors = [T.tensor(rng.random_sample((tensor.shape[i], rank)), **T.context(tensor)) for i in range(T.ndim(tensor))]
         factors, _ = normalize_factors(factors)
         return factors
-
-    if init is 'svd':
+    elif init is 'svd':
         factors = []
         for mode in range(T.ndim(tensor)):
             U, _, _ = T.partial_svd(unfold(tensor, mode), n_eigenvecs=rank)
--- a/tensorly/decomposition/candecomp_parafac.py
+++ b/tensorly/decomposition/candecomp_parafac.py
@@ -78,7 +78,6 @@ def initialize_factors(tensor, rank, init='svd', random_state=None):
 
     if init is 'random':
         factors = [T.tensor(rng.random_sample((tensor.shape[i], rank)), **T.context(tensor)) for i in range(T.ndim(tensor))]
-        factors, _ = normalize_factors(factors)
         return factors
     elif init is 'svd':
         factors = []
--- a/tensorly/decomposition/candecomp_parafac.py
+++ b/tensorly/decomposition/candecomp_parafac.py
@@ -1,66 +1,152 @@
+import logging
+
 import numpy as np
+
 from .. import backend as T
 from ..random import check_random_state
 from ..base import unfold
 from ..kruskal_tensor import kruskal_to_tensor
 from ..tenalg import khatri_rao
 
+logger = logging.getLogger(__name__)
+
 # Author: Jean Kossaifi <jean.kossaifi+tensors@gmail.com>
+# Author: Chris Swierczewski <csw@amazon.com>
 
 # License: BSD 3 clause
 
 
-def parafac(tensor, rank, n_iter_max=100, init='svd', tol=10e-7,
-            random_state=None, verbose=False):
-    """CANDECOMP/PARAFAC decomposition via alternating least squares (ALS)
+def normalize_factors(factors):
+    """Normalizes factors to unit length and returns factor magnitudes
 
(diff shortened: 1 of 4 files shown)


cswiercz:
Given previous discussions I would like to hear your opinion of parameterized test cases.

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -1,66 +1,152 @@
+import logging
```
JeanKossaifi: logging seems unnecessary here

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -1,66 +1,152 @@
...
         factors = []
         for mode in range(T.ndim(tensor)):
             U, _, _ = T.partial_svd(unfold(tensor, mode), n_eigenvecs=rank)
 
             if tensor.shape[mode] < rank:
                 # TODO: this is a hack but it seems to do the job for now
-                factor = T.tensor(np.zeros((U.shape[0], rank)), **T.context(tensor))
-                factor[:, tensor.shape[mode]:] = T.tensor(rng.random_sample((U.shape[0], rank - tensor.shape[mode])), **T.context(tensor))
+                factor = T.tensor(np.zeros((U.shape[0], rank)))
+                factor[:, tensor.shape[mode]:] = T.tensor(rng.random_sample((U.shape[0], rank - tensor.shape[mode])))
```
JeanKossaifi: Why did you remove the **T.context(tensor)? 
It is needed to preserve context (dtype and GPU/CPU)

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -1,66 +1,152 @@
...
                 factor[:, :tensor.shape[mode]] = U
-                U = factor
+                U = T.tensor(factor)
```
JeanKossaifi: factor is already a tensor

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -1,66 +1,152 @@
...
     if init is 'random':
-        factors = [T.tensor(rng.random_sample((tensor.shape[i], rank)), **T.context(tensor)) for i in range(T.ndim(tensor))]
+        logger.info('[CPDecomp] Using random initialization')
+        factors = [T.tensor(rng.random_sample((tensor.shape[i], rank))) for i in range(T.ndim(tensor))]
+        factors, _ = normalize_factors(factors)
+        return factors
 
-    elif init is 'svd':
+    if init is 'svd':
+        logger.info('[CPDecomp] Using svd initialization')
```
JeanKossaifi: i would just use a print, using the `verbose` parameter that should be propagated from the main function

tensorly/decomposition/tests/test_candecomp_parafac.py
```diff
@@ -1,10 +1,127 @@
...
+    'order3_mixed_orthogonal': {
+        'dimensions': (16, 12, 8), 'rank': 8, 'seed': 0, 'orthogonal': True},
+    'order4': {
+        'dimensions': (8, 8, 8, 8), 'rank': 4, 'seed': 0},
+    'order4_orthogonal': {
+        'dimensions': (8, 8, 8, 8), 'rank': 4, 'seed': 0, 'orthogonal': True},
+}
+
+
+def create_random_decomposition(dimensions=(10,10,10), rank=10, seed=None,
```
JeanKossaifi: This is a duplicate of `tensorly.random.cp_tensor`, the orthogonal option should be added there.
I think it would make more sense to i) sample a random cp tensor and ii) add some noise rather than have it as a parameter in the generation.

tensorly/decomposition/tests/test_candecomp_parafac.py
```diff
@@ -1,10 +1,127 @@
...
+        factors = [T.qr(factor)[0] for factor in factors]
+    tensor = kruskal_to_tensor(factors, weights)
+
+    if noisy:
+        scale = rank/(rank+10.**(order-1))
+        tensor += T.tensor(scale*np.random.randn(*dimensions))
+    return tensor, factors, weights
+
+
+def test_parafac_errors():
```
JeanKossaifi: I would remove the parameterized test and have just one function.
Also i'd use numpy.testing.assert_raises rather than pytest (this way it still runs with nose).

JeanKossaifi:
For this project I would rather not use parameterized test cases and keep it simple with one test function per class/fun, i do not think the overhead is justified given the use case (c.f. scikit-learn for good examples). 

Generally I try to privilege readability and simplicity of the code for end-users, otherwise, in addition to making it harder for newcomers, things can quickly get out of hand (code rot, over-complexity etc).

tensorly/decomposition/candecomp_parafac.py
cswiercz: This was unintentional.

tensorly/decomposition/candecomp_parafac.py
cswiercz: I had a good reason to do this before but now I forgot. I think it was related to the `float32` vs. `float64` issues in the original code.

tensorly/decomposition/candecomp_parafac.py
cswiercz: As you wish. However, I recommend looking into the `logging` module for controlling what sort of statements are sent to stdout. The opposing argument to using `verbose` parameter is that it adds clutter to the argument list (every related function and sub-function would need to handle verbosity) and there are no built-in level controls: during "normal" operation you may want to only print high-level information but while debugging you may want to run the script with lower level output.

But since no other part of Tensorly uses logging I can remove it.

tensorly/decomposition/tests/test_candecomp_parafac.py
cswiercz: 1. Didn't realize that `random.cp_tensor` existed.`
2. Sure.

tensorly/decomposition/tests/test_candecomp_parafac.py
cswiercz: 1. (See comments below regarding parametrized testing.)
2. Sure.

cswiercz:
*(Re: parametrized tests)* Would you still be okay with my using several hand-written test cases? Also, what are your thoughts when (or if, for that matter) I introduce line search and QR-decomposition options? My plan was that these would be flags to the main function:

```python
def parafac(..., enable_qr=False, enable_ls=False):
    # function body
```

So if I wanted to test whether `parafac()` returns accurate enough results then based on our discussion so far you would prefer a single test with the following format:

```
def test_parafac():
    # single test case with enable_qr=False, enable_ls=False
    ...
    # single test case with enable_qr=True, enable_ls=False
    ...
    # single test case with enable_qr=False, enable_ls=True
    ...
    # single test case with enable_qr=True, enable_ls=True
    ...
```

(You mentioned before that you prefer having a single test function per library function.) Though simple, long functions affect readability. This is a classic case of using parametrization to make the code more succinct. (Though, I agree that the parametrized test I currently have written is somewhat obfuscating.) If you agree that parametrizing would be okay in this future situation, I could certainly manually create an auxiliary test function which accepts these flags as input and the main test would just call this function. If you don't agree, then I'm welcome to suggestions. Just laying out my thoughts.

cswiercz:
@JeanKossaifi - I didn't touch `/tensorly/metrics/regression.py`, which is what coveralls is complaining about. What's the deal?

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -1,54 +1,87 @@
...
     rng = check_random_state(random_state)
 
     if init is 'random':
-        factors = [T.tensor(rng.random_sample((tensor.shape[i], rank)), **T.context(tensor)) for i in range(T.ndim(tensor))]
+        factors = [T.tensor(rng.random_sample((tensor.shape[i], rank))) for i in range(T.ndim(tensor))]
+        factors, _ = normalize_factors(factors)
+        return factors
 
-    elif init is 'svd':
+    if init is 'svd':
```
JeanKossaifi: Why not elif? 
(it's the same since we use return inside the statement but might make the logic more immediately obvious?)

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -1,54 +1,87 @@
...
 
-    References
-    ----------
-    .. [1] T.G.Kolda and B.W.Bader, "Tensor Decompositions and Applications",
-       SIAM REVIEW, vol. 51, n. 3, pp. 455-500, 2009.
     """
     rng = check_random_state(random_state)
 
     if init is 'random':
-        factors = [T.tensor(rng.random_sample((tensor.shape[i], rank)), **T.context(tensor)) for i in range(T.ndim(tensor))]
```
JeanKossaifi: You removed the `**T.context(tensor)` when creating the factors, this line should be reverted to the previous one.

JeanKossaifi:
Looks good!

There just misses a `**context` in the init really, and ready to merge!
Let's ignore coveralls, I'll fix it separately since it's unrelated to this PR..

tensorly/decomposition/candecomp_parafac.py
cswiercz: Eh, it's a style thing. I was tempted to use this style in other places I touched but realized that you prefer the `elif`s and wanted to remain consistent. I guess I forgot about that decision, here.

tensorly/decomposition/candecomp_parafac.py
cswiercz: Thanks for catching this one, too. I'll add it back in.

JeanKossaifi:
Thanks, merging!

PR #42, opened 2018-04-08
Python 2 compatibility

jacobgil:
- import_lib.reoad isn't supported in Python 2, so reload is used if import_lib.reload raises an exception.
- Exception chaining (raise ImportError(message) from err) isn't supported in Python 2.

### DIFF
--- a/tensorly/__init__.py
+++ b/tensorly/__init__.py
@@ -25,8 +25,11 @@ def set_backend(backend_name):
     _BACKEND = backend_name
 
     # reloads tensorly.backend
-    importlib.reload(backend)
-
+    try:
+        importlib.reload(backend)
+    except:
+        reload(backend)
+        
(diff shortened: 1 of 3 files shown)


jacobgil:
Ok, Now I see that 30 minutes ago you added a Python 3 warning to the readme :-) 

I was working seamlessly with Python 2 though until now (with the numpy backend).

JeanKossaifi:
Yes, I had accidentally removed the warning while editing the README, but fixed it today as I was doing some improvement work on the documentation...

It was a deliberate choice to develop TensorLy [exclusively for Python 3](http://tensorly.org/stable/installation.html#why-python-3) and I don't think it would add enough value ([especially](http://python3statement.org/) [now](https://pythonclock.org/)) to justify limiting the new functionalities we can use, adding complexity to the code base, etc..

PR #43, opened 2018-04-09
Make example tensor visible in docs

SamJohannes:
Fixes a display error at http://tensorly.org/stable/user_guide/tensor_decomposition.html where the example tensor isn't shown. 

Also a few other very minor spelling changes in the docs


JeanKossaifi:
Great, thanks!

PR #45, opened 2018-04-11
eigenvalue computation requires conjugate transpose

gbaier:
For tensors containing complex numbers the partial SVD needs to take the conjugate transpose. All unit tests still pass.


JeanKossaifi:
Awesome, thanks @gbaier! :)

PR #46, opened 2018-04-18
Fix unfload undefined

tvpavan:
Making the example to work


JeanKossaifi:
Thanks for spotting and fixing! :)

PR #55, opened 2018-06-26
ENH: get sparse tensors

stsievert:
This implements getting sparse tensors from FROSTT, http://frostt.io/. It stores these tensors as sparse tensors using https://sparse.pydata.org/.

Because it relies on pydata/sparse, these datasets will only be effective with the NumPy backend.


JeanKossaifi:
Thanks @stsievert, this is a great addition.

For now, it is probably be best to put in the `tensorly/contrib` submodule (e.g. `tensorly.contrib.frostt`) as it breaks the backend system (works only for NumPy). 

I am planning to have sparse tensor support added in TensorLy so this would be a great match (ideally supporting all backends but in the short term might need to implement a backend specific solution). Until this is added, we should probably check that the current backend is NumPy. I could also add a tensorly-wide variable to store the folder in which data is stored.

stsievert:
> I am planning to have sparse tensor support added in TensorLy

I have similar plans – there's some interest in sparse tensor decomposition, and I'm involved with some of that.

This PR doesn't have support for column labels (though they're provided in the dataset), though there is a relevant xarray issue: https://github.com/pydata/xarray/issues/1375.

PR #59, opened 2018-07-03
Update candecomp_parafac.py

AlexStasik:
Made the output of the iteration errors an optional argument and adjusted the documentation.
Also removed the return of weights from the documentation.


JeanKossaifi:
Great thanks!

PR #60, opened 2018-07-04
ENH: get pretrained PyTorch models

stsievert:
Gets a pretrained model from https://pytorch.org/docs/stable/torchvision/models.html. This function returns a list of PyTorch tensors representing the different learned weights of the neural network. These tensors are of different sizes and different orders.

The test for this function downloads 46MB, the smallest file I could find after a brief search. Subsequent tests after the data has been downloaded are very fast, 0.7 seconds on my machine.

### DIFF
--- a/tensorly/contrib/__init__.py
+++ b/tensorly/contrib/__init__.py
@@ -2,3 +2,5 @@
 
     Allows to add quickly and test new functions for which the API is not necessarily fixed
 """
+
+from ._torch import pytorch_model
--- a/tensorly/contrib/_torch.py
+++ b/tensorly/contrib/_torch.py
@@ -1,7 +1,7 @@
 import torchvision.models as models
 
 
-def get_pytorch_model(model='resnet18'):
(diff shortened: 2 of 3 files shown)


stsievert:
How do you want to test this? PyTorch isn't a requirement, and this depends on torchvision (typically installed alongside PyTorch). How should that be handled?

JeanKossaifi:
I would add it in `contrib` and test whether the backend is set to pytorch (using `tensorly.get_backend`).

However, given how trivial the function is I'm not sure it's actually needed..

stsievert:
> However, given how trivial the function is I'm not sure it's actually needed..

I agree it's trivial. I see this as being advantageous for discoverability and ease-of-use. Users may not know where to get pretrained weights, even if they know to think of neural net weights as tensors.

I've sped up the tests by making sure `model` parameter is passed to PyTorch correctly and that `torchvision.models` raises an error.

Issue #57, opened 2018-07-03
Sparse array support


mrocklin:
Sparse arrays have been mentioned in a few issues.  Are there concrete plans for this?  

Assuming that no, there are no concrete plans, I'd be inclined to throw [this sparse array](http://sparse.pydata.org/en/latest/) library as a backend and see how well the current algorithms work.  My guess is that things would densify pretty quickly and that sparse-specific algorithms are likely to be necessary, but it might be worth a try.

JeanKossaifi:
Yes, it's worth a try, I'll give it a shot when I get a moment (feel free to open a PR if you want!)
There is a related PR to load sparse tensors - #55

JeanKossaifi:
Added in #77, #84, #84 ...

PR #64, opened 2018-07-18
ENH: add sparse tensors for numpy backend

stsievert:
This uses [pydata/sparse] to implement sparse tensors, at least for the NumPy backend.

pydata/sparse has a fairly complete implementation for sparse tensors. I would like to use this work for the other backends, but am uncertain on how to best do that. I've opened up https://github.com/pydata/sparse/pull/166 for this.

[pydata/sparse]:https://github.com/pydata/sparse/


JeanKossaifi:
Sparse tensors is a great and needed addition. However, it is not clear what the best way forward is. 

One would be to add a new sparse module (e.g.`contrib.sparse`) that could potentially work only with some of the backends.
Another would be to add sparse as a backend, as suggested by @mrocklin in #57. The issue with that approach is that it would potentially break in some of the algorithms.

There is another, separate issue on whether to have a class `Tensor` (and `SparseTensor`) that would inherit from the backend's array class, or whether to keep the current model (that directly returns a tensor as implemented by the backend library). I will create a separate issue to expand a little on that.

stsievert:
I have pushed some updates; I now implement a `sparse_backend.py` alongside `numpy_backend.py`.

There are sparse tensors in other backends, besides [pydata/sparse](https://sparse.pydata.org/en/latest/) which relies on NumPy.

* PyTorch: [torch.sparse](https://pytorch.org/docs/stable/sparse.html).
* Tensorflow: [tf.SparseTensor](https://www.tensorflow.org/api_docs/python/tf/SparseTensor)
* MXNet: [mxnet.ndarray.sparse](https://mxnet.incubator.apache.org/api/python/ndarray/sparse.html), but it seems to be limited to 2D arrays.

Both Torch and Tensorflow support the COO sparse N dimensional arrays.

tensorly/backend/sparse_backend.py
```diff
@@ -0,0 +1,87 @@
...
+    data = np.asanyarray(data)
+    if dtype is not None:
+        data = data.astype(dtype)
+    if isinstance(data, np.ndarray):
+        if data.dtype.kind == 'f':
+            i = data < np.finfo(data.dtype).resolution
+            data[i] = 0
+        return sparse.COO.from_numpy(data)
+    assert coords is not None
+    return sparse.COO(coords, data)
```
mrocklin: I'm curious, what is this function supposed to do?  Does it ever get a sparse array?  If so then it looks like it converts it into a numpy array and then back out to sparse, which is probably something to avoid.

tensorly/backend/sparse_backend.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def to_numpy(coo_tensor):
+    if coo_tensor.ndim == 2:
+        return coo_tensor.to_scipy_sparse()
+    return coo_tensor.todense()
```
mrocklin: I would have assumed that this requires the output to be a numpy array.  Returning a scipy sparse matrix works?

tensorly/backend/sparse_backend.py
```diff
@@ -0,0 +1,87 @@
...
+        n_eigenvecs = min(6, *matrix.shape)  # 6 is the default for eigsh
+        if min(matrix.shape) == n_eigenvecs:
+            n_eigenvecs -= 1
+    if min(matrix.shape) == n_eigenvecs:
+        raise ValueError('Cannot compute all eigenvectors of matrix')
+    if min(matrix.shape) < n_eigenvecs:
+        msg = 'n_eigenvecs={} is larger than the minimum dimension ({})'
+        raise ValueError(msg.format(n_eigenvecs, min(matrix.shape)))
+
+    U, S, V = numpy_backend.partial_svd(matrix, n_eigenvecs=n_eigenvecs)
```
mrocklin: Is the matrix passed in always a dense array or is it sometimes a sparse array?  If the latter then does it make sense to try to find a sparse svd implementation?  I think that `scipy.sparse.linalg` has something if so.

stsievert: I implicitly formatted the input so the `numpy_backend.partial_svd` would use `scipy.sparse.linalg`. There's still some more work to do on this; it's a WIP.

tensorly/backend/sparse_backend.py
stsievert: Good catch. This function tries to convert to a sparse array, and I've added a catch for when it's a sparse array.

JeanKossaifi:
Nice job!
I don't have any time for myself at the moment but will do a more detailed pass asap.

I am aware of the sparse support in the other framework and it is probably a better choice in the long run. The issue with adding sparse as a backend is that most algos are not adapted at all. It probably is a better way forward to have a module (e.g. tensorly.contrib.sparse), for which we implement a structure (`sparse_tensor`) along specific, adapted algorithms. This would also be (somewhat) backend agnostic.

Also, why do you need `einsum`? We can easily replace it if it is blocking.

stsievert:
> a better way forward to have a module

I've thought about this some, and I agree. I think we should separate {computation on, storage of} sparse tensors.

* `backends.*_backend.py`: storage and simple computation on sparse tensors. This module can...
    * create sparse tensors for that backend
    * obtain relevant attributes from them (e.g., `shape`)
    * do basic math (element-wise or reductions; e.g., `min` or `sqrt`).
* `contrib.sparse`: computation on sparse tensors. This module can provide specialized algorithms for sparse tensors.

This will cover the use case of using a dense algorithm with a sparse tensor too, as long as the backend can accept both sparse and dense arrays. We should write a test suite that checks for this.

It might be worth looking at XND: https://xnd.io/. They have a lot of the same goals as this project, to more cleanly separate the storage/computation/typing of ndarrays with the hopes of generalizing to other frameworks and to resolve the mess around PyTorch/Tensorflow/CuPy/etc.

stsievert:
I've merged https://github.com/tensorly/tensorly/pull/55 into this branch, because they rely on the libraries, require some of the same modifications to requirements.txt and I'm developing examples off the two combined.

JeanKossaifi:
Great work @stsievert!

I have a concern with the current approach in that it adds a new mandatory dependency. One of the core concepts of the project is to keep dependencies to a minimum. However, it is not obvious to me which way forward is the best, I think it's worth having a good think/discussion about this.
I have opened issue #65 to do so.

JeanKossaifi:
Thanks @stsievert for starting this!
Merged in #95

PR #70, opened 2018-08-29
Update candecomp_parafac.py

eriche7133:
Changed "is" to "==" for comparing strings in the initialize_factors function.


JeanKossaifi:
Merging, thanks!

PR #76, opened 2018-10-02
Refactor backends

jcrist:
First off, apologies for the rather large PR. I thought about breaking it up into smaller pieces, but the full changes make sense best when seen together. Much of this is moving around code that already existed and adding docstrings, a minority of the lines are truly *new*.

This refactors the backend design and dispatching in tensorly. For the most part this should have *no* user-facing changes (based on what I understand as the "public" api). The goals here were to:

- Standardize what a backend looks like. Currently a backend is whatever a module in the backend folder happens to contain. This made it tricky to add new backends, as it was hard to differ what was needed in the backend module vs what just happened to be imported. We now have a base class documenting all the methods/attributes needed for a backend, with a docstring and standardized signatures for each.

- Remove use of `importlib` to set/reset backends. The current backend mechanism uses `importlib.reload` to change backends. This can cause problems when switching between backends, as users must only access functions using the namespace (e.g. `tl.tensor`) instead of local imports (`from tensorly import tensor, set_backend; set_backend('mxnet')`). We now dispatch internally in all backend methods, making this less error prone.

- Make changing the backend cheap and thread/context local. This allows swapping between backends at runtime with basically no cost, and doing so in a thread/context local way. This was needed to make the sparse functionality easier to dispatch on (see follow-up PR #77).

- Document the backend-provided functions that are considered public. Everything in the top-level namespace is now contained in the docs, and has at least a rudimentary docstring.

While working/testing, I made the following additional changes:

- Fixed the tensorflow tests, and readded tensorflow to travis
- Added a rudimentary flake8 config, and flaked most of the files I happened to touch. Note that flake8 doesn't pass for the whole repo (would have made this already large PR even larger).
- Removed the backend-specific testing functions, as these all identical across backends. These now live in `tensorly.testing` instead of `tensorly.backend`.


jcrist:
This also makes it possible to parametrize the tests per-backend inside `py.test`. Since setting the backend is now context-local and cheap, we can now handle iterating through the various backends inside pytest instead of externally (e.g. the Makefile). Running `py.test tensorly` would then test all backends that were available, skipping tests for unavailable backends. Adding flags to test only a specific backend would also be easily doable using py.test's test hooks. This is a common pattern in the scipy ecosystem.

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+    Set the backend to numpy globally for all threads:
+
+    >>> tl.set_backend('numpy', make_default=True)
+
+    Set the backend to use pytorch inside a context:
+
+    >>> with tl.set_backend('pytorch'):  # doctest: +SKIP
+    ...     pass
+    """
+    def __init__(self, backend, make_default=False):
```
JeanKossaifi: Not sure I see the point of `make_default` since anyway it will be local to that session -- while I'd expect persistence as a user?
Not sure in which cases we'd care about thread locality.

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+                    importlib.import_module('tensorly.backend.%s_backend' % backend)
+                else:
+                    msg = ("Unknown backend %r, known backends are {%s}"
+                           % (backend, ', '.join(map(repr, _KNOWN_BACKENDS))))
+                    raise ValueError(msg)
+            backend = _LOADED_BACKENDS[backend]
+
+        # Set the backend
+        self._old_backend = getattr(_STATE, 'backend', None)
+        self._new_backend = _STATE.backend = backend
```
JeanKossaifi: What about making these class attributes?
If we want separation between the context and the set_backend function (not sure we do), we could also have set_backend as a staticmethod..

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+
+    >>> with tl.set_backend('pytorch'):  # doctest: +SKIP
+    ...     pass
+    """
+    def __init__(self, backend, make_default=False):
+        if not isinstance(backend, Backend):
+            if backend not in _LOADED_BACKENDS:
+                # load the backend
+                if backend in _KNOWN_BACKENDS:
+                    importlib.import_module('tensorly.backend.%s_backend' % backend)
```
JeanKossaifi: So this automatically stores the backend in thread.local().backend?
We could explicitly do `new_backend = importlib.import_module`?

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+
+def get_backend():
+    """Returns the backend currently used
+
+    Returns
+    -------
+    backend_name : str
+        The backend currently in use
+    """
+    return getattr(_STATE, 'backend', _DEFAULT_BACKEND).backend_name
```
JeanKossaifi: Seeing these two classes act on essentially the same data, it might make sense to have a BackendManager class? These could be staticmethod? 
Again, this is more for discussion than a hard point :)

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+
+
+class BackendAttributeModuleType(types.ModuleType):
+    """A module type to dispatch backend generic attributes."""
+
+    _dynamic_attributes = {'int64', 'int32', 'float64', 'float32', 'SVD_FUNS'}
```
JeanKossaifi: Why did you choose to add the dtypes and SVD-funs here rather than in their respective backends?

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+
+
+def wrap_module(module_name):
+    """Wrap a module to dynamically dispatch attributes to the registry.
+
+    Intended use is
+
+    >>> registry.wrap_module(__name__)
+    """
+    sys.modules[module_name].__class__ = BackendAttributeModuleType
```
JeanKossaifi: This is also pretty neat -- but is it needed?
Unless we have a good reason to use it, i worry it ends up obfuscating needlessly what is happening -- also making understanding and maintaining the code harder, especially for new comers.

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+            V = V.T.conj()
+
+        if not is_numpy:
+            U = self.tensor(U, **ctx)
+            S = self.tensor(S, **ctx)
+            V = self.tensor(V, **ctx)
+        return U, S, V
+
+
+def dispatch(method):
```
JeanKossaifi: Is the purpose of this function to dynamically update the backend being used even in the case where it was imported as 
` from tensorly.backend import some_fun` ? 

Is there a "cleaner" way to do this?

tensorly/testing.py
```diff
@@ -0,0 +1,29 @@
...
+                                  *args, **kwargs)
+
+
+def assert_array_almost_equal(a, b, *args, **kwargs):
+    np.testing.assert_array_almost_equal(T.to_numpy(a), T.to_numpy(b),
+                                         *args, **kwargs)
+
+
+def assert_equal(actual, desired, *args, **kwargs):
+    def _tensor_to_numpy(x):
```
JeanKossaifi: Minor point: maybe this test (checking whether the tensor is one before converting to numpy) should be in the backend.to_numpy?

tensorly/backend/tensorflow_backend.py
```diff
@@ -18,270 +9,159 @@
...
+
+for name in ['float64', 'float32', 'int64', 'int32']:
+    TensorflowBackend.register_method(name, getattr(np, name))
+
+for name in ['ones', 'zeros', 'zeros_like', 'eye', 'reshape', 'transpose',
+             'where', 'sign', 'abs', 'sqrt', 'qr']:
+    TensorflowBackend.register_method(name, getattr(tf, name))
+
+for name in ['min', 'max', 'mean', 'sum', 'prod', 'all']:
+    TensorflowBackend.register_method(name, getattr(tf, 'reduce_' + name))
```
JeanKossaifi: Just a thought: we could just have a list of tuples ('source_fun', 'target_fun_name') which would also give us an easy way to document the correspondance between the wraped function and the original ones from the backend, should we want to.

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+import threading
+import types
+import warnings
+
+import numpy as np
+import scipy.linalg
+import scipy.sparse.linalg
+
+
+_STATE = threading.local()
```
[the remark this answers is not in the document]
jcrist: > Why did you choose to specify the state as a thread-local data?

This is a thread-local because the sparse mechanism works by swapping the backends frequently, and global state should be threadsafe if possible. This means that tensorly can work with e.g. [dask](http://dask.pydata.org/) (my main project) easier, as it's already threadsafe.

>  Would there be any situation where we would want it/not want it to be local?

See comment on `set_default` below.

> Not sure whether we'd use multithreading but I guess in that case we'd have to create one of these per thread?

I'm not sure what you mean by this. All threads share a `threading.local()` object, but attributes dispatch to c-level threadlocals. If you're asking about tensorly backends specifically, the default backend propagates to newly created threads (there's a test for this too).

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+_DEFAULT_BACKEND = None  # changed in `_initialize_default_backend`
+_KNOWN_BACKENDS = ('numpy', 'mxnet', 'pytorch', 'tensorflow', 'cupy')
+_LOADED_BACKENDS = {}
+
+
+def _initialize_default_backend():
```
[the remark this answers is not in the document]
jcrist: I prefer to isolate global setup code into functions, and then call it once at a clearly defined spot. Could go either way on this though.

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+    if backend not in _KNOWN_BACKENDS:
+        msg = ("TENSORLY_BACKEND should be one of {%s}, got %r. Defaulting to "
+               "'%r'") % (', '.join(map(repr, _KNOWN_BACKENDS)), backend, default)
+        warnings.warn(msg, UserWarning)
+        backend = default
+
+    set_backend(backend, make_default=True)
+
+
+class set_backend(object):
```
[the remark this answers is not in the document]
jcrist: > Should it inherit from contextlib.ContextDecorator if that is the goal?

I'd never heard of that class before. It'd be fine to inherit from that, but isn't necessary to be a context manager.

> Not sure if it "deserves" to be class, I guess it's necessary to get the decorator, which in turn is needed for the sparse backend.

Yes, to support setting backends both as a function and as a contextmanager this needs to be a class. I don't expect anyone to ever use it as an object, the only intent is the following use cases:

```python
tl.set_backend(...)  # as a function, sets globally in this thread
with tl.set_backend(...):  # as a context manager, sets contextually
   ...
```

tensorly/backend/core.py
```diff
@@ -0,0 +1,854 @@
...
+    """
+    def __init__(self, backend, make_default=False):
+        if not isinstance(backend, Backend):
+            if backend not in _LOADED_BACKENDS:
+                # load the backend
+                if backend in _KNOWN_BACKENDS:
+                    importlib.import_module('tensorly.backend.%s_backend' % backend)
+                else:
+                    msg = ("Unknown backend %r, known backends are {%s}"
+                           % (backend, ', '.join(map(repr, _KNOWN_BACKENDS))))
```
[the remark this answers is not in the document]
jcrist: I swap between them, and sometimes prefer `%` as it's shorter and I don't have to escape `{`. No strong opinions though, happy to meet your style guide.

tensorly/backend/core.py
jcrist: See comment on `wrap_module` below. They're still defined on their respective backends, but the dispatching needs to be handled here to match the current interface.

tensorly/backend/core.py
jcrist: > Is the purpose of this function to dynamically update the backend being used even in the case where it was imported as
> `from tensorly.backend import some_fun` ?

Yes. The purpose of this is:
- Dispatch methods to the current backend at call time
- Provide stable functions and documentation on all top-level methods (i.e. `tl.ones` is always the same object, regardless of backend)

> Is there a "cleaner" way to do this?

I don't think this is particularly magical, and can't see a simpler way. We're creating methods to call the specific backend functions. If a new method is added, the author needs to:

- Add it to the `Backend` class, with a docstring
- Implement it in whatever backends are needed
- Add a line here creating a new dispatched function

These are process steps, can be documented (happy to add this in a subsequent PR), and don't seem particularly tricky. Methods removing any of these steps would be more magical and [less explicit](https://www.python.org/dev/peps/pep-0020/#id3) (IMHO).

tensorly/testing.py
jcrist: Perhaps, but that becomes tricky because what should `tl.to_numpy([1, 2, 3])` do? It's not a tensor, but the user probably wants it to be an array (and currently `tl.to_numpy` does this, whereas the implementation below won't). I opted to not change the implementation of any of the backend functions in this large PR - future changes could be discussed later.

tensorly/backend/core.py
jcrist: No, this just ensures the requested backend is loaded - in the imported module the backend object is added to the mapping of name -> backend. The lookup in `_LOADED_BACKENDS` is where the backend object is retrieved.

We could do `new_backend = importlib.import_module(...).backend`, but that requires that the backend object be named `backend` in the module (a weird interface). Having a mapping of strings to backend also allows tensorly backends to be defined in other projects (other projects have similar interfaces (e.g. joblib), and while this isn't needed now, it may be nice later).

tensorly/backend/core.py
jcrist: I assume you mean `set_backend` (a class for implementation reasons) and `get_backend` (a function)? I think these make more sense as top-level definitions - the only state shared is global, and nesting them under a class would overly complicate things.

tensorly/backend/core.py
jcrist: I see this being a class as an implementation detail, and would prefer to keep all attributes on this object private. Users should use `get_backend` to query the state, not look at the contextmanager directly.

tensorly/backend/core.py
jcrist: Calling `tl.set_backend` only sets the backend globally for the calling thread (it's a thread local). New threads inherit the default backend unless/until `tl.set_backend` is called explicitly in that thread. For users using threading (in e.g. dask) you may want to set the backend for *all* threads in one call - in that case you can use `make_default` to change the default backend for all threads:

```python
tl.set_backend('pytorch')

# All operation in *this* thread now uses pytorch

# Newly created threads will still default to tensorly's default
# backend, as `set_backend` hasn't been called yet in that thread.

# To get around this, you can change the (not thread local) default
# backend by setting `make_default=True`.

tl.set_backend('pytorch', make_default=True)

# Now all threads that haven't explicitly set a backend themselves will use `pytorch`
```

I think this distinction (and the default value of `False`) makes sense, but it could possibly use a better keyword name than `make_default`. Alternatively, this could be better as a new function `tl.set_default_backend('pytorch')`, which can't be used as a context manager (as that doesn't make sense in this case). Thoughts?

tensorly/backend/core.py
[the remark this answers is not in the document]
JeanKossaifi: > All threads share a threading.local() object

AFAIK the threading.local() object is **not** shared between all threads, the whole point being that the data is thread specific. 

Why not just store all this in a global BackendManager rather than threading.local()?
Though if this is crucial to your applications, we could keep it like this.. 

> the sparse mechanism works by swapping the backends frequently

While an easy way to handle the sparse case, in the end, this might be more of an issue than a feature..

tensorly/backend/core.py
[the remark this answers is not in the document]
JeanKossaifi: In general, I agree. However, in that case, given that this only happens once and calls the `set_backend` function, I think it might be clearer to just move that code to the end of the file and not have it in a function -- this is in any case a minor point.

tensorly/backend/core.py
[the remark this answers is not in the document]
JeanKossaifi: tbh, I am not a big fan of having a context manager to switch backend, and to have to frequently switch, but, as discussed, I guess we can go for this until (and *if*) we find a better alternative to handle the sparse backend..

tensorly/backend/core.py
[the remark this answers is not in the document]
JeanKossaifi: Mainly for consistency :)
Also there is no need for a map here, `', '.join(_KNOWN_BACKENDS)` suffices.

tensorly/backend/core.py
JeanKossaifi: what about `make_session_default`? I personally would simply have the backend be global to the session..

tensorly/backend/core.py
JeanKossaifi: I am suggesting *class attributes*, which *can* be private.

We could have a BackendManager class which would store the state, backend, etc and implement staticmethods which could then be called by various functions.

e.g. 
```python
class BackendManager():
    _old_backend = ...
    _all_backend_info = ...

    def __init__(self, ...):
        pass
```

tensorly/backend/core.py
JeanKossaifi: It is not very explicit though.. 

I think the most common is to explicitly import or use the module name (e.g. in Joblib, `mpickle = import_module(LOKY_PICKLER)`)

Here we could simply get the backend, and do everything (including registering it etc) in the same place.. When you explicitly get the module back it is loaded in the original way as well anyway so this shouldn't break things (e.g. if you do new_backend = importlib.import_module(...), it still imports tensorly.backend.random_backend, in addition to returning said backend).

tensorly/backend/core.py
JeanKossaifi: I would be tempted to get rid of the STATE and simply have a single (global) BackendManager storing and handling everything, including registering the backends functions. Thoughts?

tensorly/backend/core.py
JeanKossaifi: We could group all the loading in a common place? 
The modules, instead of register themselves, could be registered by a BackendManager?

tensorly/backend/core.py
JeanKossaifi: Fair enough - as long as we document this clearly :)
What do you mean by `registry`? Could we put a clearer example using actual tensorly code?

tensorly/backend/core.py
JeanKossaifi: The dispatch function is pretty neat. My issue is more with having this dynamical dispatch at call time: I would be more in favour of dynamically updating the imported functions when needed.

Back to the previous comment, the dispatching could be done by the BackendManager.. 

Clear, explicit, readable code **is** a guiding principle in TensorLy ;)

tensorly/testing.py
JeanKossaifi: Agreed.

tensorly/backend/__init__.py
```diff
@@ -1,21 +1,10 @@
...
-    from .cupy_backend import *
-else:
-    import warnings
-    warnings.warn('_BACKEND should be either "mxnet", "pytorch", '
-                  '"tensorflow", "cupy" or "numpy". '
-                  '{} was given.'.format(_BACKEND))
-    warnings.warn('Using MXNet backend.')
+from .core import wrap_module
+wrap_module(__name__)
+del wrap_module
```
JeanKossaifi: I am not a big fan of this part: could we potentially explicitly import the default backend here rather than in the core? Or some other alternative that makes explicit what is being done? We could potentially register the backends here? Or do it all in `core.py`?

tensorly/backend/core.py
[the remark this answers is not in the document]
jcrist: The `repr` is so they each have quotes around them `{'numpy', 'pytorch', ...}` vs `{numpy, pytorch}`.

JeanKossaifi:
Sorry for the delay -- comments are basically focused on trying to make it all more explicit/clearer, specifically the loading of the backends. 

Side note: do you mind if I merge #73 in the mean time? It has little overlap with this PR but does change a few things in the backend (mainly updating some functions, and adding argmax/min)..

jcrist:
Fine by me. One of these two PRs would have merge conflicts, might as well just fix them here.

tensorly/backend/core.py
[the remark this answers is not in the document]
jcrist: Coming back to this code after a few weeks, I'm thinking it might be best to do the following:

- `set_backend` is not a contextmanager, same api and rules as before
- backend dispatch internally supports a threadsafe context sensitive version (not publically exposed) that is used to make sparse functionality work.

> AFAIK the threading.local() object is not shared between all threads, the whole point being that the data is thread specific.

Apologies, I think we were talking past each other. Semantically yes they are different objects, implementation wise they are the same object which dispatches on attribute access to c threadlocal storage. Quibbles, not important.

tensorly/backend/core.py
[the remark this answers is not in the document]
mrocklin: I suspect that the approach here was modeled after other context-manager-optional solutions that have arisen throughout the ecosystem.  There are, I think, two camps of users:

1.  People who are very cautious and avoid global state like the plague.  They tend to view context managers as mandatory, especially when using code like this in other libraries.  They don't want the use of one library to affect another.  Context managers help to ensure that state changes are localized.
2.  People who are just doing interactive work, and want to set something once and be done with it 

They're both pretty common groups of people.  We've seen the context-manager-optional solution here seems to satisfy both parties.  Dask uses this, Scikit-Learn uses this (for their joblib system), and others as well.

From your statement above it sounds like you tend to operate in mode 2, and presumably most use that you see also is in mode 2.  This is pretty doable by using the "optional" part of this.  The following would work fine:

```python
tl.set_backend(...)
```

You don't need to use this thing as a context manager.

tensorly/backend/core.py
jcrist: Oops, that docstring is from an older implementation. Can fix.

dswalter:
Is there any help or testing that would be useful in getting these things merged? I'd be very interested in being able to get the sparse arrays working in master.

JeanKossaifi:
Hi @dswalter, aiming at having backend refactor and sparse merged by the end of the week.

PR #77, opened 2018-10-02
Add tensorly.contrib.sparse module

jcrist:
This adds a `tensorly.contrib.sparse` module, mirroring the `tensorly` namespace, but with sparse functionality instead of dense. This builds on #76, only the last commit is for this PR.

I'm not 100% happy with the mechanism here, but it's easy with #76 (which I think is a good change regardless if it's used for this). The gist is that the sparse functions are simple wrappers around the normal tensorly functions, using the relevant sparse backend instead of the dense backend. This means that you need to use `tensorly.contrib.sparse.unfold` instead of `tensorly.unfold`, but the sparse versions are just wrapped versions of the dense ones.

So far only the numpy backend implements sparse functionality, using the [pydata/sparse](https://github.com/pydata/sparse/) library (note that you need master to try this). All backend methods are supported, as well as all top-level tensorly methods (e.g. everything wrapped in the `tensorly.contrib.sparse` namespace).

I also included tucker decomposition in the `tensorly.contrib.sparse.decomposition` namespace. For now this is a wrapper around the dense version (as above), but could be replaced with a sparse-specific implementation later.

Supersedes #64.

---

Diff of only the changes in this PR: https://github.com/tensorly/tensorly/pull/77/commits/aa076e5483441e1b607bcc97cadeb1cf159ce9d2

Issue #65, opened 2018-08-07
API for sparse tensors


JeanKossaifi:
API for sparse tensors in TensorLy
============================

This thread is to discuss. collect opinions and thoughts about the best way to incorporate sparse tensors in TensorLy. @stsievert has already started adding `sparse` support in #64.

One thing to consider is whether we want sparse tensors as separate classes with their own methods (probably makes most sense computationally) or whether we want all methods to work with both dense and sparse tensors (might mean lots of inefficient *to_dense* conversions). 

 Some ways forward:

1- sparse_backend
================

Have a separate backend for sparse.

Pros
------

* optional dependency, no issue with current numpy backend

Cons
-------
* separate backend for sparse, means that we are not leveraging pytorch or mxnet's sparse structure
* not clear how to make specialize algorithms for the sparse structure in that context

Example 
---------

``` python
import tensorly as tl
tl.set_backend('sparse') 
a = tensor(...) #this would be a sparse tensor
```

2- as part of the numpy backend
==========================
This is valid for any backend: in addition to `tensor`, the backends would expose a `sparse_tensor` structure. 

Pros
------
* transparently works with all algos
* going forward can be used to leverage other backend's sparse structure (but at the same time can be tough having a unified API for all)

Cons
-------
* additional dependency made mendatory (though we could have a try/except to just check if `sparse` is available)
* forces us to handle sparse tensors in all functions, even where it might not make sense..

Example 
---------

``` python
import tensorly as tl
tl.set_backend('numpy') # the default anyway
t = tl.sparse_tensor(...) 
```

3- a separate `sparse` module
========================
e.g. `tensorly.contrib.sparse`. This would load the correct structure depending on the current backend and add `sparse_tensor` (and potentially specialized algorithms)
Potentially the easiest way forward and seems to encompass the advantages of both previous strategies.

Pros
------

* We can have specialised algorithms for sparse tensors
* We can have the same backend system for sparse tensors as we do for the main tensors (e.g. we load the correct sparse_tensor structure depending on the backend)
* No additional dependency

Cons
-------
* ...

Example 
---------

``` python
from tensorly.sparse import sparse_tensor
```

4- augment the existing tensor
========================

Change the tensor structure to support both dense or sparse tensors, this would be specified in the context.

Pros
------

* Nice transparent interface

Cons
-------

* Same as 2
* Might be trickier to implement for other backends?

Example
---------

``` python
import tensorly as tl
tl.set_backend('numpy') # the default anyway
[...]

PR #84, opened 2018-11-19
Sparse support

asmeurer:
This is a continuation of the work from #76 and #77 to refactor the backends and add a backend for pydata/sparse.

So far I have fixed the merge conflicts with master. We also intend to address the review comments from #76, and fix any test failures. 

We also plan to improve the sparse support beyond what was done in #77, but I will do this in a separate pull request, so that this pull request can be merged sooner.


asmeurer:
I'm assuming tensorflow is failing because of argmin/argmax. I'm unclear why they were not added to the tensorflow backend in https://github.com/tensorly/tensorly/pull/73

JeanKossaifi:
Hi @asmeurer, thanks for the PR!

Indeed argmax/argmin should have been added in #73, feel free to add them here --or I can add this separately.

scopatz:
Hi @JeanKossaifi - Just to let you know, the team here at Quansight is @asmeurer @hameerabbasi and myself. 

I'll take a look at adding the argmin/argmax.

scopatz:
Really excited to be able to work on tensorly!

asmeurer:
From what I could tell [tensorflow's argmin/argmax](https://www.tensorflow.org/api_docs/python/tf/math/argmax) use the same API as numpy, so this should work:

```diff
diff --git a/tensorly/backend/tensorflow_backend.py b/tensorly/backend/tensorflow_backend.py
index b6f0298..0627b3c 100644
--- a/tensorly/backend/tensorflow_backend.py
+++ b/tensorly/backend/tensorflow_backend.py
@@ -154,7 +154,7 @@ for name in ['float64', 'float32', 'int64', 'int32']:
     TensorflowBackend.register_method(name, getattr(np, name))

 for name in ['ones', 'zeros', 'zeros_like', 'eye', 'reshape', 'transpose',
-             'where', 'sign', 'abs', 'sqrt', 'qr']:
+             'where', 'sign', 'abs', 'sqrt', 'qr', 'argmin', 'argmax']:
     TensorflowBackend.register_method(name, getattr(tf, name))

 for name in ['min', 'max', 'mean', 'sum', 'prod', 'all']:
```

That doesn't fix the test failures, though, so I'm not sure what is going on (I haven't had a chance to investigate further).

asmeurer:
Ah, it's not related to argmax, though most likely it will need to be added. https://github.com/tensorly/tensorly/pull/73 was never tested against tensorflow, because Travis testing for it was only added in #76. CC @wumming

jcrist:
The only additional major change I wanted to do was remove making `set_backend` a contextmanager, and make an additional (possibly internal) method for doing that (perhaps `set_backend_locally`). I think this makes sense - most users will want to set the backend globally (and only once), and the current `make_default` kwarg is a bit confusing. The remainder of changes in how the implementation works are all debatable, and I'd be fine with whatever @asmeurer and co want to do.

Apologies for letting this sit after our meeting, I got busy with conferences and other things and this slipped my mind. Glad to have others picking this work up.

JeanKossaifi:
I had actually assumed you were working together. No worries - I also got caught up with work, not easy to keep up with everything! 

I think we were also considering preventing explicit backend imports and not allow this:
```python
import tensorly.backend as T
```
and instead use `import tensorly as tl` directly and then `tl.fun` which would simplify a little the implementation?

jcrist:
> I think we were also considering preventing explicit backend imports and not allow this:
>
> ```python
> import tensorly.backend as T
> ```
>
> and instead use `import tensorly as tl` directly and then `tl.fun` which would simplify a little the implementation?

Many of your examples use `import tensorly.backend as T`, but if you'd be fine with a little backwards incompatibility, then that would simplify the number of public modules.

The function dispatching and attribute proxying will still be needed to make `tl.foo` and `from tensorly import foo` work. As I've said before, I don't think that implementation is overly complex, and users adding new methods won't need to know any of the internals of *how* it works, just how to use it.

scopatz:
Hello All, I have gotten the tests to pass by skipping some of the MPS decomposition tests when on tensorflow.  It turns out that tensorslow does not support fancy indexing.  This was causing [this line](https://github.com/Quansight/tensorly/blob/f532a1678164cb9fd90eeab735ec21bfe480f8b1/tensorly/contrib/decomposition/mps_decomposition_cross.py#L242) to fail because, for example,

```python
idx = (slice(None, None, None), (6, 5, 1, 4), (9, 9, 3, 8), (2, 7, 6, 7))
```

So it is all of those tuple entries (e.g `(6, 5, 1, 4)`) which are not supported. Presumably, this would be something that an `EagerTensor` could implement relatively easily, but that is not really tensorly's job.  Skipping broken behavior seems reasonable for the time being.

wumming:
Hello Anthony, 

You are right. The cross approximation does not support tensorflow since it doesn’t support fancy indexing and its tensors are immutable.
We tried to implement it by using tensorflow's gather_nd, but it turned out to be not efficient. We felt this kind of algorithm is good for numpy but not for tensorflow.
Please let me know if there is anything I can help.

Thank you,


Zongyi

asmeurer:
Some things that are not clear to me on the backend refactor (apologies if this is already mentioned in the review comments on #76, I didn't read them all):

- Why is the backend called 'numpy.sparse'?
- Should we make it so `set_backend('numpy.sparse')` works (presumably after importing `tensorly.contrib.sparse.backend`)? 
- Why is the file with the sparse backend called `numpy_backend.py`?

I guess the idea is that there could be sparse "sub-backends" like a sparse tensorflow or a sparse mxnet, and `sparse` is the sparse version of NumPy, but it's confusing because `numpy.sparse` looks like a module name, which it isn't.

scopatz:
Thanks @wumming - I figured something like that wouldn't be super efficient :wink: Let's not worry about it for the moment.

@asmeurer - yeah, I agree there are some funny things about the naming here.  Especially since `pydata.sparse` is tentatively planned to be absorbed into SciPy.  I'll work on the naming tomorrow, but I think it should just be called sparse.

hameerabbasi:
Small note -- Not planned to be absorbed, just be integrated and to become part of the larger SciPy ecosystem.

scopatz:
Sorry for my abuse of language @hameerabbasi!

jcrist:
> Why is the backend called 'numpy.sparse'
> Why is the file with the sparse backend called numpy_backend.py

This was because the other backends (pytorch, etc...) all have their own sparse implementations, while numpy does not. The intention is that the user picks a backend (`numpy`, `pytorch`, etc...), and the proper container for sparse or dense tensors is used, depending on what method they're calling. The string was just there as a marker for "this is the set of sparse functions that should be used when the backend is set to numpy".

> Should we make it so set_backend('numpy.sparse') works (presumably after importing tensorly.contrib.sparse.backend)?

No, the top level backends should remain the same, the string representation is just some hashable to make the lookup method work. Users should choose to use the numpy backend, and can use sparse functionality using `tensorly.contrib.sparse.*` methods.

scopatz:
I think part of the problem here is that pydata/sparse is *not* the NumPy implemention (nor the SciPy implemetion) of sparse ndarrays. For instance, if NumPy were to add a sparse object, there would be no way in the current code to diambiquate this new object from pydata/sparse.

pydata/sparse is its own project that simply lacks a dense interface (or rather falls back to numpy for its dense interface).

jcrist:
I think that's fine though. My interpretation of `set_backend` is the user taking a stance on what should be doing the computation. The numpy backend could be better read as "use the standard scipy/pydata array interfaces for dense and sparse ndarrays". numpy is a sane default for dense arrays, just as pydata/sparse is a sane default for sparse arrays.

I don't really care what you name the file or key for the sparse interface, my main goal here is that:

```
tl.set_backend('numpy')
tl.contrib.sparse.tensor([1, 0, 0, 0, 0])
```

results in a pydata/sparse ndarray. Since tensorly is designed to abstract all operations away from the user (never use methods, always use the `tl.*` functions), the user really shouldn't need to know what opinionated stance tensorly takes on a numpy sparse representation.

scopatz:
Ahh sure, that makes sense since numpy lacking a sparse implementation could presumably use anything it wanted, including pydata/sparse.  Coming at this from the other direction, though, what should the following return?

```python
tl.set_backend('pydata/sparse')
tl.tensor([1, 0, 0, 0, 0])
```

It seems like if the above returned a numpy array, then pydata/sparse and numpy are aliases for one another.  However, if the above returned a COO pydata/sparse ndarray that just happens to be dense, then the pydata/sparse and numpy backends are pretty different.

Thoughts @jcrist @asmeurer @hameerabbasi @JeanKossaifi?

hameerabbasi:
I've found that you almost always need a mixture of sparse and dense for optimal performance. E.g. `argmax` will always be dense, and that's nicer if your array is small enough.

How to put this into Tensorly is the thing to figure out.

jcrist:
I don't think that should be allowed. Right now the sparse implementations are the same as the dense ones, but that won't be the case for all algorithms (there are better algorithms for sparse tensors). Since pytorch, tensorflow, etc... all have sparse implementations as well, the user should never explicitly set a *sparse backend*, they should set a *backend* and use the sparse implementations in `tensorly.contrib.sparse`.

jcrist:
To be clear, the backend switching mechanism developed here was an easy way to bootstrap the sparse implementations without copying all the dense code over, it shouldn't be taken as an example of user api. The user should explicitly call the sparse methods, which internally swap in a sparse backend.

JeanKossaifi:
Yes, it is not clear which way forward is the best to add sparse tensors (see #65), the goal of `contrib` is to be able to test experimental features and iterate quickly :)

JeanKossaifi:
I do think that for now calls on sparse methods should be explicit (e.g. having `sparse_tensor` and `sparse.fun` special methods).

scopatz:
@jcrist @hameerabbasi and I had a call today, I think we are now all on the same page with respect to how things should be working and what the meaning of the different pieces are.  I'll write these up in the docs.

JeanKossaifi:
Great, looking forward to reading it!

doc/development_guide/sparse_backends.rst
```diff
@@ -0,0 +1,36 @@
...
+
+TensorLy's sparse support exists in the ``tensorly.contrib.sparse`` sub-package.
+This is to enable fast iteration of sparse support.  The sparse sub-package
+contains a mirror of the interfaces in the usual, dense ``tensorly`` package.
+For example, unfolding a sparse tensor would use::
+
+    >>> import numpy as np
+    >>> from tensorly.contrib.sparse import tensor, unfold
+    >>> X = tl.tensor(np.arange(24).reshape((3, 4, 2)))
+    >>> unfold(X, 0) # mode-1 unfolding
```
asmeurer: This is confusing. Neither X nor unfold(X) is sparse.

doc/development_guide/sparse_backends.rst
```diff
@@ -0,0 +1,36 @@
...
+being used.
+
+Backends
+--------
+When users select a backend to be used for tensorly, this backend will also
+be applied to the sparse implementations. This is because many backends
+naively support both dense and sparse tensors as distinct objects. For example,
+TensorFlow and PyTorch both support separate interfaces for dense and sparse
+representations. Using TensorLy's dense or sparse interfaces will give you
+appropriate objects of for the selected backend.
```
asmeurer: typo here

asmeurer:
```suggestion
appropriate objects for the selected backend.
```

doc/development_guide/sparse_backends.rst
scopatz: I have resolved these.

JeanKossaifi:
Looks good to me overall.

Just have a few comments:
* Why did you replace all relative imports (`from .. import xyz`) by absolute ones (`from tensorly.module import xyz`) ?
* Per the discussion with @jcrist, should we simplify a little the implementation, make set_backend set the default backend (e.g. for all threads) and not allow importing the backend directly? 
* Finally, I think it's worth using `sparse_tensor` instead of `tensor` in the sparse backend

I think we should also add some documentation on the way the refactored backend system works, at least in the development section, to make it easier for new users/contributors, but we can add this as we go.

scopatz:
@JeanKossaifi 

> Why did you replace all relative imports (from .. import xyz) by absolute ones (from tensorly.module import xyz) ?

That was @jcrist - and seemed to only apply to the tests.  I have changed them back to relative imports.

> Per the discussion with @jcrist, should we simplify a little the implementation, make set_backend set the default backend (e.g. for all threads) and not allow importing the backend directly?

Coming to this more recently, it is a little difficult to follow precisely what you mean here. I largely agree with the design decisions and justifications provided by @jcrist and @mrocklin in #76 though. Having support for Dask and joblib is a huge boon to usability by downstream packages. I think that getting rid of this kind of support would only entail adding it later on.   Could you help me understand exactly what you would like done here?

> Finally, I think it's worth using sparse_tensor instead of tensor in the sparse backend

I have added `sparse_tensor`.

> I think we should also add some documentation on the way the refactored backend system works, at least in the development section, to make it easier for new users/contributors, but we can add this as we go.

I have added some developer docs. Just let me know if / how you would like this expanded upon!

JeanKossaifi:
Thanks for the update @scopatz. 

I don't mean getting rid of support for dask or such - just floating the idea to not have `set_backend` be a context manager (or perhaps just swap the meaning of the current kwarg, e.g. use local_threadsafe with default False). To most users, the expected behaviour would be to set the backend once and globally. For Dask, we could then use the keyword arg or a specific function -- I don't know the specific requirements as I haven't used it..
I'm not sure why there would be any issue with Joblib though?

Regarding the other comment it's just saying that if we do not make the backend directly usable (e.g. users should import tensorly and use `tensorly.method` rather than `import tensorly.backend as T` and `T.method`) then we can simplify a little the code while enforcing only one way to use the backend properly -- good to decide now rather than later either way.

scopatz:
> re. sparse_tensor: why don't we get rid of tensor and expose only sparse_tensor? So there is no collision of name between the sparse backend and dense tensors?

My argument for keeping `tensor()` around is that this allows the sparse contrib to be a drop in replacement for the base level tensor library, without needing to modify the interface at all.

scopatz:
> use `local_threadsafe` 

OK, I believe that I have added what you were asking for here. 

> users should import tensorly and use tensorly.method

done

> Regarding dev doc I think it would be useful to have some explanations for dev regarding how the dispatching works

Added more docs for this too!

scopatz:
@JeanKossaifi - I think this is aready for another look!

JeanKossaifi:
Thanks for all the improvements! 

I am happy to merge this so we can start using it, and we can still make changes as we go -- especially since the PR is already huge.

I have some general comments: 

#### Backend manager

Back to a previous comment, I am wondering whether we could group some of the logic in a `BackendManager` class, which would have the global variables as class attributes (e.g. `_LOADED_BACKENDS`). We would create one such instance which would handle all the logic?
This should also be easier since we now no longer allow to directly import the backend?

#### Dynamic dispach

While it is a nice way to dynamically dispach functions, etc, manipulating `sys.modules` is not fully satisfying: ideally, I would like a simpler way of dealing with sparse_tensors, explicitely. 

(As a side note, the [doc](https://docs.python.org/3.4/library/sys.html#sys.modules) warns that: `"However, replacing the dictionary will not necessarily work as expected and deleting essential items from the dictionary may cause Python to fail"`. Not 100% sure what this entails -- I believe other references may be  kept, outside this `sys.modules` dict?)

In short, the current state of support for sparse arrays seems to make this sort of "tricks" inevitable, but, on the long run, I would like to find a simpler way.


On an unrelated note, @scopatz, I just realised you are the lead dev of xonsh, it's great! :)

asmeurer:
One note about sparse that may be relevant to this discussion: for many algorithms, it is preferable to use dense arrays in the algorithm, even if the input is sparse. The ALS algorithm appears to be one of these. Because the initial vectors are completely filled, it is a waste to represent them using a sparse datatype.

JeanKossaifi:
Good point, we want to be able to easily combine dense and sparse tensors..

A --somewhat related-- comment: I don't think it makes sense to have both `tensor` and `sparse_tensor` in the sparse backend, we should probably choose one.

asmeurer:
IMHO we should use the same names everywhere and use namespacing to differentiate.

hameerabbasi:
We had a talk with @jcrist about the back ends, he was of the opinion that sparse algorithms should have their own versions and be associated with a dense back-end as well. For example pydata/sparse would be associated with NumPy.

This also fixes the issue of having "mixed" algorithms, using sparse arrays where appropriate and dense ones otherwise.

asmeurer:
Having used this some more, it would definitely be helpful if all the backend functions were directly importable from their respective modules. Right now, with #85, you end up needing "mixed" sparse and dense functions. It is better to use dense functions in some cases because your arrays are dense (including the parafac algorithm itself), and in other cases it is absolutely forbidden to use dense versions of some functions, because they would densify a sparse array.

For instance, the `kr` function in the (dense) numpy backend densifies the array, because `sparse` does not support `einsum`. Instead you have to use the one in the sparse backend (which is just the one from `core`). But I don't want to set the global backend to `numpy.sparse` because I want `parafac`  to be the dense version (the sparse version is about 10x slower). Actually for testing purposes I want to access both the dense and sparse `parafac`. 

I would like to be able to 

```py
from tensorly.contrib.sparse.backend import kr as kr_sparse
```

but that doesn't work. I have to instead use 

```py
from tensorly.contrib.sparse.backend import _LOADED_BACKENDS
kr_sparse = _LOADED_BACKENDS['numpy.sparse'].kr
```

So my suggestion would be

- Functions from the various backends would be namespaced in their respective backend modules
- The default `tl.func` would use the set backend, which the machinery here seems to handle just fine (also considering the other comments in this thread about context managers and whatnot).

JeanKossaifi:
Why not import directly from the parent module instead, e.g. 
```python
from tensorly.sparse import kr as sparse_kr
``` 
rather than `from tensorly.sparse.backend import kr`? (same for the regular backend)

I totally agree that having to manipulate the loaded backend is not desirable. I think users should never have to use/come accross `_LOADED_BACKENDS` and such directly. This is why I suggested gathering all the machinery in a `BackendManager` class which would handle this all.


Regarding the specific issue with the khatri-rao, if you use the implementation I made for the other backends, it might solve the issue:
```python
def kr(matrices):
    """Khatri-Rao product of a list of matrices
        This can be seen as a column-wise kronecker product.
   """
    if len(matrices) < 2:
        raise ValueError('kr requires a list of at least 2 matrices, but {} given.'.format(len(matrices)))

    n_col = shape(matrices[0])[1]
    for i, e in enumerate(matrices[1:]):
        if not i:
            res = matrices[0]
        s1, s2 = shape(res)
        s3, s4 = shape(e)
        if not s2 == s4 == n_col:
        res = reshape(reshape(res, (s1, 1, s2))*reshape(e, (1, s3, s4)),
                      (-1, n_col))
    return res
``` 

 I actually considered moving it completely from the backend to the `tenalg` module, and use this same implementation for all. The only reason I didn't is to allow for specialized implementation (e.g. if CUDA kernels were added later and interfaced with PyTorch, etc).

asmeurer:
Yes, that is the implementation that ends up being used in my example. It's `Backend.kr`, and the sparse backend subclasses from `Backend`.

asmeurer:
Oh sorry I completely missed that they are in `__init__.py`. So yes, that is sufficient. As I said before, I would keep everything as the same name, like `tensor` instead of `sparse_tensor`, and just rely on `import tensor as sparse_tensor` or direct namespacing (`from tensorly import sparse; sparse.tensor`) to differentiate. That's the standard Pythonic way to handle things like this. 

> I actually considered moving it completely from the backend to the tenalg module

Presumably the einsum version is faster for NumPy, since NumPy can compute einsums efficiently. I would have used that version for sparse as well but `sparse` doesn't have einsum implemented yet.

JeanKossaifi:
Great, shall we just remove `sparse_tensor` (leaving only `tensor`) and merge this?

Any comments? @scopatz @hameerabbasi @jcrist @stsievert

scopatz:
> Great, shall we just remove sparse_tensor (leaving only tensor) and merge this?

Done. And sorry about the delay! I do think that this is good-to-go enough for us to proceed. 

> On an unrelated note, @scopatz, I just realised you are the lead dev of xonsh, it's great! :)

Thanks! And thanks for making TensorLy

tensorly/contrib/sparse/backend/numpy_backend.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+_MIN_SPARSE_VERSION = '0.4.1+10.g81eccee'
+if LooseVersion(sparse.__version__) < _MIN_SPARSE_VERSION:
+    raise ImportError("numpy sparse backend requires `sparse` version >= %r"
+                      % _MIN_SPARSE_VERSION)
+
+
+def is_sparse(x):
+    return isinstance(x, sparse.COO)
```
hameerabbasi: This should be `sparse.SparseArray` to ensure that all formats are supported. I may improve `DOK` support and also add more formats.

tensorly/contrib/sparse/backend/numpy_backend.py
```diff
@@ -0,0 +1,144 @@
...
+        if order == 1:
+            return np.sum(np.abs(tensor), axis=axis)
+        elif order == 2:
+            return np.sqrt(np.sum(tensor**2, axis=axis))
+        else:
+            return np.sum(np.abs(tensor)**order, axis=axis)**(1 / order)
+
+    def dot(self, x, y):
+        if is_sparse(x) or is_sparse(y):
+            return self.tensor(x).dot(self.tensor(y))
```
hameerabbasi: This shouldn't *always* run the COO-COO dot product, that could get slow in practice.

scopatz: Ahh so this should just be `x.dot(y)`?

hameerabbasi:
We could run sparse tests as usual, but adding a skip if the back-end is not NumPy.

asmeurer:
Can you clarify what you mean? Is there already an existing way to run the tests with the sparse backend on this branch?

hameerabbasi:
My reasoning is that we run the sparse tests only for the NumPy backend, skipping if the backend isn't NumPy. We run the tests the same way we'd run our notebooks.

asmeurer:
I see the confusion. You are talking about tests against sparse arrays, that would only run against the sparse backend. Yes, for those, we should do it as you say. But I also want to run the existing tests against the sparse backend (for those functions that are supported), to verify that they work. The way the tests are currently set up, you have to set an environment variable to set the backend, but the way the sparse backend was written, it's not so easy to add it as to the `TENSORLY_BACKEND` environment variable.

asmeurer:
@jcrist do you have any suggestions?

JeanKossaifi:
I am actively working on this in my free time! 

I will try to some of my thoughts here. First, I have to reiterate that these are great contributions, I just want to fully understand where this is going.

## 1 - Context Manager
Why do we need the context manager in practice? Aside from switching back and forth between dense and sparse functions (which I am not a big fan of, though it seems the easiest way forward for now), we do not really want people to switch from backend to backend all the time (I get a lot of question from people trying to use tensors from *different* backends *together*). This is a fairly minor point.

## 2 - Overhead of the dispatch
What is the overhead of the dynamic dispatch? I am thinking in cases where we'd call dozens of such functions per iteration of an algorithm for instance. Would we notice a performance drop in practice?

## 3 - Dynamical dispach mechanism 
Currently we "dispatch" the functions, i.e. we decorate them so that they call the function from the current backend. The reason is that this allows to dynamically update functions even if they are imported as `from tensorly import fun`. However, this does not extend to types (e.g. `from tensorly import float32` will be set when importing and not updating upon changing the backend). 

My question: is this a desirable behaviour? Python discourages direct imports of function and encourages imports of modules instead. Should we assume if a function is imported directly then maybe the user wants the function originally imported, while if they use `tensorly.fun` they want the one currently linked to the Backend? 

If we decide to go for this then all could be done directly by wrapping the modules in `sys.modules` as is currently done for the attributes (removing the need for dispatch).

As a side note, in addition to the overhead, there are some issues with the dispatch:
```python
>>> import tensorly as tl
>>> import inspect
>>> tl.set_backend('numpy')
tensorly.set_backend('numpy')

>>> print(inspect.getsource(tl.tensor))
    def inner(*args, **kwargs):
        backend = getattr(_STATE, 'backend', _DEFAULT_BACKEND)
        return getattr(backend, name)(*args, **kwargs)

>>> help(tl.tensor)
Help on function tensor in module tensorly.backend.core:

tensor(data, **context)
    Tensor class
    
    Returns a tensor on the specified context, depending on the backend.
    
    Examples
    --------
    >>> import tensorly as tl
    >>> tl.set_backend('numpy')
    >>> tl.tensor([1, 2, 3], dtype=tl.int64)
    array([1, 2, 3])

```
For discussion's sake I created PR #95 with a Backend Manager centralising the logic and discards `dispatch`. It's a little simplified and mainly just refactoring. I also used module __getattr__ and __dir__ (PEP 562) so this would only work for Python3.7+ but we can use the `sys.module[__name__].__class__` trick instead :)

## Alternative for dealing with sparse tensors
As an alternative to the current switching mechanism between dense and sparse, I considered multiple dispatch. It does have at least 2 issues: 

i) it prevents us from explicitely choosing which function to use 

ii) it requires to define the sparse and dense functions together
    
Note that currently we cannot easily choose whether to use the sparse or dense function either though so we suffer from i) already...

## Final comments
These are the reasons why I feel slightly uneasy and want to be sure we get them right before merging:

i) this is quite a big change in the architecture and I am not entirely sure how easy it would be to change back. I am assuming most of you might not have much time later so also have to take into account how much time it would take to maintain/update.

ii) it adds quite a lot of magic. Of course some magic is good (I am quite amazed by the replacement of a sys.module.__class__) but too much can also make the codebase harder to read and add complexity. In addition, currently the logic is spread in several places and quite a few things are happening (e.g. wrapping modules, dispatching methods, registering methods (and attributes), etc) -- hence my suggestion to have a BackendManager to centralize the backend related logic-- it doesn't matter whether there actually is a BackendManager or not, as long as the logic is grouped.. 

Looking forward to feedback/discussion and having it merged :)

asmeurer:
> Why do we need the context manager in practice? Aside from switching back and forth between dense and sparse functions (which I am not a big fan of, though it seems the easiest way forward for now), we do not really want people to switch from backend to backend all the time (I get a lot of question from people trying to use tensors from different backends together). This is a fairly minor point.

It could be useful for any external library code that needs to use more than one backend. Otherwise, it seems to me to be a nice way to make sure a backend switches on and then off again, e.g., if I call tensorly.sparse.parafac, then all the functions that `parafac` calls should use the sparse variants, but as soon as the function returns it should not use the sparse backend again. 

> 2 - Overhead of the dispatch

I don't suspect it would be the bottleneck, unless there is a use-case where one calls a tensorly function many times on small input (I was under the impression that the main usage is calling the functions once on large input, but correct me if I am wrong on that). 

> As a side note, in addition to the overhead, there are some issues with the dispatch:

Looks like the manual `functools.wraps` logic is missing some things. 

> BackendManager

I'll take a closer look at this and see what I think. 

>  I also used module `__getattr__` and `__dir__` (PEP 562) so this would only work for Python3.7+ but we can use the `sys.module[__name__].__class__` trick instead :)

The module `__getattr__` lets you avoid most sys.module patching, but you can also just continue do the patching in older versions for compatibility. I've seen it used in some other places before and in my experience it only breaks in very obscure scenarios, if at all. 

> As an alternative to the current switching mechanism between dense and sparse, I considered multiple dispatch. It does have at least 2 issues:

That could work for the backend functions. That would at least let you import and use the functions without having to remember use `tl.func` each time. 

It doesn't solve the question of, for instance, using the dense backend `parafac` on a sparse tensor, because the internal tensors are dense anyway, because the arrays that should be dense are the ones created by `initialize_factors`, which calls `tl.tensor`. I guess a way to fix this would be to make `tl.tensor` return a dense array if the input is dense, even if it should otherwise be sparse.

JeanKossaifi:
Thanks for the feedback @asmeurer and Happy Holidays! :)

> I don't suspect it would be the bottleneck, unless there is a use-case where one calls a tensorly function many times on small input (I was under the impression that the main usage is calling the functions once on large input, but correct me if I am wrong on that).

Only the backend functions are dispatched, and these can be called multiple times per iteration (e.g. `reshape` is dispatched and typically called multiple times at each iteration of most algorithms). I guess I need to do some actual timing..

> The module __getattr__ lets you avoid most sys.module patching, but you can also just continue do the patching in older versions for compatibility. I've seen it used in some other places before and in my experience it only breaks in very obscure scenarios, if at all.

I agree -- would you test the version or just use the patching?

My question really is: since we are already using this patching to dynamically dispatch arguments (dtypes etc), is there a good reason not to use it also for the methods (instead of using `dispatch`)? This avoids wrapping functions and we keep only one logic (overriding __class__ in `sys.modules`). 

This also has an effect that can be seen as both a pro and a con: functions imported explicitely (`from tensorly import fun`) will not be updated when changing the backend. However, this might be one solution to the last problem you mention (namely that we cannot explicitly call sparse or dense functions). 

e.g. in sparse functions we could do:
```python
from tensorly import tensor as dense_tensor #fixed to the dense version
import tensorly as tl #will be updated when the backend changes

tl.set_backend('sparse')
tl.tensor(...) # sparse tensor
dense_tensor # regular tensor
```

Another solution: it is not ideal but we could just rewrite all functions we want for the sparse backend and making minor modifications where needed.. I am guessing other backends like pytorch etc will come with their own set of challenges..

asmeurer:
I'm afraid I don't have enough experience with threading to answer that question. Maybe someone else can comment. 

> My question really is: since we are already using this patching to dynamically dispatch arguments (dtypes etc), is there a good reason not to use it also for the methods (instead of using dispatch)? This avoids wrapping functions and we keep only one logic (overriding class in sys.modules).

I might be misunderstanding what you are suggesting, but an advantage of explicitly wrapping the functions is that it makes it explicit which ones are actually supported by the backend.

mrocklin:
Sorry for jumping in here, but as an external user and someone that interacts with a lot of different libraries, I'd strongly prefer that libraries not choose to do fancy things with `sys.modules` or other magical tricks.  While it does provide a neat experience it often results in very difficult to work around issues when multiple libraries are used together.  FWIW I find explicit dispatch solutions like `functools.singledispatch` or `multipledispatch` to be pleasant solutions (though I'm biased towards the latter library).

jcrist:
> I'd strongly prefer that libraries not choose to do fancy things with sys.modules or other magical tricks. While it does provide a neat experience it often results in very difficult to work around issues when multiple libraries are used together.

In this case, this is just a patch around older versions of Python not having `__getattr__` definitions for modules. I don't see a way this can bleed out of being an implementation detail for tensorly, it shouldn't affect other libraries at all.

mrocklin:
OK, I'll disappear back into the hedges then.

JeanKossaifi:
Happy new year all! :)

@mrocklin I totally agree and that's been my reaction too. I actually mentioned multiple dispatch it in a comment above, and found your excellent work on the subject while researching it!

As @jcrist mentioned there doesn't seem to be easier ways to dynamically update the attributes when changing the backend.. And while PEP 562 makes it nice, it only works for Python 3.7+..

Issue #80, opened 2018-10-21
Optimization submodule


cohenjer:
# Adding an optimization module
For now, Tensorly (TL) ships with one API for each particular tensor decomposition model. While this has the advantage of simplicity for the end-user, this limits the settings where TL can be used.

### Proposition: optimization submodule
I think a very cool feature of tensorly would be, just like in PyTorch or SciPy, to have an optimization module with a few usual algorithms often used for constrained tensor decompositions (nonnegative least squares, Lasso, Gauss Newton, ADMM...). This is important in my opinion since specific applications sometimes require specific algorithms and constraint sets, and there should not be a default one for every scenario.
### API Design
There are mainly two possibilities for such an integration. 
##### 2/ Write a contrib.optim module, and use it in a decomposition.constrained_parafac function where one may choose the optimization method and the constraint set (among a few specific choices).

My opinion: go for contrib.optim. But I will make some tests and see how both solutions behave in practice. The only constraint I think should be handled in a very different way is sparsity, along the lines of issue #79. 

### Remark: using optimization backends?
After some tests and thinking (see for instance [this post](http://www.tensorworld.org/discussions/2018/09/30/tensorly-starting-point.html)) I do not think using an optimization backend such as CVXPY for instance is satisfactory since:
- It increases the dependencies, thus development and maintenance complexity
- It is non trivial to make use of the particular structure of tensor decompositions in already existing, non tensor-optimized software.

JeanKossaifi:
Great to have you on board @cohenjer and looking forward to the new algos! :)

As we discussed I agree on all points: 
* we want more decomposition algorithms
* it is not fully clear how the optim module would compare to "hard-coded" algos, let's compare
* it is probably best not to add more dependencies to optim packages that might additionally not be compatible with all decompositions and all backends

Regarding sparsity, it is still work in progress (#76, #77) but ideally we could simply try using the same algos (as done in #77) and specialise them as we go where needed.

JeanKossaifi:
One really cool feature to add would be, given a CP decomposition learned on a given tensor, approximate the CP decomposition of new one. This would allow us to create a scikit-learn-like object implementing `fit` and `transform`. Thoughts?

animakumar:
Yes! But can we ultimately make it more general, e.g. if it is Tucker,
maintain as Tucker. For instance if you contract CP tensor with factors,
Tucker form is more efficient for output. Can we also design heuristics to
figure that out?

Anima

[...]

PR #85, opened 2018-12-06
Update the parafac algorithm to work with sparse arrays

asmeurer:
This is based off https://github.com/tensorly/tensorly/pull/84. You can see the diff from that branch at https://github.com/Quansight/tensorly/compare/sparse...Quansight:sparse-parafac?expand=1.

The parafac algorithm required several changes to avoid creating intermediate arrays that are large if the input array is large. In particular, if the input array has shape (i0, i1, ..., in) then no intermediary array should have shape that contains more than one of the i's, unless it is just a reshape of the original tensor, which we assume will be sparse. In practice, this means avoiding explicitly computing the Khatri-Rao of the factors at any point in time.

The two main parts of the algorithm that required modification are the matrix-times-khatri-rao, and the computation of the norm. The matrix-times-khatri-rao can be computed directly without explicitly computing the Khatri-Rao of the factors. The norm of ||tensor - kruskal_to_tensor(factors)|| can be computed by expanding the norm out. See the commit messages for more information. The ideas here were mostly taken from https://gitlab.com/tensors/tensor_toolbox/ and its associated paper, Bader, Brett W., and Tamara G. Kolda. "Efficient MATLAB computations with sparse and factored tensors." SIAM Journal on Scientific Computing 30.1 (2007): 205-231.

This also required adding `stack` to the backends, and modifying the backends' `moveaxis` functions to work with negative indices. 

One thing to note is that although the input array to `parafac` can now be sparse, the arrays used in the algorithm are still generally dense, so it makes sense to use the dense backend. This has a few issues which I haven't worked out yet, like the sparse array computations "absorbing" the dense arrays and making them sparse. The algorithm itself works with a sparse input both the dense (numpy) and sparse backends.

I have not yet added any new tests, though the existing tests pass.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,25 +1,25 @@
 package:
   name: tensorly
-  version: "0.4.2"
+  version: "0.4.3"
 
 source:
   path: ../
 
 build:
   number: 0
-  noarch: generic
-  script: python setup.py install --single-version-externally-managed --record=record.txt
+  noarch: python
+  script: python -m pip install --no-deps --ignore-installed .
 
 requirements:
   build:
     - python >=3.4
     - setuptools
+    - pip
 
   run:
     - python
     - numpy
     - scipy
-    - matplotlib
 
 test:
   requires:
@@ -34,3 +34,10 @@ about:
   home: https://github.com/tensorly/tensorly/
   license: BSD
   summary: "Tensor learning in Python"
+  description: |
+    TensorLy is a Python library that aims at making tensor learning simple 
+    and accessible. It allows to easily perform tensor decomposition, 
+    tensor learning and tensor algebra. Its backend system allows to 
+    seamlessly perform computation with NumPy, MXNet, PyTorch, TensorFlow 
+    or CuPy, and run methods at scale on CPU or GPU.
+
--- a/doc/development_guide/backend_system.rst
+++ b/doc/development_guide/backend_system.rst
@@ -1,0 +1,63 @@
+.. _backend_system:
+
+Backend System
+==============
+The TensorLy backend system allows for switching between multiple backends in
+a thread-local way.  You can obtain the back that is currently being used with the
+``get_backend()`` function::
+
+    >>> import tensorly as tl
+    >>> tl.get_backend()
+    'numpy'
+
+You may set the backend globally with the ``set_backend()`` function::
+
+    >>> tl.set_backend('tensorflow')
+    >>> tl.get_backend()
+    'tensorflow'
+
+Setting the backend is local to the thread that ``set_backend()`` is
+executed on. This enables third-party packages (e.g. Dask) to parallelize
+operations over multiple backends, as needed. Threads inherit the backend
+from the thread that spawned them (which is typically the main thread).
+Globally setting the backend supports interactive usage.
+
+Additionally, ``set_backend()`` may be used as a context manager to
+safely use a backend only for limited context::
+
+    >>> with tl.set_backend('pytorch'):
+    ...     pass
+
+This is also thread-safe. The context manager approach is useful in
+third-party libraries that wish to ensure a particular backend is used,
+no matter what the global backend is set to at the time the library code
+is executed.
+
+How the Backend System Works
+----------------------------
+When a backend is set, the backend specific module (called
+``tensorly.backend.{name}_backend``) is dynamically imported (via ``importlib``)
+if the backend has not already been loaded. Once the backend is loaded, it is
+grabbed from an internal cache (called ``tensorly.backend.core._LOADED_BACKENDS``)
+and set as the current backend (called ``tensorly.backend.core._STATE.backend``).
+
+The backends themselves are instances of the ``tensorly.backend.core.Backend`` class
+or subclasses, which implement ``staticmethods`` of the common TensorLy API
+(e.g. ``tensor``, ``fold``, ``norm``, etc.). These backend classes are added to the
+``_LOADED_BACKENDS`` cache on import of the backend module
+(``tensorly.backend.{name}_backend``) via the ``register_backend`` functions.
+
+The TensorLy API functions are then dynamically farmed out to the backend
+staticmethods via the dispatching mechanism provided by
+``tensorly.backend.core.dispatch``. This ensure that the API function is
+wrapped such that it has the correct docstring, name, function signature, and
+other required minutia.
+
+Additionally, the modules themselves are also wrapped such that certain
+global (module-level) variables ar dynamically dispatched as well, like
+``property`` or other class descriptors. This is to ensure that variables
+such as ``int32`` or ``float64`` point to the correct, backend-specific
+object.  For example, numpy's ``np.int32`` and tensorflow's ``tf.int32``
+are not compatible. The dynamic dispatch of these module-level varaibles
+is implemented in the ``tensorly.backend.core.BackendAttributeModuleType``
+class.
--- a/doc/development_guide/index.rst
+++ b/doc/development_guide/index.rst
@@ -6,5 +6,7 @@ Development guide
 
    contributing.rst
    documentation.rst
+   backend_system.rst
+   sparse_backend.rst
 
 
--- a/doc/development_guide/sparse_backends.rst
+++ b/doc/development_guide/sparse_backends.rst
@@ -1,0 +1,39 @@
+.. _sparse_backend:
+
+Sparse Backend Development
+==========================
+
+TensorLy's sparse support exists in the ``tensorly.contrib.sparse`` sub-package.
+This is to enable fast iteration of sparse support.  The sparse sub-package
+contains a mirror of the interfaces in the usual, dense ``tensorly`` package.
+For example, unfolding a sparse tensor would use::
+
+    >>> import numpy as np
+    >>> import sparse
+    >>> from tensorly.contrib.sparse import tensor, unfold
+    >>> X = sparse.COO(np.arange(24).reshape((3, 4, 2)))
+    >>> X = sparse.COO(X.coords, X.data, shape=(100, 100, 100))
+    >>> X = tl.tensor(X)
+    >>> unfold(X, 0) # mode-1 unfolding
+    array([[ 0,  1,  2,  3,  4,  5,  6,  7],
+           [ 8,  9, 10, 11, 12, 13, 14, 15],
+           [16, 17, 18, 19, 20, 21, 22, 23]])
+
+This separation makes it explicit whether a sparse or dense algorithm is
+being used.
+
+Backends
+--------
+When users select a backend to be used for tensorly, this backend will also
+be applied to the sparse implementations. This is because many backends
+naively support both dense and sparse tensors as distinct objects. For example,
+TensorFlow and PyTorch both support separate interfaces for dense and sparse
+representations. Using TensorLy's dense or sparse interfaces will give you
+appropriate objects for the selected backend.
+
+However, not all backends have a native sparse representation. For instance,
+NumPy lacks a sparse object. In these cases, TensorLy makes opinionated decisions
+about how to handle sparse support, if at all. It is usually necessary for
+non-native sparse support to require additional dependencies. For the NumPy
+backend, the `PyData/Sparse <https://sparse.pydata.org>`_ project is used
+as the sparse representation.
--- a/doc/modules/api.rst
+++ b/doc/modules/api.rst
@@ -150,6 +150,7 @@ For each backend, tensorly provides the following uniform functions:
     proximal.svd_thresholding
     proximal.procrustes
     inner
+    contract
 
 
 :mod:`tensorly.decomposition`: Tensor Decomposition
--- a/doc/themes/tensorly/home.html
+++ b/doc/themes/tensorly/home.html
@@ -69,7 +69,7 @@
                     </div>
 
                     <div class="tile is-parent">
-                    <a class="tile is-child box" href="{{pathto('user_guide')}}">
+                    <a class="tile is-child box" href="{{pathto('user_guide/index')}}">
                         <p class="title">User Guide</p>
                         <p class="subtitle">A friendly guide to tensor learning</p>
                     </a>
--- a/examples/regression/plot_kruskal_regression.py
+++ b/examples/regression/plot_kruskal_regression.py
@@ -10,7 +10,7 @@ from tensorly.base import tensor_to_vec, partial_tensor_to_vec
 from tensorly.datasets.synthetic import gen_image
 from tensorly.random import check_random_state
 from tensorly.regression.kruskal_regression import KruskalRegressor
-import tensorly.backend as T
+import tensorly as tl
 
 # Parameter of the experiment
 image_height = 25
@@ -22,7 +22,7 @@ ranks = [1, 2, 3, 4, 5]
 
 # Generate random samples
 rng = check_random_state(1)
-X = T.tensor(rng.normal(size=(1000, image_height, image_width), loc=0, scale=1))
+X = tl.tensor(rng.normal(size=(1000, image_height, image_width), loc=0, scale=1))
 
 
 # Paramters of the plot, deduced from the data
@@ -35,14 +35,14 @@ for i, pattern in enumerate(patterns):
 
     # Generate the original image
     weight_img = gen_image(region=pattern, image_height=image_height, image_width=image_width)
-    weight_img = T.tensor(weight_img)
+    weight_img = tl.tensor(weight_img)
 
     # Generate the labels
-    y = T.dot(partial_tensor_to_vec(X, skip_begin=1), tensor_to_vec(weight_img))
+    y = tl.dot(partial_tensor_to_vec(X, skip_begin=1), tensor_to_vec(weight_img))
 
     # Plot the original weights
     ax = fig.add_subplot(n_rows, n_columns, i*n_columns + 1)
-    ax.imshow(T.to_numpy(weight_img), cmap=plt.cm.OrRd, interpolation='nearest')
+    ax.imshow(tl.to_numpy(weight_img), cmap=plt.cm.OrRd, interpolation='nearest')
     ax.set_axis_off()
     if i == 0:
         ax.set_title('Original\nweights')
@@ -56,7 +56,7 @@ for i, pattern in enumerate(patterns):
         estimator.fit(X, y)
 
         ax = fig.add_subplot(n_rows, n_columns, i*n_columns + j + 2)
-        ax.imshow(T.to_numpy(estimator.weight_tensor_), cmap=plt.cm.OrRd, interpolation='nearest')
+        ax.imshow(tl.to_numpy(estimator.weight_tensor_), cmap=plt.cm.OrRd, interpolation='nearest')
         ax.set_axis_off()
 
         if i == 0:
--- a/examples/regression/plot_tucker_regression.py
+++ b/examples/regression/plot_tucker_regression.py
@@ -10,7 +10,7 @@ from tensorly.base import tensor_to_vec, partial_tensor_to_vec
 from tensorly.datasets.synthetic import gen_image
 from tensorly.random import check_random_state
 from tensorly.regression.tucker_regression import TuckerRegressor
-import tensorly.backend as T
+import tensorly as tl
 
 # Parameter of the experiment
 image_height = 25
@@ -22,7 +22,7 @@ ranks = [1, 2, 3, 4, 5]
 
 # Generate random samples
 rng = check_random_state(1)
-X = T.tensor(rng.normal(size=(1000, image_height, image_width), loc=0, scale=1))
+X = tl.tensor(rng.normal(size=(1000, image_height, image_width), loc=0, scale=1))
 
 # Paramters of the plot, deduced from the data
 n_rows = len(patterns)
@@ -36,14 +36,14 @@ for i, pattern in enumerate(patterns):
 
     # Generate the original image
     weight_img = gen_image(region=pattern, image_height=image_height, image_width=image_width)
-    weight_img = T.tensor(weight_img)
+    weight_img = tl.tensor(weight_img)
 
     # Generate the labels
-    y = T.dot(partial_tensor_to_vec(X, skip_begin=1), tensor_to_vec(weight_img))
+    y = tl.dot(partial_tensor_to_vec(X, skip_begin=1), tensor_to_vec(weight_img))
 
     # Plot the original weights
     ax = fig.add_subplot(n_rows, n_columns, i*n_columns + 1)
-    ax.imshow(T.to_numpy(weight_img), cmap=plt.cm.OrRd, interpolation='nearest')
+    ax.imshow(tl.to_numpy(weight_img), cmap=plt.cm.OrRd, interpolation='nearest')
     ax.set_axis_off()
     if i == 0:
         ax.set_title('Original\nweights')
@@ -58,7 +58,7 @@ for i, pattern in enumerate(patterns):
         estimator.fit(X, y)
 
         ax = fig.add_subplot(n_rows, n_columns, i*n_columns + j + 2)
-        ax.imshow(T.to_numpy(estimator.weight_tensor_), cmap=plt.cm.OrRd, interpolation='nearest')
+        ax.imshow(tl.to_numpy(estimator.weight_tensor_), cmap=plt.cm.OrRd, interpolation='nearest')
         ax.set_axis_off()
 
         if i == 0:
--- a/tensorly/__init__.py
+++ b/tensorly/__init__.py
@@ -9,7 +9,7 @@ from .backend import (set_backend, get_backend,
                       clip, where,
                       max, min, argmax, argmin, all, mean, sum, prod,
                       sign, abs, sqrt,
-                      norm, dot, kron, kr, solve, qr, partial_svd)
+                      norm, dot, kron, kr, solve, qr, partial_svd, stack)
 
 from .base import unfold, fold
 from .base import tensor_to_vec, vec_to_tensor
--- a/tensorly/__init__.py
+++ b/tensorly/__init__.py
@@ -1,15 +1,5 @@
-__version__ = '0.4.3'
-
-from .backend import (set_backend, get_backend,
-                      tensor, is_tensor,
-                      context, shape, ndim,
-                      to_numpy, copy,
-                      concatenate, reshape, moveaxis, transpose,
-                      arange, ones, zeros, zeros_like, eye,
-                      clip, where,
-                      max, min, argmax, argmin, all, mean, sum, prod,
-                      sign, abs, sqrt,
-                      norm, dot, kron, kr, solve, qr, partial_svd, stack)
+__version__ = '0.4.4'
+import sys
 
 from .base import unfold, fold
 from .base import tensor_to_vec, vec_to_tensor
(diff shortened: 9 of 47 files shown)


JeanKossaifi:
Thanks - it's great to have more concrete examples when making decisions for the sparse backend! 

Regarding the khatri-rao times matrix, etc, I have been meaning to fix these for a while. One clean way is to add a kruskal_tensor class that allows efficient manipulation of the decomposed tensor -- I locally started this but was waiting for the new backend system to be merged. 

Just opened PR #87 that implements such structure, curious to have your thoughts.

asmeurer:
Yes, I had been thinking it could be useful to have a direct class to manipulate Kruskal tensors. The MATLAB library has one of these, and it makes things much cleaner (for instance, [this line](https://github.com/tensorly/tensorly/pull/85/files#diff-c2a8e1dc4b7e45c4ba50816dd0d5c13bR206) would be replaced with `factors.norm()`). It could be a non-trivial undertaking to implement all the important array ufuncs on KruskalTensor, but it would assumedly make things nicer for someone using tensorly to work with things after they have the decomposition. 

I was also wondering why weights was absent from the algorithm. They can easily be reconstructed post hoc, but assumedly they make things more numerically stable if things are normalized in the algorithm (?). A few minor things in my modifications would need to be modified to handle weights (they need to be dotted into the norm). 

On a related note, I was wondering if it would be better to have a separate mttkrp function that did the matrix-times-khatri-rao. 

I probably don't have time to work on most of these things but I'm happy to give feedback.

asmeurer:
Something I'm a little unclear about here. As I noted, it can make sense to use either the dense `parafac` or the sparse one. Usually you want to use the dense one since the factors will be dense anyway so it's generally faster and uses less memory. The only reason I've seen where you might prefer the sparse one is if the dense `numpy.linalg.solve` fails with a singular matrix error in an early iteration, but `scipy.sparse.linalg.spsolve` with scikit-umfpack proceeds with only a warning and can still produce an answer after convergence. 

But another issue is that if you want to use `init='svd'` you have to use the sparse backend, because that's the only way to make it call the sparse svd. This is because the way the sparse backend works, the `tensorly.contrib.sparse.decomposition.parafac` is wrapped so that all function calls within it are sparse. If you don't use the sparse SVD, it densifies the tensor. 

I've just been using `init='random'` for my testing, but this is something to think about in the way the backends are designed.

asmeurer:
I have merged the code from `non_negative_parafac` into `parafac(non_negative=True)` (`non_negative_parafac` is still there as an alias). That means that it now works for sparse tensors as well (no changes to the nonnegative multiplicative updates were required to work with sparse).

JeanKossaifi:
@asmeurer, thanks for the update! Re. your comment on the backend design, do you think we should change how things are handled?

I am not sure about merging parafac and the non-negative version directly -- I was thinking to have standalone functions for the individual algorithms, combined via a (scikit-learn like) class (see #80).

asmeurer:
I don't know about the backend stuff. Most likely such cases would be rare enough that they could be handled by special keyword arguments as necessary. 

I merged nonnegative and normal parafac them because they are almost identical code. The only difference is the few lines of code under the `if nonnegative` blocks. The algorithm itself is different, but all the surrounding code with things like calculating the norm or the mttkrp is the same for both. This way any fixes to one would always be made to the other. But I can split them back up if you really want me to.

I'm not familiar enough with the scikit-learn API to comment on that idea, but it seems like it could be a good idea.

asmeurer:
@JeanKossaifi thanks for merging the backend stuff. The best part is that the diff of this pull request is now much easier to read. 

What will be the easiest way for you to review the remaining pull requests? They are:

- Sparse parafac: #85 (this PR)
- Sparse parafac with missing values: #91 (which is just a few commits on top of this PR)
- Sparse Robust PCA: #90 (independent of this PR)
- Documentation for sparse: #89 (commits do not depend on any of the above, the it documents behavior from them all)

- Notebooks: https://github.com/JeanKossaifi/tensorly-notebooks/pull/6 (includes notebooks from all the above)

Is it easier to review and merge them separately, or should we merge them into one PR? I have fixed the merge conflicts for this branch. I will hold off on the others until you say how you would prefer to do it.

asmeurer:
The pytorch failures exist on master. I am looking into what happened with mxnet.

asmeurer:
Ah it was a stupid merge conflict resolution mistake. Should be fixed now. 

As for the other error I'm not sure. Maybe there is a bug in my mxnet moveaxis implementation?

JeanKossaifi:
Hi @asmeurer thanks a lot!
Happy to go through these in the orders you had.

Was the mxnet error in the `contrib.decomposition.mps_decomposition_cross`? I think it might need some adjustments indeed. Not sure what happened with pytorch on master, I don't have any issue locally but was using Pythojn 3.7, will check with 3.6

asmeurer:
The mxnet error is `test_multi_mode_dot` in `tensorly/tenalg/tests/test_n_mode_product.py`. As I said it fails/passes randomly, so you may need to run the test a few times to reproduce the failure.

JeanKossaifi:
The updated mttkrp is deceivingly slower (I had tried this when originally writing the CP).

```python
import numpy as np
import tensorly as tl
from tensorly import random
from tensorly import testing

def mttkrp1(tensor, factors, mode):
    mttkrp_parts = []
    for r in range(rank):
        partial_factor = tl.transpose(tl.moveaxis(tensor, mode, -1))
        for i, f in enumerate(factors):
            if i == mode:
                continue
            partial_factor = tl.dot(partial_factor, tl.reshape(f[:,r], (f[:,r].shape[0], 1)))
            partial_factor = tl.reshape(partial_factor, partial_factor.shape[:-1])
        mttkrp_parts.append(partial_factor)
    mttkrp = tl.stack(mttkrp_parts, axis=1)
    return mttkrp 

def mttkrp2(tensor, factors, mode):
    return tl.dot(tl.unfold(tensor, mode), tl.tenalg.khatri_rao(factors, skip_matrix=mode))
```

Timing for various sizes:
```python
for size in [10, 25, 50, 60]:
    print(f'\ntensor of size {[size]*3}')
    factors = random.random_kruskal(shape=[size]*3, rank=5)
    tensor = tl.kruskal_to_tensor(factors)
    for mode in range(tl.ndim(tensor)):
        print(f'  mode {mode}')
        %timeit mttkrp1(tensor, factors, mode)
        %timeit mttkrp2(tensor, factors, mode)
```

Since I only have my latptop right now, I only tested on CPU.
Output (NumPy backend)
```
tensor of size [10, 10, 10]
  mode 0
108 µs ± 641 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
21.3 µs ± 318 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 1
108 µs ± 817 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
23.7 µs ± 345 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 2
108 µs ± 1.1 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
21.6 µs ± 237 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

tensor of size [25, 25, 25]
  mode 0
263 µs ± 2.15 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
35.9 µs ± 935 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 1
261 µs ± 1.41 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
47.9 µs ± 1.98 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 2
271 µs ± 3.41 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
37.2 µs ± 1.5 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

tensor of size [50, 50, 50]
  mode 0
1.02 ms ± 26.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
94.8 µs ± 721 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 1
1.03 ms ± 11.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
166 µs ± 2.72 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 2
1.34 ms ± 8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
100 µs ± 758 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

tensor of size [60, 60, 60]
  mode 0
1.47 ms ± 9.46 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
144 µs ± 1.02 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 1
1.6 ms ± 12.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
274 µs ± 4.87 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
  mode 2
2.23 ms ± 60.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
153 µs ± 1.13 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

For PyTorch backend:
```
tensor of size [10, 10, 10]
  mode 0
274 µs ± 8.16 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
30.4 µs ± 237 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 1
287 µs ± 16 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
37.5 µs ± 516 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 2
284 µs ± 16.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
31.4 µs ± 1.22 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

tensor of size [25, 25, 25]
  mode 0
432 µs ± 61.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
50 µs ± 3.46 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 1
463 µs ± 46.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
73.3 µs ± 4.48 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 2
416 µs ± 13.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
47.7 µs ± 2.27 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

tensor of size [50, 50, 50]
  mode 0
1.43 ms ± 16.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
137 µs ± 26 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 1
1.5 ms ± 133 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
278 µs ± 11.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
  mode 2
1.13 ms ± 82.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
114 µs ± 931 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

tensor of size [60, 60, 60]
  mode 0
2.28 ms ± 31.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
159 µs ± 652 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
  mode 1
2.15 ms ± 74.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
432 µs ± 20.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
  mode 2
1.67 ms ± 69.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
164 µs ± 1.59 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

asmeurer:
@JeanKossaifi I hadn't considered that. Do you have any guesses why it is? Do you think we can make the new version as fast? If not, we should factor it out so it is only used for sparse tensors (IMO mttkrp should be a helper function anyway). 

I'll take a closer look next week.

asmeurer:
I can definitely reproduce this. I had assumed that since the algorithm was more "efficient" it would be faster, but I guess this isn't the case. I'll play with it and see if it can be fixed. I already noticed that you can get some speedup by avoiding `moveaxis` and using `transpose` to do the axis manipulation.

asmeurer:
Here is a line profile. I've split out subexpressions so you can see their individual timings

```
Total time: 0.887433 s
File: test.py
Function: mttkrp1 at line 7

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
     7                                           @profile
     8                                           def mttkrp1(tensor, factors, mode):
     9      1200        854.0      0.7      0.1      mttkrp_parts = []
    10      7200       4619.0      0.6      0.5      for r in range(rank):
    11      6000      17776.0      3.0      2.0          l = list(range(tl.ndim(tensor)))
    12      6000       4510.0      0.8      0.5          l.remove(mode)
    13      6000       3278.0      0.5      0.4          l.reverse()
    14      6000       5174.0      0.9      0.6          newshape = (mode,) + tuple(l)
    15      6000      19293.0      3.2      2.2          partial_factor = tl.transpose(tensor, newshape)
    16     24000      16763.0      0.7      1.9          for i, f in enumerate(factors):
    17     18000       9191.0      0.5      1.0              if i == mode:
    18      6000       2604.0      0.4      0.3                  continue
    19     12000      54156.0      4.5      6.1              a = tl.reshape(f[:,r], (f[:,r].shape[0], 1))
    20     12000     661049.0     55.1     74.5              partial_factor = tl.dot(partial_factor, a)
    21     12000      50721.0      4.2      5.7              partial_factor = tl.reshape(partial_factor, partial_factor.shape[:-1])
    22      6000       4191.0      0.7      0.5          mttkrp_parts.append(partial_factor)
    23      1200      32579.0     27.1      3.7      mttkrp = tl.stack(mttkrp_parts, axis=1)
    24      1200        675.0      0.6      0.1      return mttkrp

Total time: 0.299471 s
File: test.py
Function: mttkrp2 at line 26

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    26                                           @profile
    27                                           def mttkrp2(tensor, factors, mode):
    28      1200      69320.0     57.8     23.1      a = tl.unfold(tensor, mode)
    29      1200      66469.0     55.4     22.2      b = tl.tenalg.khatri_rao(factors, skip_matrix=mode)
    30      1200     163682.0    136.4     54.7      return tl.dot(a, b)
```

This is from running all the shapes and modes as in your example. 

Nothing is popping out to me yet.

asmeurer:
I think a big part of problem is that the mttkrp1 loops over the rank, whereas mttkrp2 takes care of all the ranks in parallel. If you set the `rank = 1` the difference in time is less extreme. 

I wasn't able to figure out how to handle all the ranks at once when I wrote this, but perhaps I should take a second look and see if it can be done.

asmeurer:
I've pushed a faster mttkrp that works directly on the slices. I suspect it's about as fast as we can get while staying sparse-safe. In my tests it's still not as fast as the old version. Actually the old version could potentially be even faster if the whole thing were done as a single einsum (though I haven't tested this). 

Let me know what you think. Should I split out mttkrp to the backend, with the old version for numpy and my version for sparse?

asmeurer:
Here is an analysis using [einsum_path](https://docs.scipy.org/doc/numpy/reference/generated/numpy.einsum_path.html#numpy.einsum_path)

```py
>>> import numpy as np
>>> rank = 5
>>> shape = (20, 30, 40)
>>> import tensorly as tl
>>> from tensorly import random
>>> factors = random.random_kruskal(shape=shape, rank=rank)
>>> tensor = tl.kruskal_to_tensor(factors)
>>> tensor.shape
(20, 30, 40)
```

By default, it chooses the same path I use, to merge each factor into the tensor one-by-one to avoid dotting the (dense) factors together (the `(0, 1), (0, 2)` part):

```
>>> E = np.einsum_path('ijk,ih,kh->jh', tensor, factors[0], factors[2])
>>> print(E[0])
['einsum_path', (0, 2), (0, 1)]
>>> print(E[1])
  Complete contraction:  ijk,ih,kh->jh
         Naive scaling:  4
     Optimized scaling:  4
      Naive FLOP count:  3.600e+05
  Optimized FLOP count:  2.460e+05
   Theoretical speedup:  1.463
  Largest intermediate:  3.000e+03 elements
--------------------------------------------------------------------------
scaling                  current                                remaining
--------------------------------------------------------------------------
   4                 kh,ijk->hij                               ih,hij->jh
   3                  hij,ih->jh                                   jh->jh
```

However, it says the optimal version is this:

```
>>> E = np.einsum_path('ijk,ih,kh->jh', tensor, factors[0], factors[2],
... optimize='optimal')
>>> print(E[0])
['einsum_path', (1, 2), (0, 1)]
>>> print(E[1])
  Complete contraction:  ijk,ih,kh->jh
         Naive scaling:  4
     Optimized scaling:  4
      Naive FLOP count:  3.600e+05
  Optimized FLOP count:  2.440e+05
   Theoretical speedup:  1.475
  Largest intermediate:  4.000e+03 elements
--------------------------------------------------------------------------
scaling                  current                                remaining
--------------------------------------------------------------------------
   3                  kh,ih->hik                              ijk,hik->jh
   4                 hik,ijk->jh                                   jh->jh
```

Here it Khatri-Rao's the factors, then multiplies them into the matrix. 

The optimized FLOP count isn't that much lower, but this does suggest that the way the current dense mttkrp works is indeed more efficient and won't be beatable by a sparse-safe version.

JeanKossaifi:
I also played with various implementations (in the doctstring). 
In the core of the function is the fastest version. Deceivingly, taking the n-mode product with the full matrix and then selecting the super-diagonal is faster than taking the n-mode product with each of the factors individually (for each rank) and then stacking..

```python
def  unfolding_times_kruskal(tensor, factors, mode, skip=None):
    """Returns the dot product of an unfolded tensor and a khatri-rao product
    
    
    Notes
    -----
    This is a variant of::
    
        unfolded = unfold(tensor, mode)
        kr_factors = khatri_rao(factors, skip_matrix=mode)
        mttkrp2 = tl.dot(unfolded, kr_factors)

    Multiplying with the Khatri-Rao product is equivalent to multiplying,
    for each rank, with the kronecker product of each factor. 
    In code::

        mttkrp_parts = []
        for r in range(rank):
            component = tl.tenalg.multi_mode_dot(tensor, [f[:, r] for f in factors], skip=mode)
            mttkrp_parts.append(component)
        mttkrp = tl.stack(mttkrp_parts, axis=1)
        return mttkrp 
    
    The same idea could be expressed using einsum::
    
        ndims = tl.ndim(tensor)
        tensor_idx = ''.join(chr(ord('a') + i) for i in range(ndims))
        rank = chr(ord('a') + ndims + 1)
        op = tensor_idx
        for i in range(ndims):
            if i != mode:
                op += ',' + ''.join([tensor_idx[i], rank])
            else:
                result = ''.join([tensor_idx[i], rank])
        op += '->' + result
        factors = [f for (i, f) in enumerate(factors) if i != mode]
        return tl_einsum(op, tensor, *factors)
    """
    rank = factors[0].shape[1]
    projected = tl.tenalg.multi_mode_dot(tensor, factors, skip=mode, transpose=True)
    ndims = tl.ndim(tensor)
    res = []
    for i in range(rank):
        index = tuple([slice(None) if k == mode  else i for k in range(ndims)])
        res.append(projected[index])
    return tl.stack(res, axis=-1)
```

Why not just add kr to the sparse backend (the dense function I wrote should work transparently).
The result will still be sparse. I made a quick test and it seems to work.

JeanKossaifi:
Thinking about it, it could be nice to have a flag to choose which version to use. The current version is faster but in some cases we might want to avoid forming the whole khatri-rao product and trade memory efficiency for compute..

asmeurer:
The problem is if you start with a tensor whose individual modes (times the rank) fit in memory fine, but if you take the product of any of those modes it doesn't. In parafac the factors are always dense, so the Khatri-Rao always takes the full space in memory.

By the way, for `einsum`, it's somewhat hidden in the documentation, but you can use `einsum(op0, sublist0, op1, sublist1, ..., [sublistout])` instead of using a string, where `op1`, etc. are tuples of indices. For instance `np.einsum('ijk,ih,kh->jh', tensor, factors[0], factors[2])` is equivalent to `np.einsum(tensor, (0, 1, 2),  factors[0], (0, 3), factors[2], (2, 3), [1, 3])`. This makes it bit easier for programmatic use, and doesn't limit you to 26 dimensions.

asmeurer:
See the `sparse_parafac` notebook [here](https://github.com/JeanKossaifi/tensorly-notebooks/pull/6/files) for an example that blows up memory if the algorithm isn't sparse-safe.

JeanKossaifi:
> In parafac the factors are always dense, so the Khatri-Rao always takes the full space in memory.

Right. There are two problems really: tensor completion (sparse tensor, dense factors) and sparse decomposition (sparse tensor and factors). I guess we were talking of different ones.

In the case we want sparse factors, the khatri-rao would also be sparse. In the completion case indeed we need a different strategy.  I guess we could have a flag for this (e.g. sparse.parafac could take an argument sparse to actually control sparsity of factors, similar to the non-negative flag).

> By the way, for `einsum`, it's somewhat hidden in the documentation, but you can use `einsum(op0, sublist0, op1, sublist1, ..., [sublistout])`

Thanks, it's much better indeed, would have saved me a lot of string crafting!

JeanKossaifi:
By the way, your updated matrix times khatri-rao function is quite neat! :)

The Travis error is just from tensorflow (using tensor.reshape instead of tl.reshape(tensor)).

asmeurer:
> I guess we could have a flag for this (e.g. sparse.parafac could take an argument sparse to actually control sparsity of factors, similar to the non-negative flag).

How would that work? I tried something like that back when I started on parafac but I couldn't get it to work. For noncontrived examples, like the NIPS dataset, even when starting with sparse factors, they eventually became dense after a few iterations (granted, the rank on the NIPS tensor is so large I never got convergence with it, but I imagine this would hold for any "real world" tensor). 

> By the way, your updated matrix times khatri-rao function is quite neat! :)

It took me a while to figure it out. Eventually the einsum_path helped. It could maybe be made cleaner, I don't know. 

What I wish is that `sparse` had an `einsum` function that automatically chose a sparse path for dense inputs when possible. Then you could just write the whole mttkrp as an einsum and let the library figure out how to do it efficiently.

JeanKossaifi:
It seems that, strangely, this version is still slower than taking the n-mode product along each mode (which in turn is slower than the original)... 
 
The original:
```python
def mttkrp(tensor, factors, mode):
    return tl.dot(tl.unfold(tensor, mode), tl.tenalg.khatri_rao(factors, skip_matrix=mode))
```

Using the n-mode product with the full factor along each mode:
```python
def mttkrp2(tensor, factors, mode):
    rank = factors[0].shape[1]
    projected = tl.tenalg.multi_mode_dot(tensor, factors, skip=mode, transpose=True)
    ndims = tl.ndim(tensor)
    res = []
    for i in range(rank):
        index = tuple([slice(None) if k == mode  else i for k in range(ndims)])
        res.append(projected[index])
    return tl.stack(res, axis=-1)
```

Latest version on the PR:
```python
def mttkrp3(tensor, factors, mode):
    mttkrp = tensor.reshape(tensor.shape + (1,))
    for i, f in enumerate(factors):
        if i == mode:
            continue
        a = int(i > mode)
        f_shape = (1,)*a + (f.shape[0],) + (1,)*(tl.ndim(mttkrp)-2-a) + (f.shape[1],)
        mttkrp = tl.sum(mttkrp*f.reshape(f_shape), axis=a)
    return mttkrp
```

Some timings and checks:
```python
for size in [25, 50, 60]:
    print(f'\nTensor of size {[size]*3}')
    factors = random.random_kruskal(shape=[size]*3, rank=15)
    tensor = random.random_kruskal(shape=[size]*3, rank=15, full=True)#tl.kruskal_to_tensor(factors)
    for mode in range(tl.ndim(tensor)):
        print(f'  Mode {mode}')
        r1 = mttkrp(tensor, factors, mode)
        r2 =  mttkrp2(tensor, factors, mode)
        r3 =  mttkrp3(tensor, factors, mode)
        tl.testing.assert_array_almost_equal(r1, r2)
        tl.testing.assert_array_almost_equal(r1, r3)
        %timeit r1 = mttkrp(tensor, factors, mode)
        %timeit r2 =  mttkrp2(tensor, factors, mode)
        %timeit r3 =  mttkrp3(tensor, factors, mode)

```
```
Tensor of size [25, 25, 25]
  Mode 0
46.4 µs ± 853 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
142 µs ± 4.15 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
528 µs ± 14.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
  Mode 1
57.6 µs ± 511 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
113 µs ± 1.26 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
503 µs ± 3.12 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
  Mode 2
47.1 µs ± 348 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
121 µs ± 620 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
492 µs ± 3.92 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

Tensor of size [50, 50, 50]
  Mode 0
175 µs ± 1.35 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
335 µs ± 3.46 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
5.26 ms ± 21.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
  Mode 1
252 µs ± 9.79 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
193 µs ± 2.21 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
5.23 ms ± 36.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
  Mode 2
180 µs ± 970 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
215 µs ± 1.96 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
5.22 ms ± 67.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

Tensor of size [60, 60, 60]
  Mode 0
257 µs ± 1.76 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
495 µs ± 16 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
9.08 ms ± 79.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
  Mode 1
404 µs ± 2.19 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
255 µs ± 2.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
8.99 ms ± 20.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
  Mode 2
265 µs ± 2.16 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
292 µs ± 11.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
8.96 ms ± 36.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

asmeurer:
Version 2 isn't sparse safe, correct? 

How about a version that uses einsum for the whole thing? You can also compare optimize='greedy' (the default) vs. optimize='optimal') (you can also factor out the optimization with `einsum_path`). I suspect that would be the fastest for the dense version, if only because einsum is written in C.

JeanKossaifi:
> Version 2 isn't sparse safe, correct?

Why is it not sparse safe?
It's just dot product - isn't it similar to element-wise multiplication followed by sum?

asmeurer:
Oh my bad. I was just guessing from glancing at the code, but I think you are right that it is. I can confirm mttkrp2 is faster too. 

I tested with

```py
import tensorly as tl
tl.set_backend('sparse') # Note, this requires the sparse-fix branch

shape = (1000, 1001, 1002, 100)
rank = 5

starting_factors = [sparse.random((i, rank)) for i in shape]
from tensorly.contrib.sparse.kruskal_tensor import kruskal_to_tensor
tensor = kruskal_to_tensor(starting_factors)

factors = [sparse.COO.from_numpy(numpy.random.random((i, 5))) for i in shape]
```

This is based off [this notebook](https://github.com/JeanKossaifi/tensorly-notebooks/pull/6/files#diff-dd0217d5264e8986b9415b5cbdef9277).

The `mttkrp2(tensor, factors, 1)` took 14.6 ms and `mttkrp3(tensor, factors, 1)` took `89.5 ms`. `mttkrp(tensor, factors, 1)` just hangs until it fills up all the memory (`tensor` would densely take about 800 GB of memory to represent, and hence the kr of the factors would take about 4 TB). 

Once #97 is merged I'd like to add this as a test to ensure it stays sparse safe.

Issue #88, opened 2018-12-09
Robust PCA


jcreinhold:
Thank you very much for the package, it is great work

I am trying to use the `robust_pca` function on a large dataset and the computation is very slow. This issue seems to be referenced in #18, #36, and #23. I initially tried to use the pytorch backend, thinking that the GPU would speed up the computation. However, I see that in all backends, the `partial_svd` function converts the data into a numpy array and uses the numpy backend (scipy.linalg.svd or scipy.sparse.linalg.eigsh) to calculate the SVD.

My question is two-fold. If I am following the code correctly, it would appear that `robust_pca` will never use the scipy.sparse.linalg.eigsh function. If we examine the code and follow the chain of calculations starting in [robust_pca](https://github.com/tensorly/tensorly/blob/3a35360d82ddefd87c472998cd429f91dd463c49/tensorly/decomposition/robust_decomposition.py#L93) we have a call to `svd_thresholding`. The minimum size of the shape of the input matrix is used as the `n_eigenvecs` keyword (see [here](https://github.com/tensorly/tensorly/blob/3a35360d82ddefd87c472998cd429f91dd463c49/tensorly/tenalg/proximal.py#L70)). However, once `partial_svd` is called, `partial_svd` checks `n_eigenvecs >= min_dim` (see [here](https://github.com/tensorly/tensorly/blob/3a35360d82ddefd87c472998cd429f91dd463c49/tensorly/backend/numpy_backend.py#L219)) to decide to calculate the standard SVD or the partial SVD. Since `n_eigenvecs == min_dim` (because that is what was passed), the check always passes and the partial SVD is never used for calculation.

Is this intentional or should the >= be a >? If it is intentional, i.e., we want to calculate the standard SVD every time, wouldn't it be more computationally efficient to enable backends with that capability to use the GPU enabled SVD calculations? I'm specifically thinking of pytorch which does have an SVD calculation.

Thank you very much for the help

scopatz:
I have a fix for some of the issues (specifically the `n_eignvecs`) ones over at https://github.com/quansight/tensorly/tree/sparse-robust-pca (the `sparse-robust-pca` branch of the Quansight fork). 

It is still quite slow and I have yet to find a sparse data set that is a full working example.

JeanKossaifi:
[...]

PR #90, opened 2018-12-12
Sparse robust pca

scopatz:
Like the other sparse PRs, this builds on #84.  This enables `robust_pca()` to work with the sparse interface. As mentioned in #88, this may not be the most efficient thing to do because the low rank part is necessarily dense (even though it is stored sparsely). However, the sparse part does end up being sufficiently sparse that it makes sense for that tensor.  

An example of this [is available here](https://github.com/Quansight/tensorly-notebooks/blob/sparse/07_pydata_sparse_backend/sparse-robust-pca-demo.ipynb).

Additionally, this currently needs the master version of PyData/Sparse for the `COO.dot()` method. This should be released soon.


tensorly/decomposition/robust_decomposition.py
```diff
@@ -90,7 +90,10 @@ def robust_pca(X, mask=None, tol=10e-7, reg_E=1, reg_J=1,
     for iteration in range(n_iter_max):
 
         for i in range(T.ndim(X)):
-            J[i] = fold(svd_thresholding(unfold(D, i) + unfold(L[i], i)/mu, reg_J/mu), i, X.shape)
+            svd_thresh = svd_thresholding(unfold(D, i) + unfold(L[i], i)/mu, reg_J/mu)
+            if not T.is_tensor(svd_thresh):
+                svd_thresh = T.tensor(svd_thresh)
```
JeanKossaifi: When/why would this happen?
It seems to me that svd_thresholding always returns a tensor.

JeanKossaifi: @asmeurer I think that's the only change to the actual Robust PCA here, the rest is already added in other PRs

asmeurer: CC @scopatz

asmeurer:
I don't know for sure but I think if you [resolve the merge conflicts](https://github.com/tensorly/tensorly/pull/90/conflicts) the diff on GitHub will clean up.

JeanKossaifi:
Merging this. Thanks @asmeurer and @scopatz!

PR #91, opened 2018-12-12
Sparse parafac with missing values

asmeurer:
This is an extension of https://github.com/tensorly/tensorly/pull/85 that adds missing value support to `parafac` in a sparse friendly way. 

The idea to modify the tensor with the masked array is from Tomasi, Giorgio, and Rasmus Bro. "PARAFAC and missing values." Chemometrics and Intelligent Laboratory Systems 75.2 (2005): 163-180.

This required adding a mask flag to kruskal_to_tensor() and kr() so that the mask could be absorbed into the calculation in a sparse friendly way. In order to be sparse friendly, the mask should be a sparse array with a fill value of 1 (True). In other words, the number of missing values should itself be sparse.

The generic kr() algorithm was rewritten to use a generic outer product, and kruskal_to_tensor() now uses sum instead of dot (the NumPy kr() just multiplies the mask).

### DIFF
--- a/tensorly/backend/core.py
+++ b/tensorly/backend/core.py
@@ -687,7 +687,7 @@ class Backend(object):
         b = self.reshape(b, (1, s3, 1, s4))
         return self.reshape(a * b, (s1 * s3, s2 * s4))
 
-    def kr(self, matrices):
+    def kr(self, matrices, mask=None):
         """Khatri-Rao product of a list of matrices
 
         This can be seen as a column-wise kronecker product.
@@ -720,17 +720,17 @@ class Backend(object):
                              'given.'.format(len(matrices)))
 
         n_col = self.shape(matrices[0])[1]
-        for i, e in enumerate(matrices[1:]):
-            if not i:
-                res = matrices[0]
-            s1, s2 = self.shape(res)
-            s3, s4 = self.shape(e)
-            if not s2 == s4 == n_col:
+        if any(i.shape[1] != n_col for i in matrices):
                 raise ValueError('All matrices should have the same number of columns.')
-            a = self.reshape(res, (s1, 1, s2))
-            b = self.reshape(e, (1, s3, s4))
-            res = self.reshape(a * b, (-1, n_col))
-        return res
+        shapes = [(1,)*i + (s,) + (1,)*(len(matrices)-i-1) + (n_col,) for i, s in
+                  enumerate([m.shape[0] for m in matrices])]
+        reshaped_matrices = [self.reshape(i, s) for i, s in zip(matrices, shapes)]
+
+        m = mask[:, ..., None] if mask is not None else 1
+        # This could also use np.multiply.reduce
+        for e in reshaped_matrices:
+            m = m*e
+        return self.reshape(m, (-1, n_col))
 
     def partial_svd(self, matrix, n_eigenvecs=None):
         """Computes a fast partial SVD on `matrix`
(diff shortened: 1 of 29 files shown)


asmeurer:
I forgot to mention you have to use the sparse backend `parafac` here. Otherwise it uses the dense `kr` (the einsum version) which doesn't include the mask in a sparse safe way.

asmeurer:
@JeanKossaifi aside from https://github.com/tensorly/tensorly/pull/97 (not sure what you want to do with it), this PR is next (see https://github.com/tensorly/tensorly/pull/85#issuecomment-460750830).

tensorly/kruskal_tensor.py
```diff
@@ -36,12 +41,14 @@ def kruskal_to_tensor(factors, weights=None):
...
+
     """
     shape = [T.shape(factor)[0] for factor in factors]
     if weights is not None:
-        full_tensor = T.dot(factors[0]*weights, T.transpose(khatri_rao(factors[1:])))
+        factors[0] = factors[0]*weights
+        full_tensor = T.sum(khatri_rao(factors, mask=mask), axis=1)
     else:
-        full_tensor = T.dot(factors[0], T.transpose(khatri_rao(factors[1:])))
+        full_tensor = T.sum(khatri_rao(factors, mask=mask), axis=1)
```
JeanKossaifi: This will use more memory and I assume will be a little slower?

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -4,8 +4,6 @@
 import tensorly as tl
```
JeanKossaifi: Should we use relative import here too?

tensorly/backend/core.py
```diff
@@ -593,17 +593,17 @@ def kr(self, matrices):
...
-            if not s2 == s4 == n_col:
+        if any(i.shape[1] != n_col for i in matrices):
                 raise ValueError('All matrices should have the same number of columns.')
-            a = self.reshape(res, (s1, 1, s2))
-            b = self.reshape(e, (1, s3, s4))
-            res = self.reshape(a * b, (-1, n_col))
-        return res
+        shapes = [(1,)*i + (s,) + (1,)*(len(matrices)-i-1) + (n_col,) for i, s in
+                  enumerate([m.shape[0] for m in matrices])]
+        reshaped_matrices = [self.reshape(i, s) for i, s in zip(matrices, shapes)]
```
JeanKossaifi: Is there any performance gain? I find it significantly harder to read.

tensorly/decomposition/candecomp_parafac.py
asmeurer: Won't it not use the kruskal_to_tensor in the correct backend otherwise?

tensorly/backend/core.py
asmeurer: I don't remember the details of why I did this. I think it has to do with being able to inject the mask into the calculation.

tensorly/kruskal_tensor.py
asmeurer: Sorry it's been a while since I've looked at this. I don't remember why I made this change. If it still works and is sparse safe after reverting it (see https://github.com/JeanKossaifi/tensorly-notebooks/blob/7179c121cc2f21215f9e14cd1b452f8576683c48/07_pydata_sparse_backend/sparse_missing_values.ipynb), that's fine.

PR #95, opened 2018-12-23
Backend manager

JeanKossaifi:
Draft of a grouped logic for managing backend.


JeanKossaifi:
I used a long train trip to do a bit of refactoring and add back the thread local storage and context manager. I seem to have broken something though, will try to look at it soon.

JeanKossaifi:
I meant to add, upon looking at this again, grouping the logic is better but there probably is no need to put it all in a class, might as well just have functions since anyway it's all static methods.

asmeurer:
Some of the methods aren't static. For example [kron](https://github.com/tensorly/tensorly/pull/95/files?utf8=✓&diff=split#diff-4d4aa6f9d8adfb29e2b35efc2cb1afedR552) calls `self.reshape`. I agree the ones that are static methods could be functions. There's functionally no difference when you have everything in a distinct module namespace anyway.

asmeurer:
I guess static methods make it easier to subclass things. That matters for the base class but not as much for the subclasses (though I would keep them for consistency).

JeanKossaifi:
I agree - I was referring to the `BackendManager` class I just added, not the actual `Backend` classes and subclasses.

asmeurer:
Oh sorry I missed that you were talking about the manager class. I agree you should either just use regular functions, or else make the methods real methods.

scopatz:
@asmeurer - can you help with this please?

mcg1969:
Hey folks! I'm curious to see what the status is here, and if there's anything I can do to help. I'm keen to see the sparse tensor support merged into master ASAP and it is my understanding that this is a lynchpin of that effort.

asmeurer:
@JeanKossaifi let me know if there is anything I can do to get this merged.

asmeurer:
Here are @JeanKossaifi's changes against #84, for easier viewing: https://github.com/Quansight/tensorly/compare/sparse...JeanKossaifi:backend-manager

tensorly/backend/tensorflow_backend.py
```diff
@@ -18,261 +9,176 @@
...
+    (tf.reduce_mean, 'mean'),
+    (tf.reduce_sum, 'sum'),
+    (tf.reduce_prod, 'prod'),
+    (tf.reduce_all, 'all'),
+    ]
+for source_fun, target_fun_name in _FUN_NAMES:
+    TensorflowBackend.register_method(target_fun_name, source_fun)
+del _FUN_NAMES
+
+BackendManager.register_backend(TensorflowBackend())
```
asmeurer: I think this needs to be fixed as well.

asmeurer:
Did you also intend to leave the `register_backend` calls out of the various backend files? The function is still there.

asmeurer:
@JeanKossaifi any thoughts on how to make the sparse backend testable with the TENSORLY_BACKEND environment variable? See https://github.com/tensorly/tensorly/pull/84#issuecomment-448302810.

tensorly/backend/tensorflow_backend.py
JeanKossaifi: Yes, this should be removed, thanks for flagging

JeanKossaifi:
@asmeurer  you were right, I had forgotten these bits, thanks!
Regarding the TENSORLY_BACKEND, I am not sure how to best deal with this. There are also the sort of details that make me wonder whether we could find a better solution. Do you have any suggestion?

A naive approach would be to just copy the code completely to the sparse backend... 

Regarding this PR I opened it mainly for discussion, it barely changes anything from what you did :) 
To summarize:
1. A BackendManager was probably not needed
2. I grouped all the logic in the init of backend. `core` contains the base `Backend` class. Each backend inherit from this and implements the various functions. Then, in init, we import from the backend needed and register the corresponding backend. I wanted to separate the logic from the Backend themselves and from the way they are imported.
3. Currently, I just use module's `getattr` to dispatch the methods, rather than explicitly wrapping each method, since we already use that mechanism for the attributes (dtypes). This means that `from tensorly import x` will be fixed to the backend used at that time, while using `import tensorly as tl; tl.x` will always point to the backend currently in use. This seems to be the expected behaviour for Python, and allows to have coherent behaviour for attributes and methods.

Side note: because module's getattr only was introduced for Python 3.7 I will need to re-add the wrap_module function.. Maybe with a check for the Python version?

asmeurer:
> Regarding this PR I opened it mainly for discussion, it barely changes anything from what you did :)

OK. So is the aim to get this PR merged or #84? What are the blockers to getting #84 merged? My priority is to get #85, #90, #91 merged, and am able to help with any changes necessary to get that done. 

> Regarding the TENSORLY_BACKEND, I am not sure how to best deal with this. There are also the sort of details that make me wonder whether we could find a better solution. Do you have any suggestion?

The simplest solution would be to move the sparse code to tensorly.backends, instead of having it in tensorly.contrib.

Another solution that I think would work would be to allow the environment variable TENSORLY_BACKEND to be a fully qualified name to the backend class, like `tensorly.contrib.sparse.backend.numpy_backend.NumpySparseBackend`.

JeanKossaifi:
> OK. So is the aim to get this PR merged or #84? 

Whichever really. I think it’s neater to have i) all the logic grouped and ii) as explicit a code as possible. This is basically what I try to achieve here. I opened this PR to push things forward rather than just commenting on the other PR. Feel free to add to this directly, I will also try to fix the issues I introduced. 

> What are the blockers to getting #84 merged?

This is quite a big change in the architecture and since I'am assuming most of you might not have much time later, I also have to take into account how much time it would take to maintain/update. The added magic makes this a little harder and the core a little more obscure.

1) Is it worth breaking this PR into two parts, first the backend refactoring and second the sparse support?
2) Are we ok with just using module’s getattr rather than dispatching them?
3) Should we tie the sparse backend to the main backend (we still maintain it optional, in a separate `sparse` module, but it is then set depending on the main backend)? A related question: someone mentioned on another thread sparse would become part of Numpy/SciPy: is it still the plan? If so why not just use sparse as the standard numpy sparse counterpart, and for the other frameworks, directly use their `sparse` submodules?
4) Not as important: should we have a `is_sparse` method in the main backend? 

> My priority is to get #85, #90, #91 merged, and am able to help with any changes necessary to get that done.

Do you need the sparse support added now? Would you rather we merged this and change the API later if needed? I would like to experiment a little with the other backends (e.g. PyTorch) to see how it would work there but have quite a few deadlines at the moment, so don’t want to slow things down too much on your end. 

Curious about your use-cases for sparse tensors too :)

> The simplest solution would be to move the sparse code to tensorly.backends, instead of having it in tensorly.contrib.

Do you mean having sparse as a separate backend? It would solve a lot of the issues.

With the cons that it makes it hard to mix sparse and dense. Also, most frameworks seem to have a `.sparse` submodule, does it make as much sense to have a separate backend in these cases?

asmeurer:
> Is it worth breaking this PR into two parts, first the backend refactoring and second the sparse support?

If that makes it easier for you to review then absolutely. I can work on doing this if you want it done.

> Are we ok with just using module’s getattr rather than dispatching them?

`__getattr__` seems better than patching sys.modules for the versions of the language that support it. I'm not clear what you mean by "rather than dispatching them". You mean we don't wrap the various functions with a context manager/decorator so that `tensorly.sparse.parafac` is different from `tensorly.parafac`? I think there were some thread safety concerns here, but others should comment on this.

> Should we tie the sparse backend to the main backend (we still maintain it optional, in a separate sparse module, but it is then set depending on the main backend)?

I think that would help that testing issue, if I understand what you are suggesting. 

I'll also point out that one advantage of having the algorithm functions themselves wrapped is that it makes it explicit which ones have been written to be sparse safe. Calling a non-sparse safe algorithm on an undensifiable sparse input will generally result in a memory error or related crash.

> A related question: someone mentioned on another thread sparse would become part of Numpy/SciPy: is it still the plan?

CC @hameerabbasi to answer this.

> If so why not just use sparse as the standard numpy sparse counterpart, and for the other frameworks, directly use their sparse submodules?

I guess that was the reasoning in calling the sparse backend "NumpySparseBackend". 

> Do you need the sparse support added now? Would you rather we merged this and change the API later if needed? I would like to experiment a little with the other backends (e.g. PyTorch) to see how it would work there but have quite a few deadlines at the moment, so don’t want to slow things down too much on your end.

It would be best to merge now. Things are starting to get messy with all the branches, and merge conflicts are already happening from unrelated changes to master. 

> Do you mean having sparse as a separate backend? It would solve a lot of the issues.

It already is a separate backend. I just mean that having it in the main `backends` directory instead of the `contrib` directory would make testing easier the way the `TENSORLY_BACKEND` environment variable works.

As to why its a separate backend, `sparse` the library is a separate library with a separate type and separate functions. Most NumPy functions do the right thing on sparse as ufuncs, but some don't and require special care, e.g., `einsum`. 

I don't know how tensorflow, pytorch, etc. work with their sparse types. Does the sparse type work exactly like the main type with all the functions? If so, then the existing backend codes will work as is. In fact, you could probably just use their sparse types directly, assuming the algorithm is sparse safe, and it will already work out of the box. If not, that is, if separate functions are needed for sparse arrays vs. dense, then a separate backend could make sense there too. The one caveat is that if you are working with sparse, you want `tl.tensor` to create a sparse tensor, at least if the data isn't dense. That is, intermediate tensors created in an algorithm should still be sparse.

Actually I think I mentioned this before, but I think simply making `tl.tensor` automatically create a sparse or dense tensor based on the density of the data could solve the sparse/dense mixing issues. For instance, there's no reason the random initial vectors from `parafac` should be sparse. I believe the backend functions should still be the dense versions even if presented with a dense type, for instance, at https://github.com/tensorly/tensorly/pull/91, `kr` with a sparse mask only works properly in the sparse backend, though the arrays being `kr`ed are themselves dense. If anything the backend, or even the sparse library itself, could dispatch separately for dense arrays if necessary.

hameerabbasi:
So, on PyData/Sparse, after another Sparse format called CSF has been added, the plan is to integrate it with SciPy and deprecate scipy.sparse.

I’m currently travelling and on a cell phone right now but feel free to ping me if you need more details.

JeanKossaifi:
Thanks @asmeurer and @hameerabbasi!

> the plan is to integrate it with SciPy and deprecate scipy.sparse.

Great! It might be good/easier ot set the sparse backend directly along the main backend.. 
I went in that direction in the latest commit in this PR (just lookup the current backend and use the corresponding sparse backend).

> I'm not clear what you mean by "rather than dispatching them". You mean we don't wrap the various functions with a context manager/decorator so that tensorly.sparse.parafac is different from tensorly.parafac? I think there were some thread safety concerns here, but others should comment on this.

I just mean not to wrap the backend functions (reshape, tensor, etc).
We can still wrap the algorithms (in fact probably have to so it uses the sparse functions).

So basically the choice is between:

1.  not wrapping the backend functions with `dispatch`, simply use module's getattr (when doing `from tensorly import fun`, fun will **not** be updated post import when changing the backend)
2. wrap the backend functions with `dispatch` so they call the correct backend at call time ( when doing `from tensorly import fun`, fun **will** be updated post import when changing the backend)

1 seems more Pythonic and least surprises the uses. 
One big issue with 2. is that attributes, such as `dtype` will **never** be updated post import (e.g. `from tensorly import float32` will never be updated post import), and in that sense, it is more coherent to adopt the same behaviour for the methods as well, e.g. 1 (unless we can update them directly in globals somehow?)

If there are strong reasons to go for though 2, we *could* then wrap all backend functions to dynamically dispatch and just discourage imports such as `from tensorly import X`, especially for attributes we cannot update. 

> It would be best to merge now. 

I agree, we can fix the issues with this PR, merge it and we can change things later.

>  there's no reason the random initial vectors from parafac should be sparse.  I believe the backend functions should still be the dense versions even if presented with a dense type

I'm not 100% sure what you mean here. Do you mean that the sparse functions should return sparse or dense tensors depending on the input and context?
I don't see how to do this automatically unless we explicitly rewrite the sparse algorithms that require this sort of tradeoffs.

---------

Why do we actually need the thread-local storage for the backend: I don't see a practical application where we would want to change the backend only in one thread? 

I also started some work on a sparse backend for `pytorch` but it might be trickier -- e.g. they don't yet seem to support reshaping or permuting axes for sparse tensors.

asmeurer:
The no wrapping/option 1 seems fine to me (again, presuming there are no thread safety issues which others would need to comment on).

asmeurer:
> I agree, we can fix the issues with this PR, merge it and we can change things later.

What are the issues with this PR? I can help fixing any of them, but I'm not entirely clear what all remains to be done.

> I don't see how to do this automatically unless we explicitly rewrite the sparse algorithms that require this sort of tradeoffs.

I just mean that the sparse backend `tensorly.tensor` should return a dense typed tensor if given dense input. 

This is an issue that can be fixed later, though. For now things work, and this is a relatively minor inefficiency. 

> I also started some work on a sparse backend for pytorch but it might be trickier -- e.g. they don't yet seem to support reshaping or permuting axes for sparse tensors.

That doesn't sound straightforward. Reshaping is how I ended up implementing the matriced-khatri rao https://github.com/tensorly/tensorly/pull/85/files?utf8=✓&diff=unified#diff-c2a8e1dc4b7e45c4ba50816dd0d5c13bR215. But maybe it can be done in another way, with fancy slicing or something.

hameerabbasi:
> Why do we actually need the thread-local storage for the backend: I don't see a practical application where we would want to change the backend only in one thread?

There might be an application where:

a. Tensorly is used, with multiple back-ends.
b. Those back ends have to be switched/used across multiple threads.

NumPy and SciPy release the GIL, so using multiple threads can be beneficial sometimes.

Of course, we can always document that things aren’t thread safe, and that’s fine too. But if we have the ability to make things thread safe, that would be nice as well, especially as we DO give the ability to switch the back end in real time.

Thread local storage is a nice way to do this. Another even nicer way introduced in Py37 (I’m not sure about back ports) is context variables.

If we do `from tensorly import method` that won’t be thread safe unless `method` itself does the dispatch and isn’t a static pointer.

Doing `import tensorly as tl; tl.method` will be thread safe if using the `__getattr__` method.

I’d recommend a decorator for all functions that takes in the main function and performs dispatch, this will be thread safe regardless of the method used.

JeanKossaifi:
> a. Tensorly is used, with multiple back-ends.
b. Those back ends have to be switched/used across multiple threads.

I'm not sure there are actual/practical case where we'd need this but I agree: since we allow dynamically changing the backend, we might as well make it safe.

> If we do from tensorly import method that won’t be thread safe unless method itself does the dispatch and isn’t a static pointer.

The only issue is the dispatch is that it will not be doable for any of the attributes (e.g. dtypes) and therefore the behaviour will not be consistent.  Also, Python discourages `from tensorly import method` and encourages just importing the module, which will be thread safe in any case. 

I guess we can go for the dynamical dispatch and just warn users regarding attributes. Unless you know of any way to dynamically update the attributes once imported?

hameerabbasi:
> I guess we can go for the dynamical dispatch and just warn users regarding attributes. Unless you know of any way to dynamically update the attributes once imported?

This is about exactly the same as having `method` do the dispatch based on the selected back-end.

JeanKossaifi:
> This is about exactly the same as having method do the dispatch based on the selected back-end.

I'm not sure what you mean. In the case of a method we wrap it with `dispatch` so that calling `method(*args, **kwargs)` is equivalent to `current_backend.method(*args, **kwargs)`. This redirection is done at call time (also this introduces an overhead _at each call_)... However, for _attributes_, there is no call, e.g. once we import `from tensorly import float32`, I don't see how to update this when the backend is changed. There is no actual function call, and the variable `float32` is actually bound with the corresponding attribute..

hameerabbasi:
> (also this introduces an overhead at each call)

So would a module-level `__getattr__`.

> and the variable `float32` is actually bound with the corresponding attribute..

Ah, yes. I don't know a way around that particular problem other than documenting it.

asmeurer:
I agree that documentation is the best way to go here. It shouldn't be too surprising of a gotcha that if you import a name directly then it won't be updated with a backend change.

mcg1969:
Hey folks, I'm really looking forward to seeing this merged, in particular so we can see the sparse tensor support merged as well. I'm working with some folks who are interested in working with this in a conda-based environment, and while I can certainly package up a fork of the repo, I'd much rather see this in master first. So is there anything I can do to help ferry this long; break a tie in decisionmaking perhaps?

scopatz:
Yes, I think I am happy with that. @asmeurer @hameerabbasi?

asmeurer:
It sounds good to me.

JeanKossaifi:
Added back the dispatching. 
Does anyone want to have a quick look before merging?

asmeurer:
Here is the diff against the sparse branch https://github.com/Quansight/tensorly/compare/sparse...JeanKossaifi:backend-manager. It also includes some changes from master. 

Everything looks fine to me. It looks like it is mainly renaming a `set_backend` to `backend_context` and moving stuff to `__init__.py` (IMO it's better to not have too much code in `__init__.py` but I don't feel strongly about it).

JeanKossaifi:
Yes -- in the end adding a backend manager just would make things more complex, yours and @jcrist solution is best. 

The only reason to move some code to the `__init__` is to centralize all the import/registration logic in one place to make it easier to read (and adding the module `__getattr__` and `__dir__`).

So I can merge this and we can change things as needed going forward. In particular proper testing for the sparse backend. In hindsight, I actually wonder whether just overriding the `sys.module[__name__].__class__` isn't actually neater than the module __getattr__ and __dir__ (it's the same for all Python versions, and inheritance means we don't have to add hack the dir to add imported functions)?

tensorly/backend/core.py
```diff
@@ -0,0 +1,689 @@
...
+        name : str
+            The method name.
+        func : callable
+            The method
+        """
+        setattr(cls, name, staticmethod(func))
+
+    @property
+    def int64(self):
+        raise NotImplementedError
```
hameerabbasi: I think this should be `NotImplementedError()` (note the extra parantheses). We're raising an instance of the class, not the class itself.

I would much prefer that we use the [`abstractmethod` decorator](https://docs.python.org/3.7/library/abc.html#abc.abstractmethod) here instead, and made `Backend` into an [`ABC`](https://docs.python.org/3.7/library/abc.html#abc.ABC). However, I'm fine with just raising an error as well.

tensorly/tests/test_backend.py
```diff
@@ -1,47 +1,124 @@
...
+        assert tl.get_backend() == 'numpy'
+        assert isinstance(tl.tensor([1, 2, 3]), np.ndarray)
+        assert isinstance(T.tensor([1, 2, 3]), np.ndarray)
+        assert tl.float32 is T.float32 is np.float32
+
+    # Reset back to initial backend
+    assert tl.get_backend() == toplevel_backend
+
+    # Set not in context manager
+    try:
```
hameerabbasi: Is the `try` needed here? Shouldn't the test just fail if an exception is raised anyway?

asmeurer:
Python automatically instantiates the exception when you raise a class. Instantiation is only required if you add a message.

tensorly/tests/test_backend.py
JeanKossaifi: Probably since we have `torch = pytest.importorskip('torch')` at the beginning.

JeanKossaifi:
Thanks @asmeurer @hameerabbasi, removed the try.

Any other comment or should this be merged?

asmeurer:
No comments from me.

hameerabbasi:
I'm okay with how this is too.

JeanKossaifi:
Great, merging!

Thanks all for this great contribution @jcrist @scopatz @asmeurer @hameerabbasi!

scopatz:
Woohoo :+1:

PR #96, opened 2019-01-24
Make parafac() work with complex-valued tensors

maximeguillaud:
The current implementation of PARAFAC currently does not work properly with complex-valued tensors (it converges to a solution that has high reconstruction error). This fix enables a reliable convergence to a solution with low reconstruction error.

### DIFF
--- a/tensorly/backend/numpy_backend.py
+++ b/tensorly/backend/numpy_backend.py
@@ -123,7 +123,7 @@ def norm(tensor, order=2, axis=None):
     if order == 1:
         return np.sum(np.abs(tensor), axis=axis)
     elif order == 2:
-        return np.sqrt(np.sum(tensor**2, axis=axis))
+        return np.sqrt(np.sum(np.abs(tensor)**2, axis=axis))
     else:
         return np.sum(np.abs(tensor)**order, axis=axis)**(1/order)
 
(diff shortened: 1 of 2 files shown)


asmeurer:
This will need to be adjusted when https://github.com/tensorly/tensorly/pull/85 is merged.

JeanKossaifi:
Thanks @maximeguillaud for identifying and fixing the issue!

As @asmeurer pointed out, this currently won't work with all backends (AFAIK pytorch doesn't yet support complex tensors).

The easiest solution is probably to update the transpose function in the numpy backend directly, something like:
```python
def transpose(self, tensor, *args, **kwargs):
        # Handle complex matrices: use the conjugate of the transpose
        if (not args) and (not kwargs) and (self.ndim(tensor) == 2):
            return tensor.conj().transpose(*args, **kwargs)
        else:
            return tensor.transpose(*args, **kwargs)
```
(`transpose` is sometimes used to permute the axis which is why I checked for args or kwargs).

asmeurer:
I don't think that's a good idea. transpose isn't necessarily going to mean conjugate transpose in every algorithm. And as you noted, it is sometimes just used to permute axes. I think the current fix here, to be explicit in the algorithm, is best. 

I would add `conj` as a backend function (instead of using the method). If a backend doesn't support complex dtypes, then `conj` can just be a no-op for that backend, since every array from that backend will be real anyway.

JeanKossaifi:
Let's go with adding a `conj` method to the backend then.
@maximeguillaud want to add this?

JeanKossaifi:
Added  in #106.

PR #89, opened 2018-12-10
Sparse docs

asmeurer:
Here are some documentation for sparse arrays. 

This is based on the branch from https://github.com/tensorly/tensorly/pull/85, so that should be reviewed and merged first. It also references the robust_pca work which is forthcoming, and it also references the notebook from the pull request https://github.com/JeanKossaifi/tensorly-notebooks/pull/6.

PR #97, opened 2019-02-18
Fix the sparse backend

asmeurer:
It seems the latest master has broken the sparse backend. 

I still need to push some more fixes, although feel free to push your own fixes.

With this branch, if you do `import tensorly.contrib.sparse as tlsp`, `tlsp.tensor` still points to `numpy.tensor`, because `numpy` is the default backend. Is that the desired behavior?


asmeurer:
Oh my bad I fixed the imports incorrectly.

asmeurer:
I am adding some tests for the sparse backend here. This required making it available in the main backends, so that `TENSORLY_BACKEND=sparse` works. To simplify this, I renamed it to just `sparse`.

asmeurer:
I have had to reduce the ranks in some of the tucker tests so that they can be tested with the sparse backend. Unfortunately, `scipy.sparse.linalg.eigsh` raises an exception if the number of eigenvalues is >= N. I don't know if this can be fixed. If you prefer, I can put the tests back and skip them in the sparse backend for now. 

I have also parameterized all the tucker tests with init='random' and init='svd'.

asmeurer:
This is ready for review. I plan to add more tests for the sparse backend in the respective pull requests for parafac and robust pca.

JeanKossaifi:
Thanks @asmeurer! I'll make a review asap but in the meantime, I can go ahead and fix the imports?

Adding sparse as a main backend does fix the testing issue but I'm not sure it's the best way forward. Going forward I'd like want sparse support for more/all backends. Why not adapt/wrap (or even duplicate..) the tests for the sparse module?

tensorly/tenalg/n_mode_product.py
```diff
@@ -59,12 +57,12 @@ def mode_dot(tensor, matrix_or_vector, mode):
             raise ValueError('Can only take n_mode_product with a vector or a matrix.'
                              'Provided array of dimension {} not in [1, 2].'.format(T.ndim(matrix_or_vector)))
 
-        res = T.dot(matrix_or_vector, unfold(tensor, mode))
+        res = T.dot(matrix_or_vector, tl.unfold(tensor, mode))
```
JeanKossaifi: Inside the package I tried to just use relative imports. This shouldn't make any difference though -- is there any reason to change? I guess there's also no need to import both `tensorly` and `tensorly.backend`

Since we went with dynamical dispatch, `unfold` and `tl.unfold` should both refer to the correct version.

asmeurer: In general, if you import a wrapped function from the `tensorly.contrib.sparse` module (like `from tensorly.contrib.sparse.decomposition import partial_tucker`), then in that wrapped function everything is switched to calling the sparse backend. 

But I think maybe it doesn't matter for `unfold` since it isn't actually a backend function. I think I made this change while trying to fix something that was ultimately caused by something else. Everything should be tested now, so if the tests still pass with this change reverted that's fine.

asmeurer:
Feel free to push changes to this branch, or start a new one based off this. There are tests now so if anything breaks they should show it.

> Why not adapt/wrap (or even duplicate..) the tests for the sparse module?

I do have some sparse specific tests, but I also want to test that sparse works at all for the tests that are already there, especially for tucker since I didn't actually work on this method so I have very little experience with it. And the backend tests are good for testing the sparse backend itself. I mainly intend to use `test_sparse.py` to make sure the algorithms are sparse-safe. 

> Adding sparse as a main backend does fix the testing issue but I'm not sure it's the best way forward.

This was the minimal thing that could be done to make it work without messing with the backend logic. The way it is now, it's impossible to have a backend that isn't in the `tensorly.backends` module. 

To clarify, the name `sparse` of the backend is in reference to the name of the library, not "sparse" in general. It's a bit unfortunate. We could rename it to PyDataSparse or something like that if it helps with clarity (`numpy.sparse` that was there before is very misleading).

asmeurer:
I've removed the unnecessary change to `tl.fold`, etc. for non-backend functions.

JeanKossaifi:
> This was the minimal thing that could be done to make it work without messing with the backend logic. The way it is now, it's impossible to have a backend that isn't in the `tensorly.backends` module.

I think that's how it should be. Though we can always explicitly register a new backend.

> To clarify, the name `sparse` of the backend is in reference to the name of the library, not "sparse" in general. It's a bit unfortunate. We could rename it to PyDataSparse or something like that if it helps with clarity (`numpy.sparse` that was there before is very misleading).

The way I see it, the backend refers to NumPy, MXNet, etc.
Sparse support is not a backend but an additional feature of each backend. In NumPy's case, since i) the existing sparse support is lacking ii) sparse is going to be part of scipy anyway, we use `sparse` for the sparse support. In short I see `sparse` as part of the numpy backend, I don't really want users to set the backend to `sparse` (at least in the current design).

Generally I am open to a refactoring of the way the backend system works/ the design of it, if we find a better way to do it. With the current code though I'd rather specialise the existing code for sparse backend by creating a new version than compromise on speed/readability for the dense backend. Could we wrap the tests the way we do for the main functions?

asmeurer:
I think that support for mixing sparse.COO and np.array should be removed, at least for the time being. At https://github.com/tensorly/tensorly/pull/85 for instance, unless I can figure out how to make the sparse mttkrp faster, I will need to have a separate sparse and dense version. That means that only the sparse backend `parafac` will work properly on sparse inputs, because the dense `mttkrp` does the einsum in a way that blows up if the pairwise dimensions are too large. 

I think we discussed in another PR that we could make the sparse backend `tl.tensor` automatically create a dense array if the input is dense. This works in theory, but it requires a lot of testing, because there are a lot of functions that break when you mix sparse and dense arrays. Sometimes this is on purpose (see the change I made here to `arange` and the linked issues). Sometimes it's because of a lack of support (e.g., `np.dot()` doesn't work on sparse arrays; if you want to dot a sparse and a dense array have to use sparse.dot(dense) in that order).  I'm not saying this isn't possible. It should be an aim to do this, because representing sparse arrays densely is slow and wastes memory. But it isn't a trivial thing to do, and each algorithm and backend function will require special care to make sure it works.

So for now at least, I think it is best to really treat `sparse` as a separate backend. sparse backend functions may work on numpy arrays, but there may also be corner cases where they do not. 

By the way, I'm not clear what the timeline is for `sparse` being merged into `scipy`. Given how conservative scipy's development is, I imagine that could be a slow process. @hameerabbasi would have to comment on that.

JeanKossaifi:
I see your point.

Regarding #85, for this type of cases, we could have a separate function for sparse and dense. 

Having `sparse` as a separate backend also brings it's own issues --e.g. harder to specialize algorithms to support sparse, etc.  For instance, how would we have two function (one for sparse, one for dense) if we consider sparse as a normal backend?

I think the easy way forward is just to enforce that sparse tensors should --for now-- be used with functions from the sparse submodule.. We then just wrap the dense functions if no conversion is needed, and adapt the function where required.

dswalter:
Is there anything that requires help with these associate sparse PR's? Additional tests that would be useful or documentation that I could lend a hand on?

asmeurer:
I'm waiting for comments from @JeanKossaifi. If you want to help with any of the remaining tasks let me know (you can make a PR to the branch on the Quansight fork). I have a limited number of hours that I'm able to devote to this project.

The tasks I'm aware of are:

- Once this is merged, I'd like to add tests to the other pull requests, particularly some tests similar to the one I added here to make sure they algorithms stay sparse safe.  

- For https://github.com/tensorly/tensorly/pull/85 the mttkrp function needs to be factored out and split into sparse and dense versions (see the discussion).

- https://github.com/tensorly/tensorly/pull/89 may require some minor updates 

Those are all for other sparse related pull requests. I don't know what remains to be done for this one. 

Beyond that, it depends on @JeanKossaifi's comments as to what else needs to be done.

JeanKossaifi:
Hi @asmeurer, @dswalter, 

I was away for a few days with no internet. I have started on factoring out mttkrp for #85 when working on it, I can push this, meant to do this earlier. I think for now it might be OK to keep just one version (using the n-mode product). 

I mainly have one comments from above: having sparse as a separate backend brings it's own issues --e.g. harder to specialize algorithms to support sparse, etc. For instance, how would we have two function (one for sparse, one for dense) if we consider sparse as a normal backend?

I think the easy way forward is just to enforce that sparse tensors should --for now at least-- be used with functions from the sparse submodule.. We then just wrap the dense functions if no conversion is needed, and adapt the function where required. We can explicitely wrap the tests as well maybe?

asmeurer:
@JeanKossaifi what would you like to do with this branch? It would be good to know before I attempt any merge conflict resolution.

JeanKossaifi:
Just finished with my ICCV deadline, back on this.

Based on our discussions, I think it might be best to keep a separate `tensorly.sparse` submodule. This allows to easily customize algorithms for the sparse backend without making any changes to existing code, works for all backends and allows to mix dense and sparse without changing the backend (not explicitely at least).

We can wrap tests for now and skip as needed?

As I mentioned, I'd like to (as soon as I have time) start adding (some) sparse support for other backends too -- would be great to have some of the sparse algorithms work on GPU too.

asmeurer:
That sounds fine. My main concern is for things to be tested at all, so that we don't have regressions. 

As far as this branch goes, you may want to check which commits are still relevant and cherry-pick them separately. That may be easier than fixing the merge conflicts if there aren't many.

JeanKossaifi:
I went through the changes -- these were already merged in the other PRs, the only difference is whether to add sparse as a separate backend. Since we are keeping it with NumPy for now,  closing this.

PR #102, opened 2019-03-07
Sparse parafac

JeanKossaifi:
Refactors mttkrp out of parafac on top of #85

### DIFF
--- a/tensorly/decomposition/candecomp_parafac.py
+++ b/tensorly/decomposition/candecomp_parafac.py
@@ -88,7 +88,6 @@ def initialize_factors(tensor, rank, init='svd', svd='numpy_svd', random_state=N
             return [tl.abs(f) for f in factors]
         else:
             return factors
-
     elif init == 'svd':
         try:
             svd_fun = tl.SVD_FUNS[svd]
@@ -118,7 +117,8 @@ def initialize_factors(tensor, rank, init='svd', svd='numpy_svd', random_state=N
 
 
 def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd', tol=1e-8,
-            orthogonalise=False, random_state=None, verbose=False, return_errors=False):
+            orthogonalise=False, random_state=None, verbose=False,
+            return_errors=False, non_negative=False):
     """CANDECOMP/PARAFAC decomposition via alternating least squares (ALS)
 
     Computes a rank-`rank` decomposition of `tensor` [1]_ such that,
@@ -145,7 +145,8 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd', tol=1e-8,
         Level of verbosity
     return_errors : bool, optional
         Activate return of iteration errors
-
+    non_negative : bool, optional
+        Perform non_negative PARAFAC. See :func:`non_negative_parafac`.
 
     Returns
     -------
@@ -160,10 +161,14 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd', tol=1e-8,
     .. [1] tl.G.Kolda and B.W.Bader, "Tensor Decompositions and Applications",
        SIAM REVIEW, vol. 51, n. 3, pp. 455-500, 2009.
     """
+    epsilon = 10e-12
+
     if orthogonalise and not isinstance(orthogonalise, int):
         orthogonalise = n_iter_max
 
-    factors = initialize_factors(tensor, rank, init=init, svd=svd, random_state=random_state)
+    factors = initialize_factors(tensor, rank, init=init, svd=svd,
+                                 random_state=random_state,
+                                 non_negative=non_negative)
     rec_errors = []
     norm_tensor = tl.norm(tensor, 2)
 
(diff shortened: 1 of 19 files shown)


JeanKossaifi:
@asmeurer just added the refactoring out of mttkrp on top of your changes.

Given all your contribution I added you to the authors list there too, hope you don't mind -- feel free to amend should you wish. 
I think it's ready to merge unless you have any comments.

asmeurer:
I'll take a look. 

Is it possible to add sparse-only tests in this branch? I was hoping to add a least one test a la https://github.com/JeanKossaifi/tensorly-notebooks/blob/7179c121cc2f21215f9e14cd1b452f8576683c48/07_pydata_sparse_backend/sparse_parafac.ipynb to act as a smoke test to keep the algorithm sparse-safe.

JeanKossaifi:
Agreed, the test is missing. By smoke test, do you mean to just check that the type of the factors is sparse?

asmeurer:
I mean using a tensor that would lead to a memory error on most machines if it were incorrectly densified, either through misuse of calling `np.array` on a sparse dense array, or from a bad implementation of mttkrp or some other part of the algorithm. The notebook at https://github.com/JeanKossaifi/tensorly-notebooks/blob/7179c121cc2f21215f9e14cd1b452f8576683c48/07_pydata_sparse_backend/sparse_parafac.ipynb has shape (1000, 1001, 1002, 100), which would take 802 GB in dense memory, and the Khatri-rao of the initial vectors would also take up a lot of memory, e.g., using the old mttkrp routine.

asmeurer:
Here is the diff against #85 https://github.com/Quansight/tensorly/compare/sparse-parafac...JeanKossaifi:sparse_parafac?expand=1

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -7,9 +7,10 @@
 from ..kruskal_tensor import kruskal_to_tensor
 from ..tenalg import khatri_rao
 
-# Author: Jean Kossaifi <jean.kossaifi+tensors@gmail.com>
-# Author: Chris Swierczewski <csw@amazon.com>
-# Author: Sam Schneider <samjohnschneider@gmail.com>
+# Authors: Jean Kossaifi <jean.kossaifi+tensors@gmail.com>
+#          Chris Swierczewski <csw@amazon.com>
+#          Sam Schneider <samjohnschneider@gmail.com>
+#          Aaron Meurer
```
asmeurer: You can add my email here too (asmeurer@gmail.com)

tensorly/tenalg/__init__.py
```diff
@@ -3,7 +3,8 @@
 operations such as khatri-rao or kronecker product, n-mode product, etc.
 """ 
 
-from .n_mode_product import mode_dot, multi_mode_dot
+from .n_mode_product import (mode_dot, multi_mode_dot,
+                             unfolding_dot_khatri_rao)
```
asmeurer: Maybe it should also be added to the sparse backend, since it is known to be sparse safe.

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -231,16 +220,15 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd', tol=1e-8,
...
         if tol:
+            # ||tensor - rec||^2 = ||tensor||^2 + ||rec||^2 - 2*<tensor, rec>
             # This is ||kruskal_to_tensor(factors)||^2
             factors_norm = tl.sum(tl.prod(tl.stack([tl.dot(tl.transpose(f), f) for f in factors], 0), 0))
             # mttkrp and factor for the last mode. This is equivalent to the
             # inner product <tensor, factorization>
             iprod = tl.sum(mttkrp*factor)
-            # Subtract iprod from each term to avoid loss of significance
-            rec_error = tl.sqrt(tl.abs((norm_tensor**2 - iprod) + (factors_norm - iprod))) / norm_tensor
+            rec_error = tl.sqrt(tl.abs(norm_tensor**2 + factors_norm - 2*iprod)) / norm_tensor
```
asmeurer: Why did you undo this? I seem to remember it making a difference in some cases.

tensorly/decomposition/candecomp_parafac.py
JeanKossaifi: Absolutely, wasn't sure you wanted your email there and didn't want to expose it in case :)

tensorly/tenalg/__init__.py
JeanKossaifi: Yes, forgot about this.

tensorly/decomposition/candecomp_parafac.py
JeanKossaifi: I just grouped the 2*iprod to be easier to read as the direct expansion of the squared norm. 
Why would there be a difference?

JeanKossaifi:
I think it's ready to merge. Do you have any comments @asmeurer?

asmeurer:
Should we add a sparse test for the parafac itself? The mttkrp is only part of what wasn't sparse safe before. The norm calculation also wasn't.

asmeurer:
I also still prefer things as they are at https://github.com/tensorly/tensorly/pull/97, because that actually tests all supported functions with sparse, and it makes `sparse` a first class backend. But I don't feel strongly about it, and I'll leave it to you to decide (my main priority is getting the functionality merged).

JeanKossaifi:
Thanks @asmeurer 

> Should we add a sparse test for the parafac itself? 

Absolutely. On its way.

> I also still prefer things as they are at #97, because that actually tests all supported functions with sparse, and it makes sparse a first class backend.

Then there is no need for a `tensorly.sparse` module?

I am not 100% sure which solution is best. As I see it, with having `sparse` as a separate backend:
* **On the plus side**:  
   * No additional submodule (no `tensorly.sparse`)
   * When methods are sparse safe, works transparently

* **On the cons side**:
   * harder to specialize algorithms to support sparse, etc. For instance, how would we have two functions (one for sparse, one for dense) if we consider sparse as a normal backend? In other words, if methods are not sparse safe, we might have to either go for a worse solution but sparse safe, or drop the method all-together.
   * Only works for `sparse` library. This will not work with other libraries (e.g. this doesn't add support for mxnet with sparse array) which feels like a missed opportunity.

Also looking forward to `sparse` being part of Scipy I guess it won't make as much sense then to have sparse as a standalone backend?

I want to make sure that TensorLy remains backend agnostic and that all (or almost all) functionalities work transparently regardless of the backend while keeping the codebase as simple as possible. I am worried the code is becoming increasingly complex to handle all backends, switching to sparse, etc.

Having a separate `tensorly.sparse` submodule allows to easily customize algorithms for the sparse backend without making any changes to existing code, works for all backends and allows to mix dense and sparse without changing the backend (not explicitely at least) and both (dense and sparse) are compatible. This should also be the case with `mxnet`, `pytorch`, etc.

asmeurer:
The main advantage I see from https://github.com/tensorly/tensorly/pull/97 is that all the (working) tests are run with the sparse backend turned on, as opposed to just two tests. So it's better test coverage. The existing tests don't check sparse safety but they do make sure the sparse backend functions work properly together. 

There were a few minor fixes there (mostly to the tests themselves). Is it possible to get that in this branch too?

JeanKossaifi:
Can we not somehow wrap the tests (similar to what we do with the functions)?
As a side note, given that `sparse` itself tests for correctness of the methods, and that we check correctness of algorithms, we could hope there shouldn't be any issue ...

> There were a few minor fixes there (mostly to the tests themselves). Is it possible to get that in this branch too?

Sure - my idea was to push this and move to #97 but happy to do it all in one go -- feel free to add things as well!

asmeurer:
> Can we not somehow wrap the tests (similar to what we do with the functions)?

I think we can. It's a question of default in vs. default out. If we wrap each sparse test it's default in. If we skip tests that don't work in sparse as I have done in #97 it's default out. 

> Sure - my idea was to push this and move to #97 but happy to do it all in one go -- feel free to add things as well!

+1 to merge this PR.

JeanKossaifi:
> I think we can. It's a question of default in vs. default out. If we wrap each sparse test it's default in. If we skip tests that don't work in sparse as I have done in #97 it's default out.

We can wrap tests as we make the required changes to make them sparse friendly (and still skip as needed). Ideally I'd like to start adding (some) sparse support for other backends too -- would be great to have some of the sparse algos. work on GPU too.

> +1 to merge this PR.

Great, merging, thanks @asmeurer

PR #105, opened 2019-03-26
Sparse missing values

JeanKossaifi:


JeanKossaifi:
@asmeurer this adds the changes I discussed in #91, any comments?

asmeurer:
No comments on your new commits, as long as everything still works sparse safe. We should add a test here, probably based on the one from the missing values notebook at https://github.com/JeanKossaifi/tensorly-notebooks/pull/6/files.

asmeurer:
Also maybe we should add the reference from https://github.com/tensorly/tensorly/pull/105/commits/218bcac112f1535f5d0843187916fb4977be88f5 to the docstring.

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -148,6 +147,11 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd', tol=1e-8,
         Activate return of iteration errors
     non_negative : bool, optional
         Perform non_negative PARAFAC. See :func:`non_negative_parafac`.
+    mask : ndarray
+        array of booleans with the same shape as ``tensor`` should be 0 where
+        the values are missing and 1 everywhere else. Note:  if tensor is
+        sparse, then mask should also be sparse with a fill value of 1 (or
+        True).
```
asmeurer: FYI this will require some extra work if you ever support a sparse library that doesn't support nonzero fill values the way that `sparse` does.

tensorly/backend/core.py
```diff
@@ -608,10 +608,14 @@ def kr(self, matrices):
...
                 raise ValueError('All matrices should have the same number of columns.')
+
             a = self.reshape(res, (s1, 1, s2))
             b = self.reshape(e, (1, s3, s4))
             res = self.reshape(a * b, (-1, n_col))
-        return res
+        
+        m = self.reshape(mask, (-1, 1)) if mask is not None else 1
+        
+        return res*m
```
asmeurer: Did this change? I thought the mask had to be incorporated into the computation in order to stay sparse safe in some cases.

asmeurer: That was the whole point of making `mask` an argument of the function, rather than having the caller multiply the result.

JeanKossaifi: Yes, true, if this is sparse safe, `mask` doesn't need to be in the khatri-rao / kronecker.
Why would it not be sparse-safe -- It's just an element-wise multiplication with broadcasting?

asmeurer: I think there were cases where the kr itself was large but `kr*mask` was sparse (`mask` is assumed to be sparse). In the parafac code, doesn't it build the full tensor from the factors?

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -197,8 +207,10 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd', tol=1e-8,
             for i, factor in enumerate(factors):
                 if i != mode:
                     pseudo_inverse = pseudo_inverse*tl.dot(tl.conj(tl.transpose(factor)), factor)
-            
-            #factor = tl.dot(unfold(tensor, mode), khatri_rao(factors, skip_matrix=mode).conj())
+
+            if mask is not None:
+                tensor = tensor*mask + tl.kruskal_to_tensor(factors, mask=1-mask)
```
asmeurer: Yeah, here. `kruskal_to_tensor` will be the full dense shape of `tensor` (`factors` are dense in general).

tensorly/kruskal_tensor.py
```diff
@@ -36,12 +41,15 @@ def kruskal_to_tensor(factors, weights=None):
 
     There are other possible and equivalent alternate implementation, e.g.
     summing over r and updating an outer product of vectors.
+
     """
     shape = [T.shape(factor)[0] for factor in factors]
     if weights is not None:
-        full_tensor = T.dot(factors[0]*weights, T.transpose(khatri_rao(factors[1:])))
+        full_tensor = T.dot(factors[0]*weights,
+                             T.transpose(khatri_rao(factors, skip_matrix=0)))
```
asmeurer: Also shouldn't the mask be passed through here?

tensorly/backend/core.py
JeanKossaifi: We changed that part -- instead of using the khatri-rao we use ``tensorly.tenalg. unfolding_dot_khatri_rao`` so shouldn't be an issue in CP. Actually might be worth pulling out of the function for kr below and either put it back in the computation here (I think you had added this [here](https://github.com/tensorly/tensorly/pull/91/files/429b4a767b8e4085763b2da758a934d4012fd8b3#r265118645)  ).

asmeurer: I don't remember all the details of this. If the mask example notebook works then I guess the algorithm is working fine.

PR #107, opened 2019-04-10
Add sparse docs

JeanKossaifi:
Update on top of #89


doc/user_guide/sparse_backend.rst
```diff
@@ -0,0 +1,149 @@
...
+   1.3858051300048828
+
+Note that the decomposition takes much longer when using the sparse variant.
+
+   >>> from tensorly.contrib.sparse.decomposition import parafac as sparse_parafac # The sparse version
+   >>> t = time.time(); sparse_factors = sparse_parafac(tensor, 5, init='random'); print(time.time() - t)
+   14.053689002990723
+
+However, there can be advantages to using the sparse variant. It is currently
+required when using ``init='svd'`` to make TensorLy use the sparse SVD
```
asmeurer: I had thought it was also required to use a mask, but it seems the code has changed https://github.com/tensorly/tensorly/pull/105/files#r274059484. Was the sparse `kr` with a mask not implemented correctly?

asmeurer:
Thanks for this. I think all that's left is robust PCA and the notebooks https://github.com/JeanKossaifi/tensorly-notebooks/pull/6.

doc/user_guide/sparse_backend.rst
JeanKossaifi: I'm not sure I get this -- are you saying we should specify the mask only works for sparse backend?

JeanKossaifi:
Thanks for the changes!
For Robust PCA I think there is just [this](https://github.com/tensorly/tensorly/pull/90#discussion_r273892706) change to add, the rest is already merged from other places.

I'll move on to notebook after this.

doc/user_guide/sparse_backend.rst
asmeurer: The version I originally wrote only worked for the sparse wrapped `parafac`. I don't know if that is still the case after the changes that were made to `kr` and friends.

Issue #138, opened 2019-11-15
Option for starting vector in partial SVD


lan496:
#### Is your feature request related to a problem? Please describe.
I find `tensorly.partial_SVD` does not return the same result for the very same matrix.
For example, the below snippet gives AsserionError.
I think the reproducibility for this function enhance is desired.
```
import numpy as np
import scipy

from tensorly import partial_svd


def is_equal_partial_SVD(matrix):
    U1, S1, V1 = partial_svd(matrix, n_eigenvecs=2)
    U2, S2, V2 = partial_svd(matrix, n_eigenvecs=2)
    assert(np.array_equal(U1, U2))
    assert(np.array_equal(S1, S2))
    assert(np.array_equal(V1, V2))


def is_equal_eigsh(matrix, v0):
    S1, V1 = scipy.sparse.linalg.eigsh(np.dot(matrix.T.conj(), matrix), k=2, which='LM', v0=v0)
    S2, V2 = scipy.sparse.linalg.eigsh(np.dot(matrix.T.conj(), matrix), k=2, which='LM', v0=v0)
    assert(np.array_equal(S1, S2))
    assert(np.array_equal(V1, V2))

matrix = np.random.rand(10, 10)
v0 = np.random.rand(10)
is_equal_eigsh(matrix, v0)  # -> passed
is_equal_partial_SVD(matrix)  # -> assertion failed
```

#### Describe the solution you'd like
It seems the current implementation in tensorly gives a random vector for a starting vector in `scipy.sparse.linalg.eigsh`, and it causes the current behavior.
The option for fixing the starting vector is preferable.

PR #139, opened 2019-11-15
Making partial_svd and tucker reproducible

lan496:
fix https://github.com/tensorly/tensorly/issues/138

- ~~I add the option of a starting vector for `scipy.sparse.linalg.eigsh` in `tensorly.partial_SVD`~~
- I also append a test checking if the same result for partial SVD is obtained.

I've checked my commit passes all the tests by `make test`.

### DIFF
--- a/tensorly/backend/core.py
+++ b/tensorly/backend/core.py
@@ -637,7 +637,7 @@ class Backend(object):
         
         return res*m
 
-    def partial_svd(self, matrix, n_eigenvecs=None):
+    def partial_svd(self, matrix, n_eigenvecs=None, v0=None):
         """Computes a fast partial SVD on `matrix`
 
         If `n_eigenvecs` is specified, sparse eigendecomposition is used on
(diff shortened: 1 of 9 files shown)


lan496:
Sorry, this fix is not compatible with backends other than numpy.
I'll handle it.

lan496:
I've handled multiple backends.
And I've changed the code of Tucker decomposition to be deterministic with random_state.

JeanKossaifi:
Hi @lan496,
Thanks for the issue and PR to fix it!

> Sorry, this fix does not seem to be compatible with backends other than numpy.
> I'll handle it.

I like the attitude! :D

The fix looks great. One thing: rather than inspecting the signature of the SVD fun and creating a partial version, we could just add a **kwargs in all and use `random_state` when relevant?

lan496:
Thank you for your suggestion, @JeanKossaifi .
I add kwargs for all the SVD functions: That's more flexible!

JeanKossaifi:
Thanks @lan496! 

Just a comment regarding the initialization of `v0`: in general initializing correctly these algorithms is not trivial:
* If `random_state is None`, I would initialize `v0` to `None`: it is better to let Arpack choose the init.
* I also do not think that rand is the most appropriate choice, see [discussion in scikit-learn](https://github.com/scikit-learn/scikit-learn/pull/5012) and [doc of Arpack](https://github.com/scipy/scipy/blob/master/scipy/sparse/linalg/eigen/arpack/ARPACK/SRC/dgetv0.f#L225). We should probably at least use a uniform distribution in [-1, 1].

Looks good otherwise!

lan496:
I'm done.
Honestly, I wasn't confident with how to initialize `v0`. So thank you for sharing the related discussion and docs.

JeanKossaifi:
Awesome, thanks @lan496, merging!

PR #146, opened 2019-12-15
[BUGFIX]: scipy.misc.imresize is removed from latest version of SciPy

yngvem:
The documentation did not build with latest version of SciPy since `scipy.misc.imresize` is removed as of SciPy v.1.3.0. However, `scipy.ndimage.zoom`, which can accomplish the same with slightly different syntax, is. 

This pull request only includes a fix to this problem, and contains one new line and one changed line.

### DIFF
--- a/examples/decomposition/plot_image_compression.py
+++ b/examples/decomposition/plot_image_compression.py
@@ -8,7 +8,8 @@ Example on how to use :func:`tensorly.decomposition.parafac`and :func:`tensorly.
 import matplotlib.pyplot as plt
 import tensorly as tl
 import numpy as np
-from scipy.misc import face, imresize
+from scipy.misc import face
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
Great, thanks @yngvem !

Why the addition of the `squeeze` in `to_image` -- Is there a case where it is needed? `zoom` doesn't add a dimension, right?

yngvem:
No, I thought I removed that part! I removed it now.

I added that line when I looked for a bug that arose by the `zoom` function needing a tuple as input instead of a float. Otherwise, the `zoom` function shrinks the channel mode as well as the spatial modes of the image tensor.

JeanKossaifi:
Great, merging! Thanks @yngvem !

PR #147, opened 2019-12-15
Sparse parafac

juliagusak:
Sparse PARAFAC approximates tensor with a sum of low-rank and sparse components. Given rank the low-rank component is represented in Kruskal format, the sparse component has a predefined number of non-zero elements.

### DIFF
--- a/examples/sparse_parafac.ipynb
+++ b/examples/sparse_parafac.ipynb
@@ -1,0 +1,118 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import torch\n",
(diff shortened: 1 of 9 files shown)


JeanKossaifi:
Thanks @juliagusak ! 

There is a merge conflict  --in the meantime I added l2_regularization to CP, but should be easy to fix. Could you add a unit test for sparse_parafac (e.g. just adding a test case in test_parafac for the sparse case), in: https://github.com/tensorly/tensorly/blob/fdb24f477b443554143e7d0707bdfc8b0694177f/tensorly/decomposition/tests/test_candecomp_parafac.py#L15

JeanKossaifi:
Thanks @juliagusak, looks great, merging!

juliagusak:
Thanks @JeanKossaifi !

PR #150, opened 2020-01-01
PARAFAC2 in TensorLy

MarieRoald:
# PARAFAC2 in TensorLy

This pull request adds the PARAFAC2 model to TensorLy. The implementation uses the direct fitting algorithm described in (Kiers et al. 1999) to decompose a tensor or a stack of matrices.

The code was written in pair programming style with @yngvem . It is our first contribution to an open source project, so please let us know if we should change anything with the pull request.

# Model modifications

The PARAFAC2 model described in (Kiers et al. 1999) is on the form \mathcal{X}\_{ijk} = \sum\_r A\_{ir}\ [B_k]\_{jr} C\_{kr}. This form makes it cumbersome to switch between analysing stacks of matrices (as a list) and tensors. The reason for this is that if we have a list of matrices, then the indexing would be \mathcal{X}\_{ijk} = = `X[k][i, j]`. Instead, for this implementation, we have the decomposition on the form \mathcal{X}\_{ijk} = \sum\_r A\_{ir}\ [B_i]\_{jr}\ C\_{kr}, making indexing of lists of matrices the same as the indexing of an nd-array.

# TensorLy source alterations

Here is a list of all new or changed source files and the alterations made in them:

- `doc/modules/api.rst`

   - Added documentation for PARAFAC2

- `tensorly/decomposition/__init__.py`

   - Added PARAFAC2 related functions

- `tensorly/decomposition/candecomp_parafac.py`

   - Added support for init by an existing decomposition.

- `tensorly/decomposition/parafac2.py`

   - New file: functionality related to the PARAFAC2 decomposition

- `tensorly/decomposition/tests/test_parafac2.py`

   - New file: Tests for PARAFAC2

- `tensorly/parafac2_tensor.py`

   - New file: The PARAFAC2 class

- `tensorly/random/__init__.py`

   - Added PARAFAC2 related functions

- `tensorly/random/base.py`

   - Added random PARAFAC2 tensor

- `examples/decomposition/plot_parafac2.py`

   - Added an example where a tensor following the shifted Kruskal tensor model (a special case of the PARAFAC2 model) (Mørup et al. 2008) is decomposed with PARAFAC2. Includes a simple workaround for the sign indeterminacy of PARAFAC2.

# Implementation details

## The PARAFAC2 model class:

We have added a wrapper class for a PARAFAC2 decomposition, similar to the KruskalTensor class. It is a length-3 mapping on the form: `(weights, (A, blueprint_B, C), projections)`. Each `B[i]` matrix is constructed by the following formula: `B[i] = projections[i]@blueprint_B`.

However, we are seldom interested in the `blueprint_B` and projection matrices, we have therefore added the utility function `apply_projections` to `tensorly/parafac2_tensor.py`. It takes a PARAFAC2 tensor as input and outputs a two-tuple on the form (weights, (A, [B[1], ..., B[I]], C)).

Since the PARAFAC2 model can be used to describe a stack of uneven matrices, we have three different functions to construct a dense data representation:

- `parafac2_to_slices`

   - Construct a list, `X_slices` of dense matrices on the form `X_slices[i] = (projections[i]@blueprint_B)@diag(A[i])@C)`

- `parafac2_to_slice`

   - Construct a single matrix from the `X_slices` list.

- `parafac2_to_tensor`

   - Construct a tensor, `X_tensor` with the same formula as above, padding uneven slices with zeros. The shape is given by `(len(X_slices), max(X.shape[0] for X in X_slices), X_slices[0].shape[1])`

In addition, we have the `parafac2_to_unfolded` and `parafac2_to_vec`, which both use `parafac2_to_tensor` and include the zero-padded values.

Finally, we also have the `parafac2_normalise` function, which normalises a PARAFAC2 tensor so that the columns of the factor matrices have unit norm.

## The PARAFAC2 decomposition implementation

The PARAFAC2 decomposition is implemented after the MATLAB code by Rasmus Bro. One small implementation detail here that is different from (Kiers et al. 1999) is that we run the PARAFAC part of the algorithm a customisable number of iterations per projection matrix update (default=5).

For the PARAFAC part of the algorithm, we use the `parafac` function from `tensorly/decomposition/candecomp_parafac.py`. To allow for this, we added the option to initialise the PARAFAC decomposition with precomputed factor matrices. 

## Random PARAFAC2 tensors

To construct a random PARAFAC2 tensor, we first create a random third-order Kruskal tensor with our chosen length for the first and last mode. The second mode, however, we create with length equal to our chosen tensor rank. Then, the random projection matrices are generated by first generating random matrices and decomposing them with a QR decomposition to get orthonormal projection matrices.

# Future improvements

A problem with the PARAFAC2 decomposition is that there is a slice-wise sign indeterminacy. This indeterminacy means that we can change the sign of an element in `A` (e.g. `A[i, r]`) if we also change the sign of the corresponding `B[i, :, r]` factor vector (Helwig 2013). It is therefore common to run a PARAFAC2 decomposition with non-negativity constraints on the mode that the factor matrices change across (in our case, the A-mode). Unfortunately, the PARAFAC implementation in TensorLy currently only supports non-negativity on all modes. Constraining all modes does not make sense for a PARAFAC2 decomposition as the B mode is not computed directly. A possible future improvement may, therefore, be to allow for non-negativity constraints on only specific modes in the PARAFAC decomposition as well as the PARAFAC2 decomposition.

# References

Kiers HA, Ten Berge JM, Bro R. *PARAFAC2—Part I. A direct fitting algorithm for the PARAFAC2 model.* Journal of Chemometrics: A Journal of the Chemometrics Society. 1999 May;13(3‐4):275-94. ([Online link](https://onlinelibrary.wiley.com/doi/abs/10.1002/(SICI)1099-128X(199905/08)13:3/4%3C275::AID-CEM543%3E3.0.CO;2-B))

Mørup M, Hansen LK, Arnfred SM, Lim LH, Madsen KH. *Shift-invariant multilinear decomposition of neuroimaging data.* NeuroImage. 2008 Oct 1;42(4):1439-50.  ([Online link](https://www.sciencedirect.com/science/article/abs/pii/S1053811908007167))

Helwig NE. The special sign indeterminacy of the direct-fitting Parafac2 model: Some implications, cautions, and recommendations for Simultaneous Component Analysis. Psychometrika. 2013 Oct 1;78(4):725-39. ([Online link](https://link.springer.com/article/10.1007/s11336-013-9331-7))


JeanKossaifi:
Hi @MarieRoald, @yngvem !

Thanks for this amazing PR, what a great way to start 2020! :)

I think the Continuous Integration failed because in one of the functions you used tensor.T, which works with NumPy but not the other backends (PyTorch etc): this should be straight-forward to fix.

I will also do a more thorough review but at a first glance it looks great!

Jean

tensorly/decomposition/parafac2.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+    I = len(tensor_slices)
+    J = max(tensor_slice.shape[0] for tensor_slice in tensor_slices)
+    K = tensor_slices[0].shape[1]
+    unfolded = T.zeros((I, J, K))
+    for i, tensor_slice in enumerate(tensor_slices):
+        J_i = len(tensor_slice)
+        unfolded[i, :J_i] = tensor_slice
+    
+    return unfolded
```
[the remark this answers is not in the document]
MarieRoald: I don't remember anymore, but we probably meant to call it padded since that definitely is a more accurate name.

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -0,0 +1,166 @@
...
+    tensor_manual = T.zeros((I, J, K))
+
+    for i in range(I):
+        Bi = T.dot(projections[i], factors[1])
+        for j in range(J):
+            for k in range(K):
+                for r in range(rank):
+                    tensor_manual[i, j, k] += factors[0][i][r]*Bi[j][r]*factors[2][k][r]
+
+    assert_(tl.max(tl.abs(constructed_tensor - tensor_manual)) < 1e-8)
```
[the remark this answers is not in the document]
MarieRoald: This one is weird. If we set the precision in NumPy to single, then it also fails. So I believe that there is some loss of precision and roundoff errors in the PyTorch implementation. Setting the tolerance to 1e-6 made the test succeed for both 32-bit NumPy and PyTorch, so I that is probably the solution here.

MarieRoald:
No problem, the past few months have been hectic for us too. Thank you for your revisions, we have addressed them now :)

Also, there was a minor problem with the test of PARAFAC2 in MXNet and PyTorch, where it failed to recover B[i, :, r] for elements where A[i, r] was low. This is a known problem with PARAFAC2, since then there is little signal from A[i, r] (it is discussed in both section 5.2 and 6.2 of the PARAFAC2 Part I paper by Kiers et al. 1999).

There is still an issue with the TensorFlow backend. It seems like we cannot set tensor elements directly in TensorFlow (looks like a similar problem to the one described here: https://github.com/tensorflow/tensorflow/issues/33131).

JeanKossaifi:
Thanks @MarieRoald, I think we can either i) get rid of the need for indexing (e.g. by padding each factor first and concatenating them) or ii) most simply (though not idea), simply skip the tensorflow test :)

We also skipped the tensorflow test ( for the same reason) in https://github.com/tensorly/tensorly/blob/master/tensorly/contrib/decomposition/tests/test_mps_decomposition_cross.py:

https://github.com/tensorly/tensorly/blob/8746f766fd3aa5df72b211a5d5448372fcef11f7/tensorly/contrib/decomposition/tests/test_mps_decomposition_cross.py#L13-L17

JeanKossaifi:
So I decided to fix the issue altogether. Unfortunately JAX and Tensorflow do not make this as straightforward as it should be. 
To do the assignment, you should now be able to replace:
```python
tensor[i, :length] = slice_
```
with 
```python
tl.index_update(tensor, tl.index[i, :length], slice_)
```
and it should be fine for all backends.

yngvem:
Sorry for taking so long to respond, things are quite hectic for the two of us right now. 

We have fixed the`__setitem__` bug. However it seems like TensorFlow doesn't support the `+=`-like operators. Thus, to enable inplace editing of arrays with the NumPy backend, I see two solutions

 * Create functions for additions, multiplications, etc. in TensorLy, that can perform inplace assignments if possible (akin to `add`, `multiply`, etc. in NumPy, which has an optional out argument)
 * Don't support TensorFlow for PARAFAC2

Alternatively, we can change all inplace modifications to new memory allocations instead.

tensorly/parafac2_tensor.py
```diff
@@ -0,0 +1,511 @@
...
+    
+    # Skip first factor matrix since the rank is extracted from it.
+    for i, factor in enumerate(factors[1:]):
+        current_mode_size, current_rank = T.shape(factor)
+        if current_rank != rank:
+            raise ValueError('All the factors of a PARAFAC2 tensor should have the same number of columns.'
+                             'However, factors[0].shape[1]={} but factors[{}].shape[1]={}.'.format(
+                                 rank, i, T.shape(factor)[1]))
+
+    if weights is not None and len(weights) != rank:
```
JeanKossaifi, line 146: No idea why tensorflow doesn't support this. I guess we can do.

```suggestion
    if weights is not None and tl.shape(weights)[0]  != rank:
```

tensorly/parafac2_tensor.py
```diff
@@ -0,0 +1,511 @@
...
+        raise ValueError('Given factors for a rank-{} PARAFAC2 tensor but len(weights)={}.'.format(
+            rank, len(weights)))
```
JeanKossaifi, line 148:
```suggestion
            rank, tl.shape(weights)[0]))
```

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -0,0 +1,172 @@
...
+    random_parafac2_tensor = random_parafac2(
+        shapes=[(50 + rng.randint(10),60) for _ in range(40)],
+        rank=rank,
+        random_state=rng,
+    )
+    # It is difficult to correctly identify B[i, :, r] if A[i, r] is small.
+    # This is sensible, since then B[i, :, r] contributes little to the total value of X.
+    # To test the PARAFAC2 decomposition in the precence of roundoff errors, we therefore add
+    # 0.01 to the A factor matrix.
+    random_parafac2_tensor.factors[0] += 0.01
```
JeanKossaifi, line 55: Since it is just a test we don't really care about the memory (the factor is small anyway)..
```suggestion
    random_parafac2_tensor.factors[0] = random_parafac2_tensor.factors[0] + 0.01
```

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -0,0 +1,172 @@
...
+def test_parafac2_normalize_factors():
+    rng = check_random_state(1234)
+    rank = 6
+
+    random_parafac2_tensor = random_parafac2(
+        shapes=[(50 + rng.randint(10),60) for _ in range(40)],
+        rank=rank,
+        random_state=rng,
+    )
+    random_parafac2_tensor.weights[0] = 100
```
JeanKossaifi, line 113: I think this should work:
```suggestion
    tl.index_update(random_parafac2_tensor.weights, tl.index[0], 100)
```

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -0,0 +1,172 @@
...
+
+    constructed_tensor = parafac2_to_tensor((weights, factors, projections))
+    tensor_manual = T.zeros((I, J, K), **T.context(weights))
+
+    for i in range(I):
+        Bi = T.dot(projections[i], factors[1])
+        for j in range(J):
+            for k in range(K):
+                for r in range(rank):
+                    tensor_manual[i, j, k] += factors[0][i][r]*Bi[j][r]*factors[2][k][r]
```
JeanKossaifi, line 168: We can also just use index_update here, should be something like:
```suggestion
                    tl.index_update(tensor_manual, index[i, j, k],  tensor_manual[i, j, k] + factors[0][i][r]*Bi[j][r]*factors[2])[k][r]
```

tensorly/tests/test_parafac2_tensor.py
```diff
@@ -0,0 +1,203 @@
...
+    assert_equal(shape, true_shape,
+                 err_msg='Returned incorrect shape (got {}, expected {})'.format(
+                    shape, true_shape))
+    assert_equal(rank, true_rank,
+                 err_msg='Returned incorrect rank (got {}, expected {})'.format(
+                    rank, true_rank))
+
+    # One of the factors has the wrong rank
+    for mode in range(3):
+        false_shape = (len(factors[mode]), true_rank+1)
```
JeanKossaifi, line 33: Another TensorFlow issue ... 
```suggestion
        false_shape = (tl.shape(factors[mode])[0], true_rank+1)
```

JeanKossaifi:
Thanks @yngvem ! No problem at all, it's hectic here too, and I guess we tough for all to adjust to the current situation. Hope you're well, and thanks for the awesome PR :) 

Yes, TensorFlow is causing a lot of problems... I guess we could use index_update to do inplace update where possible. I think the easiest for now is to not to the operation inplace.

I made suggestions in the code where Travis failed -- hope I didn't miss any. They were mostly in the tests so it shouldn't matter if we duplicate memory there :)
In the main code it was mostly just TensorFlow oddities like `len` not working on a tensor..

JeanKossaifi:
Thanks for the fix, looks great! :)
I think the current error is in test_parafac2.py (l.168) -- just an `index` that should be `tl.index`.

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -0,0 +1,172 @@
...
+
+    constructed_tensor = parafac2_to_tensor((weights, factors, projections))
+    tensor_manual = T.zeros((I, J, K), **T.context(weights))
+
+    for i in range(I):
+        Bi = T.dot(projections[i], factors[1])
+        for j in range(J):
+            for k in range(K):
+                for r in range(rank):
+                    tl.index_update(tensor_manual, tl.index[i, j, k],  tensor_manual[i, j, k] + factors[0][i][r]*Bi[j][r]*factors[2])[k][r]
```
JeanKossaifi, line 168: The issue in Travis is that the values of the update is a 2D tensor while we are updating a single value: it is probably the parenthesis closed too early, (`factors[2])[k][r]` instead of `factors[2][k][r])` )
```suggestion
                    tl.index_update(tensor_manual, tl.index[i, j, k],  tensor_manual[i, j, k] + factors[0][i][r]*Bi[j][r]*factors[2][k][r])
```

JeanKossaifi:
One of the errors is a simple misplaced parenthesis ( I commented in the code). For the other, it seems that the second test in `test_parafac2_normalise` fails (also happens locally for me). Could it be numerical error because of testing with float32 instead of float64?

yngvem:
The PARAFAC2 normalization failed since we had an explicit option to perform inplace operations, after removing that option, all non TF backends are working again.

JeanKossaifi:
... it is not a dtype issue ..

JeanKossaifi:
I tried to debug the issue. It seems at least one of the issues comes from the normalisation. The norms seem crazy (some very large (e.g. 1e16), some very small (e.g 1e-13 to 1e-18)), depending on the factor. It most definitely goes out of range and must be causing issues. I also noticed this behaviour with the NumPy backend but weirdly the test doesn't fail there.. I tried some other way to avoid dividing by eps but it only partially helped (the test still fails). 

Also wondering: in `test_parafac2_normalize_factors`, l.115, why do you assert normalized_rec.weights[0] > 1? 

As a side note, some of the test take a while: would reducing the size of the tensor impact the test?

aarmey:
@JeanKossaifi interested in getting this merged. Are there any other remaining issues besides slow tests and `test_parafac2_normalize_factors`? I'm not sure how to commit to an open PR as a third party (file a PR with @MarieRoald?) but am happy to look into this if it'd be helpful...

yngvem:
Sorry for taking so long, we both have too much on our plates, but we're planning to look at it this weekend.

MarieRoald:
Hi! We finally had some time to look at this :)

We located the source of the weight bug when the model was fit with normalisation.  It appears a recent commit that fixed so normalisation no longer was performed inplace introduced this bug. The normalisation created a normalised factor matrix that was discarded immediately and did not update the factor matrices of the decomposition. Thus, making the decomposition norm grow exponentially.

The reason we asserted that weight should be greater than one was that the tensor we decomposed did not have a unique PARAFAC2 decomposition, which threw off the test. We have now changed the tests so that they decompose tensors with a unique PARAFAC2 decomposition, which allows us to check the weight magnitudes.

Also, for speed of tests, we reduced the dimension of the tensor. To ensure the smaller tensor is still uniquely decomposed, we also reduced the tensor rank.

Finally, we changed some bugs with the `tl.index_update` for the TensorFlow backend.

examples/decomposition/plot_parafac2.py
```diff
@@ -0,0 +1,232 @@
...
+
+est_A_normalised = est_A/la.norm(est_A, axis=0)
+est_Bs_normalised = [est_B/la.norm(est_B, axis=0) for est_B in est_projected_Bs]
+est_C_normalised = est_C/la.norm(est_C, axis=0)
+
+B_corr = np.array(est_Bs_normalised).reshape(-1, true_rank).T@np.array(Bs_normalised).reshape(-1, true_rank)/len(est_Bs_normalised)
+A_corr = est_A_normalised.T@A_normalised
+C_corr = est_C_normalised.T@C_normalised
+
+congruence_coefficient = np.mean(np.max(A_corr*B_corr*C_corr, axis=0))
```
cohenjer, line 150: A quick comment on the way you find the optimal permutation. 
You compute (A'A*B'B*C*C), which is a square matrix of "scores", and  you expect the highest scores (except on the diagonal) to give you the permutations.
Actually, you may have in practice a component in A' that is the closest to all those in A, and then what you obtain from np.argmax is not a permutation?
A "clean" way to solve the assignment problem that you have is to use the munkres algorithm (linear assignment problem). An implementation is available in scikit-learn or in scipy.

tensorly/decomposition/parafac2.py
```diff
@@ -0,0 +1,280 @@
...
+    .. math::
+
+        X_{ijk} = \sum_{r=1}^R A_{ir} B_{ijr} C_{kr},
+    
+    with the same constraints hold for :math:`B_i` as above.
+     
+
+    Parameters
+    ----------
+    tensor_slices : ndarray or list of ndarrays
```
cohenjer, line 167: Maybe I can recommend that you further develop in the comments that 
- your code only works for 3d order tensors (unless I am mistaken). although the Parafac2 model exist in any order.
- the slices are "horizontal" (whereas from my experience, most people would intuitively work with "vectical" slices). This is to avoid that a user confuses the output factors A and C with each others.

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -80,6 +81,17 @@ def initialize_factors(tensor, rank, init='svd', svd='numpy_svd', random_state=N
             factors.append(factor)
         return factors
 
+    elif isinstance(init, (tuple, list, KruskalTensor)):
```
cohenjer, line 84: Just a note: the addition of a manual initialization to the CP is also discussed in #162. I think the solution you proposed is in line with the discussions in that PR, but the syntax is a little bit different.

I am not sure about using the "init" keyword to pass the factors; this can be a little bit confusing. I suggested to use a keyword "given" and then to pass the initial factors in another argument. Not sure which is best, yours is more compact for sure.

JeanKossaifi: Just following up -- the consensus is to go with your strategy, it is probably the cleanest. We can add an intermediate function to do more fine-grained initializations later.

examples/decomposition/plot_parafac2.py
yngvem: We updated the example so we now use`scipy.optimize.linear_sum_assignment` to find the optimal permutation of the factors.

tensorly/decomposition/parafac2.py
MarieRoald: We updated the docstring to make this clearer.

tensorly/decomposition/candecomp_parafac.py
MarieRoald: We chose this option because when working on a personal implementation, I found that requiring two inputs led to a cumbersome workflow. However, both are fine, so we can change it. Should we change it here or do you prefer a separate pull request?

yngvem:
The PARAFAC2 convergence criterion was for some reason not based on the relative reconstruction error, we have fixed that now.

We fixed the convergence criterion so that it is based on the relative reconstruction error instead of merely the reconstruction error. Unfortunately, now it doesn't work with TensorFlow anymore since TensorFlow will automatically cast float32 to float64 when you square using `**2` or sum using the builtin `sum` function (which we must do for the PyTorch and MXNet backends). Consequently, TensorFlow fails when we try to compute the relative reconstruction error, as we then divide a float32 singleton tensor by a float64 singleton tensor which we cannot do.

We tried to cast the float64 back to a float32 by writing`norm_tensor = tl.tensor(norm_tensor, **tl.context(tensor_slices[0])`, but that did not work. Thus, we see two solutions:

* Adding a way to cast types in tensorly
* Circumventing the problem using `norm_tensor = tl.to_numpy(norm_tensor)`

What do you think?

JeanKossaifi:
@yngvem @MarieRoald Thanks a lot, it looks great!

For the issue with TensorFlow: the issue was actually created by the norm function (and perhaps a bug in NumPy): if the result is an order-0 tensor I cheat and return a numpy value (it was fixing a lot of other issues when computing the norm of a tensor). Tensorflow seems to have fixed the issue in 896f9b6eebf9f066a6d5e1a7ae12c4d01d3c3997.  

I think it's ready to merge, unless you want to change anything based on that last fix?

JeanKossaifi:
@MarieRoald I fixed the issue with the tensorflow backend in 896f9b6eebf9f066a6d5e1a7ae12c4d01d3c3997, so there is no need for converting to NumPy anymore, it works with just 
```
norm_tensor = tl.sqrt(sum(tl.norm(tensor_slice, 2)**2 for tensor_slice in tensor_slices))
```

tensorly/decomposition/candecomp_parafac.py
JeanKossaifi: I meant that your solution seems to be the best :) 
Let's keep that!

examples/decomposition/plot_parafac2.py
JeanKossaifi: I would like to keep this version somewhere as an example that will work with all backends (without needing to convert to NumPy) -- we can add later in another small PR :)  
We can add a disclaimer about the potential issue.

MarieRoald:
Thank you @JeanKossaifi! We just reverted that change so now it passes all tests again. 

There is nothing else we can think of :)

cohenjer:
Seems all good to me too, thanks for all the modifications :)

JeanKossaifi:
Thanks so much @MarieRoald and @yngvem, this is definitely one of the best PR ever in TensorLy!
Hope this is just the first of many contributions! :)

PR #151, opened 2020-02-24
Add Coupled and Matrix Tensor factorization, with Alternating Least Squares (CMTF-ALS)

IsabellLehmann:
Add the Coupled Matrix and Tensor Factorization (CMTF) for tensorly. Also add pinv to Backend.


tensorly/decomposition/CMTF_ALS
```diff
@@ -0,0 +1,165 @@
...
+    return X
+
+
+def calculate_f(X, Y, A, B, C, V):
+    # calculate value of f in cmtf
+    return 1 / 2 * tl.norm(X - tl.kruskal_tensor.kruskal_to_tensor((None, [A, B, C]))) ** 2 + 1 / 2 * tl.norm(
+        Y - A @ V.T)
+
+
+def factor_match_score_3d(A, B, C, V, A_hat, B_hat, C_hat, V_hat):
```
JeanKossaifi: Maybe we can use clearer names for the variables?
Also, here, perhaps we can have a signature like:

```suggestion
def factor_match_score_3d(kruskal_tensor_true, kruskal_tensor_pred):
```

and the unroll as :
```
shape, rank = tl.kruskal_tensor._validate_kruskal_tensor(kruskal_tensor)
weights_true, factors_true = kruskal_tensor
...
```

tensorly/decomposition/CMTF_ALS
```diff
@@ -0,0 +1,165 @@
...
+
+    R = A.shape[1]
+
+    # calculate weights
+    lambda_ = tl.zeros(R)
+    lambda_hat = tl.zeros(R)
+    alpha_ = tl.zeros(R)
+    alpha_hat = tl.zeros(R)
+    xi = tl.zeros(R)
+    xi_hat = tl.zeros(R)
```
JeanKossaifi: We could get directly kruskal_tensors and further apply kruskal_normalise to them to get the lambda?

tensorly/decomposition/CMTF_ALS
```diff
@@ -0,0 +1,165 @@
...
+
+    # calculate factor match score
+    FMS = tl.zeros(R)
+    for r in range(R):
+        FMS[r] = (1 - tl.abs(xi[r] - xi_hat[r]) / tl.max([xi[r], xi_hat[r]])) * tl.abs(
+            perm[r, r] * tl.dot(V[:, r], V_hat[:, r])) / (tl.norm(V[:, r]) * tl.norm(V_hat[:, r]))
+    return tl.min(FMS)
+
+
+def cmtf_als_for_third_order_tensor(X, Y, R):
```
JeanKossaifi: Perhaps we can make the signature more explicit like:
```suggestion
def cmtf_als_for_third_order_tensor(tensor_3d, matrix, rank):
```
?

tensorly/decomposition/CMTF_ALS
```diff
@@ -0,0 +1,165 @@
...
+    X : tl.tensor
+        3rd order tensor with factor matrices [[A, B, C]]
+    Y : tl.tensor
+        matrix that is coupled with tensor in first mode: Y = A @ V.T
+    R : int
+        rank for CP decomposition of X
+
+    Returns
+    -------
+    A, B, C, V : estimated matrices
```
JeanKossaifi: Could be kruskal_tensor, matrix? 
Or even we return two kruskal tensors, one third order, the other second order, with the same factor first?

tensorly/decomposition/CMTF_ALS
```diff
@@ -0,0 +1,165 @@
...
+    # alternating least squares
+    # note that no rescaling is done since it is not guaranteed that the columns in true matrices have unit norm
+    # note that the order of the khatri rao product is reversed since tl.unfold has another order than assumed in paper
+    for iteration in range(10 ** 4):
+        A = solve_ls(tl.concatenate((tl.tenalg.khatri_rao([B, C]).T, V.T), axis=1).T,
+                     tl.concatenate((tl.unfold(X, 0), Y), axis=1).T).T
+        B = solve_ls(tl.tenalg.khatri_rao([A, C]), tl.unfold(X, 1).T).T
+        C = solve_ls(tl.tenalg.khatri_rao([A, B]), tl.unfold(X, 2).T).T
+        V = solve_ls(A, Y).T
+        f_new = calculate_f(X, Y, A, B, C, V)
```
JeanKossaifi: The code might be clearer if we remove `calculate_f` and solve inline, and use a more informative name for `f`.

tensorly/decomposition/tests/test_CMTF_ALS
```diff
@@ -0,0 +1,162 @@
+import unittest
+import tensorly as tl
+# import tensorly.random
+import cmtf_als
+import numpy as np
+
+
+class MyTestCase(unittest.TestCase):
```
JeanKossaifi: In TensorLy, we typically use `pytest `rather than `unittest`. I tend to prefer having unitest functions rather than packing it all in a big class which tends to be less readable.

tensorly/decomposition/CMTF_ALS
```diff
@@ -0,0 +1,165 @@
...
+
+def rescale_to_unit_frobenius(matrix):
+    # rescale every column of matrix
+    return matrix / tl.norm(matrix, axis=0)
+
+
+def solve_ls(A, B):
+    # solve ||B-AX||^2 (AX = B -> X = A^{-1} @ B)
+    X = tl.pinv(A) @ B
+    return X
```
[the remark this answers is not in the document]
IsabellLehmann: Sure we can add it. I'll name the new file "least_squares", but feel free to rename it, if you have a better name. :)

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+    rank = A.shape[1]
+    FMS = tl.zeros(rank)
+    for r in range(rank):
+        FMS[r] = (1 - tl.abs(xi[r] - xi_hat[r]) / tl.max([xi[r], xi_hat[r]])) * tl.abs(
+            perm[r, r] * tl.dot(V[:, r], V_pred[:, r]))
+
+    return tl.min(FMS)
+
+
+def cmtf_als_for_third_order_tensor(tensor_3d, matrix, rank):
```
JeanKossaifi: Naming is tough.. Assuming people will be familiar with that, we could simply use `cmtf_als_3d` ?
Though I think it might clearest to spell out `coupled_matrix_tensor_decomposition` `coupled_matrix_tensor_3d_decomposition` ?

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+
+    Examples
+    --------
+    A = tl.tensor([[1, 2], [3, 4]])
+    B = tl.tensor([[1, 0], [0, 2]])
+    C = tl.tensor([[2, 0], [0, 1]])
+    V = tl.tensor([[2, 0], [0, 1]])
+    R = 2
+
+    X = tl.kruskal_tensor.KruskalTensor((None, [A, B, C]))  # weights are None
```
JeanKossaifi: We could just write X  = (None, [A, B, C]).

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+    tensor_3d_pred, matrix_pred = cmtf_als_for_third_order_tensor(X, Y, R)
+
+    """
+
+    # initialize values
+    shape_list = tensor_3d.shape
+    s = shape_list + (matrix.shape[1],)
+    A, B, C, V = random_kruskal(s, rank).factors
+
+    if isinstance(tensor_3d, tl.kruskal_tensor.KruskalTensor):
```
JeanKossaifi: This isn't well documented (yet, I'm on it!) but so far in TensorLY we accept both KruskalTensors or just a tuple (weight, factors). So I think a safer logic would be: a) check if it is a full tensor b) if not, check if it is a Kruskal tensor c) otherwise, raise an Error. 

In code:
```python
if tl.is_tensor(tensor_3d):
    x = tl.copy(tensor_3d) # Why the copy?
else:
    shape, rank = tl.kruskal_tensor._validate_kruskal_tensor(tensor_3d) # this will fail if it isn't a valid tuple or KruskalTensor
    X = tl.kruskal_tensor.kruskal_to_tensor(tensor_3d)
```

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+        X = tl.kruskal_tensor.kruskal_to_tensor(tensor_3d)
+    else:
+        X = tensor_3d.copy()
+
+    if isinstance(matrix, tl.kruskal_tensor.KruskalTensor):
```
JeanKossaifi: Same as above.

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # initialize values
+    shape_list = tensor_3d.shape
+    s = shape_list + (matrix.shape[1],)
+    A, B, C, V = random_kruskal(s, rank).factors
+
+    if isinstance(tensor_3d, tl.kruskal_tensor.KruskalTensor):
+        X = tl.kruskal_tensor.kruskal_to_tensor(tensor_3d)
+    else:
+        X = tensor_3d.copy()
```
JeanKossaifi: Why explicitly making a copy?

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+
+    if isinstance(matrix, tl.kruskal_tensor.KruskalTensor):
+        Y = tl.kruskal_tensor.kruskal_to_tensor(matrix)
+    else:
+        Y = matrix.copy()
+
+    objective_old = tl.tensor(np.inf)
```
JeanKossaifi: I would just start the checks for iteration > 0

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+    for iteration in range(10 ** 4):
+        A = solve_least_squares(tl.concatenate((khatri_rao([B, C]).T, V.T), axis=1).T,
+                                tl.concatenate((tl.unfold(X, 0), Y), axis=1).T).T
+        B = solve_least_squares(khatri_rao([A, C]), tl.unfold(X, 1).T).T
+        C = solve_least_squares(khatri_rao([A, B]), tl.unfold(X, 2).T).T
+        V = solve_least_squares(A, Y).T
+        objective_new = 1 / 2 * tl.norm(
+            X - tl.kruskal_tensor.kruskal_to_tensor((None, [A, B, C]))) ** 2 + 1 / 2 * tl.norm(
+            Y - tl.kruskal_tensor.kruskal_to_tensor((None, [A, V])))
+        if tl.abs(objective_new - objective_old) / objective_old <= 1e-8:
```
JeanKossaifi: As per comment above, we could define `objective_old` directly here
```python
if iteration > 0 and tl.abs(objective_new - objective_old) / objective_old <= 1e-8:
    break
objective_old = objective_new
```

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+    # alternating least squares
+    # note that no rescaling is done since it is not guaranteed that the columns in true matrices have unit norm
+    # note that the order of the khatri rao product is reversed since tl.unfold has another order than assumed in paper
+    for iteration in range(10 ** 4):
+        A = solve_least_squares(tl.concatenate((khatri_rao([B, C]).T, V.T), axis=1).T,
+                                tl.concatenate((tl.unfold(X, 0), Y), axis=1).T).T
+        B = solve_least_squares(khatri_rao([A, C]), tl.unfold(X, 1).T).T
+        C = solve_least_squares(khatri_rao([A, B]), tl.unfold(X, 2).T).T
+        V = solve_least_squares(A, Y).T
+        objective_new = 1 / 2 * tl.norm(
```
JeanKossaifi: I guess this is actually the objective function, would error be more explicit? :/
No strong opinion though, just a comment :)

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+    lambda_pred, [a_pred, B_pred, C_pred] = tl.kruskal_tensor.kruskal_normalise(kruskal_tensor_3d_pred)
+    alpha_pred, [A_pred, V_pred] = tl.kruskal_tensor.kruskal_normalise(kruskal_tensor_2d_pred)
+    xi_pred = lambda_pred + alpha_pred
+
+    # take care of permutation of normalized matrices (not mentioned in paper, but necessary)
+    perm = (A.T @ A_pred) * (B.T @ B_pred) * (C.T @ C_pred)
+    order = tl.argmax(perm, axis=1)
+    perm = perm[:, order]  # perm[r,r] now corresponds to \prod_i (a^i_r.T @ ahat^i_r) / (||(a^i_r|| ||ahat^i_r||)
+    V_pred = V_pred[:, order]
+    xi_hat = xi_pred[order]
```
JeanKossaifi: This is a pretty neat piece of code :)

 Might be worth documenting it in a Notes section in the docstring?
e.g. 
```
Notes
------
The factors are first aligned by doing xyz 
```

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,157 @@
...
+    lambda_, [a, B, C] = tl.kruskal_tensor.kruskal_normalise(kruskal_tensor_3d_true)
+    alpha_, [A, V] = tl.kruskal_tensor.kruskal_normalise(kruskal_tensor_2d_true)
+    xi = lambda_ + alpha_
+
+    lambda_pred, [a_pred, B_pred, C_pred] = tl.kruskal_tensor.kruskal_normalise(kruskal_tensor_3d_pred)
+    alpha_pred, [A_pred, V_pred] = tl.kruskal_tensor.kruskal_normalise(kruskal_tensor_2d_pred)
+    xi_pred = lambda_pred + alpha_pred
+
+    # take care of permutation of normalized matrices (not mentioned in paper, but necessary)
+    perm = (A.T @ A_pred) * (B.T @ B_pred) * (C.T @ C_pred)
```
JeanKossaifi: I don't think the `@` is supported by all backends yet, unfortunately :( 
Hence the annoying `tl.dot` everywhere...

tensorly/decomposition/tests/test_cmtf_als.py
```diff
@@ -0,0 +1,58 @@
+import numpy as np
+
+import tensorly as tl
+from ..cmtf_als import (factor_match_score_3d, cmtf_als_for_third_order_tensor)
+from ...testing import assert_array_almost_equal, assert_array_less
+
+
+def test_factor_match_score_3d():
+    # same factors
+    X_true = tl.kruskal_tensor.KruskalTensor((None, [tl.tensor([[4, 1], [0, 2]]), tl.tensor([[0, 1], [3, 5]]),
```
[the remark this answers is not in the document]
IsabellLehmann: I think in the function that gets the parameters, it is shorter if we only give for Kruskal tensors instead of four tuples, isn't it? Especially if I import KruskalTensor in the beginning

tensorly/decomposition/cmtf_als.py
IsabellLehmann: In older code, sometimes I was directly assigning X = Y and when changing X, Y was also changed. This lead to an error sometimes which was difficult to find. Since them, I use the safe variant to just use a copy of the input so that I do not accidentally change it.

tensorly/decomposition/cmtf_als.py
IsabellLehmann: In older code, sometimes I was directly assigning X = Y and when changing X, Y was also changed. This lead to an error sometimes which was difficult to find. Since them, I use the safe variant to just use a copy of the input so that I do not accidentally change it.

tensorly/tenalg/least_squares.py
```diff
@@ -0,0 +1,7 @@
+from .. import backend as T
+
+
+def solve_least_squares(A, B):
```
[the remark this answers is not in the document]
IsabellLehmann: Okay, then we'll leave it :)

tensorly/decomposition/cmtf_als.py
IsabellLehmann: But since Y is not changed, I deleted the copy here. :)

tensorly/decomposition/cmtf_als.py
IsabellLehmann: thanks :)

tensorly/tenalg/least_squares.py
```diff
@@ -0,0 +1,7 @@
...
+    # solve ||B-AX||^2 (AX = B -> X = A^+ @ B), with A^+: pseudo inverse
+    X = T.pinv(A) @ B
```
JeanKossaifi, line 6: This is currently making the test fails as unfortunately other frameworks don't support it...
```suggestion
    X = T.dot(T.pinv(A), B)
```

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,175 @@
...
+    lambda_, [a, B, C] = kruskal_normalise(kruskal_tensor_3d_true)
+    alpha_, [A, V] = kruskal_normalise(kruskal_tensor_2d_true)
+    xi = lambda_ + alpha_
+
+    lambda_pred, [a_pred, B_pred, C_pred] = kruskal_normalise(kruskal_tensor_3d_pred)
+    alpha_pred, [A_pred, V_pred] = kruskal_normalise(kruskal_tensor_2d_pred)
+    xi_pred = lambda_pred + alpha_pred
+
+    # take care of permutation of normalized matrices (not mentioned in paper, but necessary)
+    perm = tl.dot(A.T, A_pred) * tl.dot(B.T, B_pred) * tl.dot(C.T, C_pred)
```
JeanKossaifi, line 71:
```suggestion
    perm = tl.dot(tl.transpose(A), A_pred) * tl.dot(tl.transpose(B), B_pred) * tl.dot(tl.transpose(C), C_pred)
```

JeanKossaifi:
There seems to be a test failing as well - `test_coupled_matrix_tensor_3d_factorization': https://travis-ci.com/github/tensorly/tensorly/jobs/315030995

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,229 @@
...
+            tl.transpose(tl.concatenate((tl.dot(np.diag(lambda_), tl.transpose(khatri_rao([B, C]))),
+                                         tl.dot(np.diag(gamma), tl.transpose(V))), axis=1)),
+            tl.transpose(tl.concatenate((tl.unfold(X, 0), Y), axis=1))))
+        norm_A = tl.norm(A, axis=0)
+        A /= norm_A
+        lambda_ *= norm_A
+        gamma *= norm_A
+        B = tl.transpose(solve_least_squares(tl.dot(khatri_rao([A, C]), np.diag(lambda_)),
+                                             tl.transpose(tl.unfold(X, 1))))
+        norm_B = tl.norm(B)
```
aarmey, line 205: I believe this should be the norm along `axis=0`.

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,229 @@
...
+        B /= norm_B
+        lambda_ *= norm_B
+        C = tl.transpose(solve_least_squares(tl.dot(khatri_rao([A, B]), np.diag(lambda_)),
+                                             tl.transpose(tl.unfold(X, 2))))
+        norm_C = tl.norm(C)
```
aarmey, line 210: Same here.

tensorly/decomposition/cmtf_als.py
```diff
@@ -0,0 +1,229 @@
...
+        C /= norm_C
+        lambda_ *= norm_C
+        V = tl.transpose(solve_least_squares(tl.dot(A, np.diag(gamma)), Y))
+        norm_V = tl.norm(V)
```
aarmey, line 214: Same here.

tensorly/decomposition/cmtf_als.py
IsabellLehmann: thank you!

IsabellLehmann:
i messed up with the merges, i will do another PR

JeanKossaifi:
Thanks @IsabellLehmann, looking forward to the new PR!

PR #160, opened 2020-03-14
Non-negative CP fixes

aarmey:
Two brief fixes:

1. `unnorml_rec_error` is not defined in the non-negative CP function, and so verbose output caused the function to fail. I just removed it here.
2. QR orthogonalization caused negative values and bad behavior since multiplicative update relies on all values being positive. I've had it take the absolute value, since that's what used at initialization.


JeanKossaifi:
Thanks @aarmey! 

I had used list comprehension as it tends to be marginally faster but it's becoming quite long so probably more readable to have the explicit loop. Merging.

PR #161, opened 2020-03-17
[MGR] Adding ignore mode in parafac (solving #72)

cohenjer:
Hi,

I have been working on my own fork of tensorly for some time now, and I though I might start adding some contributions the easy way, with some small tweaks.

Here is a proposed solution to #72 (Fixed modes in Parafac). Since the optimization algorithm is ALS, it is quite straighforward to implement this. I simply changed the loop condition to ignore modes given by the user.

Sorry for the messy PR, there are many line-removal / blank removal... Also there are changes to README and to .gitignore that should absolutely not be pulled, but I don't know how to make a partial PR :/ hope we can work this out easily.

In principal however this PR should be functionning, since I have been using it for some time now without issues.

Issue #72, opened 2018-08-31
Fix modes in PARAFAC


Legor:
Hey everyone. Just wanted to add a feature idea here.

When using PARAFAC to process EEM data, it is very common to fix two modes (namely the modes for emission and excitation) to those values previously generated on a calibration dataset.
Thereby it is possible to estimate only the remaining mode (the concentration mode) and apply scaling factors that have been computed on the calibration data. 

I added this possibility on a personal fork of tensorly and got the expected outcome on a test dataset. However I'm not sure if my implementation will uphold your coding standards here .. since it basically just overwrites the values of a mode with given values after each iteration..

Also many other constraints (besides the non-negativity) are usually discussed together with PARAFAC, depending on the field of application. E.g. these include orthogonality and unimodality constraints on the loading matrices or the possibility to fix single components in a mode.

I think it is worth working on those additions especially since for PARAFAC there isn't any good free alternative to the well known MATLAB-Toolboxes.

Kind Greetings and thanks for your good work so far,
Gordon

JeanKossaifi:
Hi @Legor, thanks for the suggestions and feedback!

Do you mean something similar to the [`partial_tucker`](http://tensorly.org/dev/modules/generated/tensorly.decomposition.partial_tucker.html#tensorly.decomposition.partial_tucker), but for CP -- i.e. decompose only along some of the modes? Or do you want to provide the factors of the decomposition for some of the modes?

It would be a nice addition - would be great if you wanted to help add it! Happy to have a look at your code/fork, and discuss the API!

Regarding other constraints, we do have orthogonalisation in CP thanks to @cswiercz, as well as non-negative CP. Which other constraints do you need for your work?

Legor:
Greetings @JeanKossaifi, 
the second use case is what I'm currently interested in: providing the factors for some of the modes, or in some cases provide the factors for certain components of a mode. I can make a PR for this if you're interested. 

Regarding the many possibilites to extend PARAFAC it is worth looking at the N-way Toolbox (Matlab) by
Rasmus Bro & Claus A. Andersson, which you probably allready know. The code for this is freely available http://models.life.ku.dk/nwaytoolbox/download

JeanKossaifi:
How would you propose adding this functionality -- it seems the soft approach would be to do this during initialisation, the other would be a hard assignment during optimisation?

[...]

PR #162, opened 2020-03-17
[MGR] Solving #72 ignore mode feature

cohenjer:
This is a remake of PR #161, but clean :) For simplicity purpose I copy-paste #161 here.
---------------------------
Hi,

I have been working on my own fork of tensorly for some time now, and I though I might start adding some contributions the easy way, with some small tweaks.

Here is a proposed solution to #72 (Fixed modes in Parafac). Since the optimization algorithm is ALS, it is quite straightforward to implement this. I simply changed the loop condition to ignore modes given by the user. I tested this on a more advanced branch of my fork, but this should do the trick.

Also, it is straightforward to to the same change to other functions such as nonnegative parafac or Tucker. I just wanted to try this simple PR first since I have little experience with it.

### DIFF
--- a/tensorly/decomposition/candecomp_parafac.py
+++ b/tensorly/decomposition/candecomp_parafac.py
@@ -88,7 +88,8 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd',\
             tol=1e-8, random_state=None,\
             verbose=0, return_errors=False,\
             l2_reg=0, mask=None,\
-            cvg_criterion = 'abs_rec_error'):
+            cvg_criterion = 'abs_rec_error',\
+            fixed_modes = []):
     """CANDECOMP/PARAFAC decomposition via alternating least squares (ALS)
     Computes a rank-`rank` decomposition of `tensor` [1]_ such that,
 
@@ -125,6 +126,8 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd',\
        Stopping criterion for ALS, works if `tol` is not None. 
        If 'rec_error',  ALS stops at current iteration if (previous rec_error - current rec_error) < tol.
        If 'abs_rec_error', ALS terminates when |previous rec_error - current rec_error| < tol.
+    fixed_modes : list, default is []
+        A list of modes for which the initial value is not modified.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
Thanks @cohenjer!

Is `fixed_modes` meant to be a list of modes? If so, is the user meant to also pass in a list of factors? Otherwise we are just initializing the fixed modes randomly (or with SVD depending on `init`).

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -160,13 +163,17 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd',\
     weights = tl.ones(rank, **tl.context(tensor))
     Id = tl.eye(rank)*l2_reg
 
+    # MGR: fixed Modes
+    # Generating the mode update sequence
+    gen = (mode for mode in range(tl.ndim(tensor)) if mode not in fixed_modes)
```
JeanKossaifi: Why a generator rather than a list?

cohenjer:
- Yes fixed_modes would be a list of modes. And it makes indeed little sense to fix a mode if the initial factor values cannot be manually given. Arguably, this is a desirable feature anyway? I can add something like this in the PR, by modifying the initialize function and adding an optinal input.
- About generator vs list, I actually used lists in the other PR (which was more advanced). It does not really matter while we are not interested in picking an element in the generator. What is your opinion?

tensorly/decomposition/candecomp_parafac.py
cohenjer: I actually used lists in the other PR (which was more advanced). It does not really matter as long as we are not interested in picking an element in the generator. I can change that for sure.

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -15,8 +15,9 @@
 
 # License: BSD 3 clause
 
-def initialize_factors(tensor, rank, init='svd', svd='numpy_svd', random_state=None, 
-                       non_negative=False, normalize_factors=False):
+def initialize_factors(tensor, rank, init='svd', svd='numpy_svd', random_state=None,
```
cohenjer, line 18: I chose to keep the parafac function as similar as possible with the original one, by changin instead the initialization function. Also this way, we can propagate the changes easily to other algorithms in the same file, and also to other problems such as Tucker which use a similar initialization function.

JeanKossaifi:
Thanks @cohenjer, sorry for the delay.

Since we are adding this, I think it might be worth giving the option to give only a few input factors with their corresponding mode, e.g. 
`fixed_factors` being a list of tuples `(mode: int, factor : tensor)`.

cohenjer:
So, I think fixing factors and providing initial factors are two different things. 
I agree it makes no sense (to me at least) to fix a mode and not provide an initial factor on that fixed mode. However, we might want to allow the user to use his own initialization, without fixing that mode.
Then having initial factors and fixed mode in one variable is a little weird. So I suggest that we keep fixed_modes and input_factors in two different variables.

We may provide a warning however when someone is fixing a mode but not providing an initial value.

What do you think ?

Issue #167, opened 2020-04-29
Candecomp-Parafac decomposition for tf.sparse.SparseTensor


amanj120:
Currently, there is no support for CP-Decomposition on tensorflow's [Sparse Tensor](https://www.tensorflow.org/api_docs/python/tf/sparse/SparseTensor) library.

This pull request is an addition to the tensorly.contrib.sparse.decomposition module

```python
from tensorly.contrib.sparse.decomposition.tensorflow_backend import parafac
# Let st be a tensorflow.sparse.SparseTensor object
kt = parafac(st, 10)
# should return a kruskal tensor of rank = 10
```

PR #168, opened 2020-04-29
Implement CP Decomposition for tensorflow.sparse.SparseTensor

amanj120:
Aims to solve issue #167

### DIFF
--- a/tensorly/contrib/sparse/backend/tensorflow_backend.py
+++ b/tensorly/contrib/sparse/backend/tensorflow_backend.py
@@ -1,0 +1,98 @@
+import numpy as np
+import tensorflow as tf
+from . import Backend, KruskalTensor
+
+def is_sparse(x):
+    return isinstance(x, tf.sparse.SparseTensor)
+
+class TensorflowBackend(Backend):
+    backend_name = 'tensorflow.sparse'
+
+    @staticmethod
+    def tensor(data, dtype=np.float32, device=None, device_id=None):
+        if isinstance(data, tf.sparse.SparseTensor):
+            return data
+
+# This function is easy to decorate with Numba '@njit(parallel=True)' directive for ~10x improvement in performance
+def mttkrp(self, values, indices, factors, n, rank, dims):
+    output = np.zeros((dims[n], rank))
+    # if decorating with Numba '@njit(parallel=True)' directive, replcase range with numba.prange
+    for l in range(len(values)):
+        cur_index = indices[l]
+        prod = [values[l]] * rank  # makes the value into a row
+
+        for mode, cv in enumerate(cur_index):  # does elementwise row multiplications
+            if (mode != n):
+                for r in range(rank):
+                    prod[r] *= factors[mode][cv][r]
+
+        for r in range(rank):
+            output[cur_index[n]][r] += prod[r]
+
+    return output
+
+def parafac(tensor, rank, n_iter_max=100, verbose=False):
+    """CANDECOMP/PARAFAC decomposition via alternating least squares (ALS)
+        Computes a rank-`rank` decomposition of `tensor` [1]_ such that,
+
+            ``tensor = [|weights; factors[0], ..., factors[-1] |]``.
+
+        Parameters
+        ----------
+        tensor : tf.sparse.SparseTensor
+        rank  : int
+            Number of components.
+        n_iter_max : int
+            Maximum number of iteration
+        verbose: bool
+            Wether or not to print progress at every iteration
+
+        Returns
+        -------
+        KruskalTensor : (weight, factors)
+            * weights : 1D array of shape (rank, )
+                all ones if normalize_factors is False (default),
+                weights of the (normalized) factors otherwise
+            * factors : List of factors of the CP decomposition element `i` is of shape
+                (tensor.shape[i], rank)
+
+        References
+        ----------
+        .. [1] T.G.Kolda and B.W.Bader, "Tensor Decompositions and Applications",
+           SIAM REVIEW, vol. 51, n. 3, pp. 455-500, 2009.
+        """
+
+    dims = tensor.shape.as_list()
+    nd = len(dims)
+    factors = [np.random.random((d, rank)) for d in dims]
+    weights = np.ones((1, rank))
+
+    for iteration in range(n_iter_max):
+        if verbose:
+            print("finished {} iterations".format(iteration+1), end="\r")
+        for n in range(nd):
+
+            # the following block calculates inverse of the hadamard product
+            h = np.matmul(weights.T, weights)
+            for i, f in enumerate(factors):
+                if i != n:
+                    h *= np.matmul(f.T, f)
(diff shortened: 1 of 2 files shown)


amanj120:
I am not sure how to refactor the code to work with the tensorly programming guidelines/model. Any help would be appreciated. 

Running
```python
from tensorly.contrib.sparse.tests.test_decomposition import test_tf_sparse_cpd
test_tf_sparse_cpd()
```
produces the correct output

JeanKossaifi:
Hi @amanj120,

Thanks for the PR, I will try to go through it as soon as I have some time. For reference, you want to define a new backend for the TensorFlow Sparse tensors, as is done for the PyData [Sparse backend](https://github.com/tensorly/tensorly/blob/master/tensorly/contrib/sparse/backend/numpy_backend.py). 

The functions from the backend are dynamically dispatch (in other words, every time you call a function from the backend, an intermediate function decorator checks the current backend and calls the function from that specific backend). What happens when you use the [sparse context](https://github.com/tensorly/tensorly/blob/9d785e204fffc4db918b256a2b8ba2a5a169c554/tensorly/contrib/sparse/backend/__init__.py#L14) is that the functions you define in that backend will be called instead of the ones from the dense backend.

tensorly/contrib/sparse/kruskal_tensor.py
```diff
@@ -1,5 +1,29 @@
...
+from numpy import zeros
+import tensorly as tl
+
+
 kruskal_to_tensor = wrap(kruskal_to_tensor)
 unfolding_dot_khatri_rao = wrap(unfolding_dot_khatri_rao)
+
+
+def sparse_mttkrp(tensor, factors, n, rank, dims):
+    values = tl.values(tensor)
```
amanj120, line 13: I think all sparse backends should have a "values" and "indices" method

JeanKossaifi:
Hi, sorry for the delay -- so in contrib.sparse, all the backends *are* sparse, so you only need tensorflow, not tensorflow.sparse. In other words, to use sparse tensors with tensorflow you just have to set the backend to tensorflow and use tensorly.contrib.sparse

Ideally all should then work transparently.

20zinnm:
What is the status of this PR? I'm hoping to use TensorLy for some large sparse PARAFAC decompositions and right now it's limited to CPU-only. Would TensorFlow support allow GPU acceleration?

amanj120:
Hello,

I've been really busy since school ended, so I haven't had time to look at this PR in a while. @JeanKossaifi I will try to look into the changes you mentioned this coming weekend. @20zinnm The changes I've made so far are CPU-Only. 

A GPU implementation for sparse PARAFAC might use [Numba CUDA](https://numba.pydata.org/numba-doc/latest/cuda/index.html) or some similar alternative, but that falls outside the scope of this project.

tensorly/backend/core.py
```diff
@@ -574,6 +574,17 @@ def conj(x, *args, **kwargs):
...
+    def values(tensor):
+        """ Returns all the non zero values of the tensor in COO
+        """
+        raise NotImplementedError
+
+    @staticmethod
+    def indices(tensor):
+        """ Returns all the indices of non zero values of the tensor in COO
+        """
+        raise NotImplementedError
```
amanj120, line 588: I think in order to have static `indices` and `values` methods for all sparse backends, I had to provide an implementation for the non-sparse backends too, which is why I am changing `tensorly/backend/core.py`

tensorly/backend/tensorflow_backend.py
```diff
@@ -43,11 +43,11 @@ def to_numpy(tensor):
 
     @staticmethod
     def ndim(tensor):
-        return len(tensor.get_shape()._dims)
+        return len(tl.to_numpy(tensor).shape)
```
amanj120, line 47: For some reason, the old implementation was failing for me. This might be a problem with the dynamic dispatching though. If it turns out my implementation of something else was wrong, I'll go back and fix this.

tensorly/contrib/sparse/kruskal_tensor.py
```diff
@@ -1,5 +1,54 @@
...
+from numpy import zeros, multiply
+import tensorly as tl
+
 
 kruskal_to_tensor = wrap(kruskal_to_tensor)
 unfolding_dot_khatri_rao = wrap(unfolding_dot_khatri_rao)
+
+
+def sparse_mttkrp(tensor, factors, n, rank, dims):
+    with sparse_context():
```
amanj120, line 14: I don't know if I need to have a `with sparse_context():` for every method in this file. The reason I include this line is because I call the `values` and `indices` methods inside this method, and the implementations for those only exist if the backend is sparse.

tensorly/contrib/sparse/tests/test_decomposition.py
```diff
@@ -23,3 +29,51 @@ def test_sparse_parafac():
...
+
+    indices = list(set(idxs4))
+    values = np.random.rand(len(indices))
+    indices.sort()
+    st = tf.sparse.SparseTensor(indices=indices, values=values, dense_shape=dimensions)
+    return st
+
+
+def test_tf_sparse_cpd():
+    with sparse_context():
```
amanj120, line 50: Same thing about `with sparse_context:` here.

tensorly/contrib/sparse/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,82 @@
...
+from ....backend.core import Backend
+from copy import copy as _py_copy
+
+
+def is_sparse(x):
+    return isinstance(x, tf.sparse.SparseTensor)
+
+
+class TensorflowSparseBackend(Backend):
+    backend_name = 'tensorflow'
```
amanj120, line 12: Should the backend_name be `tensorflow` or `tensorflow.sparse`? In [`numpy_backend`](https://github.com/tensorly/tensorly/blob/b896b142b09f559f981237d785738467a05dcaf3/tensorly/contrib/sparse/backend/numpy_backend.py#L25) the name is `numpy.sparse`, but if I set the name to `tensorflow.sparse` here, I get a `ValueError: Unknown backend name 'tensorflow.sparse', known backends are ['numpy', 'tensorflow']` error.

tensorly/contrib/sparse/backend/tensorflow_backend.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @staticmethod
+    def tensor(data, dtype=np.float32, device=None, device_id=None):
+        if isinstance(data, tf.sparse.SparseTensor):
+            return data
+        elif isinstance(data, tuple):
```
amanj120, line 18: This `elif` branch is necessary in order to create tensors without having to import tensorflow. Importing tensorflow is what was causing the Travis builds to fail.

JeanKossaifi:
Hey @amanj120! Did you manage to make it work?

JeanKossaifi:
Hi @amanj120 -- are you still working on this?

PR #170, opened 2020-05-19
Fix mask shape during kr calculation

### LINKED ISSUE #142: Masked CP decomposition returns error in numpy backend kr
Hi, like milanlanlan in [Handling missing data in decomposition #4](https://github.com/tensorly/tensorly/issues/4#issuecomment-557899752) I am trying to decompose a tensor using the CP decomposition with missing values. Unfortunately, I also receive this error message: 

>    5 mask_parafac[score_sparse > 0] = 1
      6 print(mask_parafac)
      7 rec_score_tensor = parafac(score_sparse, rank=4, mask = mask)
      8 
      9 print(rec_score_tensor)
tensorly/decomposition/candecomp_parafac.py in parafac(tensor, rank, n_iter_max, init, svd, normalize_factors, tol, orthogonalise, random_state, verbose, return_errors, non_negative, mask)
    183 
    184             if mask is not None:
    185                 tensor = tensor*mask + tl.kruskal_to_tensor((None, factors), mask=1-mask)
    186 
    187             mttkrp = unfolding_dot_khatri_rao(tensor, (None, factors), mode)
tensorly/kruskal_tensor.py in kruskal_to_tensor(kruskal_tensor, mask)
    186                              T.transpose(khatri_rao(factors, skip_matrix=0)))
    187     else:
    188         full_tensor = T.sum(khatri_rao([factors[0]*weights]+factors[1:], mask=mask), axis=1)
    189 
    190     return fold(full_tensor, 0, shape)
tensorly/tenalg/_khatri_rao.py in khatri_rao(matrices, weights, skip_matrix, reverse, mask)
     96         # Note: we do NOT use .reverse() which would reverse matrices even outside this function
     97 
     98     return T.kr(matrices, weights=weights, mask=mask)
tensorly/backend/__init__.py in inner(*args, **kwargs)
    158 
    159     def inner(*args, **kwargs):
    160         return _get_backend_method(name)(*args, **kwargs)
    161 
    162     # We don't use `functools.wraps` here because some of the dispatched
tensorly/backend/numpy_backend.py in kr(self, matrices, weights, mask)
     67             matrices = [m if i else m*self.reshape(weights, (1, -1)) for i, m in enumerate(matrices)]
     68 
     69         return np.einsum(operation, *matrices).reshape((-1, n_columns))*mask
     70         #tensor = np.einsum(operation, *matrices).reshape((-1, n_columns))
     71         #return tensor* mask.reshape(tensor.shape)
ValueError: operands could not be broadcast together with shapes (80,4) (5,4,4) <

As the number of elements in both arrays is not equal it seems like a simple reshape will not solve the problem.

aarmey:
I believe this fixes #142. A very basic test is also added here to check that a mask with no effect gives the same result.

The problem ultimately came down to the kr form being unfolded. The mask effect could be handled within parafac more simply, but I wasn't sure whether there is a reason for the current design.

### DIFF
--- a/tensorly/backend/numpy_backend.py
+++ b/tensorly/backend/numpy_backend.py
@@ -53,11 +53,6 @@ class NumpyBackend(Backend):
             return np.sum(np.abs(tensor)**order, axis=axis)**(1 / order)
 
     def kr(self, matrices, weights=None, mask=None):
-        if mask is None:
-            mask = 1
-        else:
-            mask = mask.reshape((-1, 1))
-
         n_columns = matrices[0].shape[1]
         n_factors = len(matrices)
 
@@ -70,7 +65,8 @@ class NumpyBackend(Backend):
         if weights is not None:
             matrices = [m if i else m*self.reshape(weights, (1, -1)) for i, m in enumerate(matrices)]
 
-        return np.einsum(operation, *matrices).reshape((-1, n_columns))*mask
+        m = mask.reshape((-1, 1)) if mask is not None else 1
+        return np.einsum(operation, *matrices).reshape((-1, n_columns))*m
 
     @property
     def SVD_FUNS(self):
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
Thanks a lot Aaron, this is great, merging!

PR #171, opened 2020-05-29
Fixing modes in parafac and nonegative_parafac

cohenjer:
This is a new PR, resurrecting #162.

It allows the user to fix several modes in the parafac and nonnegative parafac decomposition. The syntax is tl.decomp.parafac(T, rank, ..., fixed_modes=[a, b, c]) for fixing modes a,b,c (which are integers here). 

It is not allowed to fix the last mode, because this will break the error computation that uses quantities such as mttkrp computed for the last mode update. In that case a warning is issued and the mode is updated.

Finally, I have added a temporary notebook in the "test" folder (did not know where else to put it) to show the code works as intended. I can remove it from the PR once everyone is convinced.

Note that the fixed-mode feature makes much more sense if a user can chose his own initialization. This was discussed in #162 and is actually part of #150.


JeanKossaifi:
Seems good to me and #150  is now merged.
We can probably add a unit-test case (based on the notebooks you made) and add to `tensorly/decomposition/tests/test_candecomp_parafac.py`.

Regarding the last mode, perhaps we can try to make it possible in the future.

cohenjer:
@JeanKossaifi Sure I will work in this today. There is no existing unit test for Parafac, I will copy the Parafac2 one if that's OK.

cohenjer:
Also the dev branch of tensorly is not up to date with master, I think the Parafac2 merge was done directly on master. My branch is on dev. I will solve the merging issues, but @JeanKossaifi  can you merge master into dev ?

JeanKossaifi:
@cohenjer dev is now even with master.
We do have tests for parafac, in the file I mentioned (and generally all functions should have a unit-test): 
https://github.com/tensorly/tensorly/blob/f9be446ec9f7fd5827dcbab8efad58ead99d124e/tensorly/decomposition/tests/test_candecomp_parafac.py#L15

tensorly/decomposition/tests/test_candecomp_parafac.py
```diff
@@ -31,6 +31,17 @@ def test_parafac():
...
 
+    # Test fixing mode 0 or 1 with given init
+    fixed_tensor = random_kruskal((3, 4, 2), rank=2)
+    rec_svd_fixed_mode_0 = parafac(tensor, rank=2, n_iter_max=2, init=fixed_tensor, fixed_modes=[0])
+    rec_svd_fixed_mode_1 = parafac(tensor, rank=2, n_iter_max=2, init=fixed_tensor, fixed_modes=[1])
+    # Check if modified after 2 iterations
+    tol_fixed = 1e-16
+    assert_(T.norm(rec_svd_fixed_mode_0.factors[0] - fixed_tensor.factors[0]) < tol_fixed,
+            'Fixed mode 0 was modified in candecomp_parafac')
+    assert_(T.norm(rec_svd_fixed_mode_1.factors[1] - fixed_tensor.factors[1]) < tol_fixed,
```
JeanKossaifi, line 42:
```suggestion
   assert_array_equal(rec_svd_fixed_mode_1.factors[1], fixed_tensor.factors[1]) 
```
We can directly use `tensorly.testing.assert_array_equal` for checking that the factors were left untouched.

JeanKossaifi:
Thanks @cohenjer, looks great! Just one detail: for the tests, we can use directly `assert_array_equal` from `tensorly.testing` to check that they actually weren't changed.

tensorly/decomposition/tests/test_candecomp_parafac.py
cohenjer: Ok let me correct that quickly

JeanKossaifi:
Thanks, merging!

PR #173, opened 2020-06-11
Implement masked Tucker

aarmey:
I believe this implements masking for Tucker decomposition, which then fixes the last of issue #4. The test here is pretty minimal; if you have an idea for something better I can implement it.


JeanKossaifi:
Thanks @aarmey, this is great!

For the test, we could add with a test case with a non uniform mask. For instance, perhaps we could build a low-rank Tucker tensor, mask a few values (e.g. from redundant components only) and check that they are correctly recovered?

aarmey:
Good suggestion—it should be implemented here. The test helped to reveal that masking really doesn't work with SVD initialization because the starting point reflects the masked value. I've added a warning about this in both Tucker and PARAFAC.

JeanKossaifi:
Thanks @aarmey, this is great!
Merging.

PR #175, opened 2020-06-15
Bugfix: PARAFAC2 not using "init" option

aarmey:
The `init` option wasn't being used, because it wasn't passed to `initialize_decomposition`. Fixed here.


JeanKossaifi:
Thanks @aarmey, good catch, not sure how we missed that!

Issue #179, opened 2020-06-22
Tolerances given in 10e-xx and 1e-xx


cohenjer:
Just a quick bug report concerning various incoherent uses of 10e-xx and 1e-xx. For instance, 
tl.decomposition.test.test_candecomp_parafac.py contains lines such as
` tol_norm_2 = 10e-2 `
but inside the decomposition, most tolerances are written as 1e-xx, such as
` def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd',\
            normalize_factors=False, orthogonalise=False,\
            tol=1e-8, random_state=None,\... `
in tl.decomposition.candecomp_parafac.py

A proposed fixed is to use solely 1e-xx, since 10e-xx is quite counter-intuitive. For instance, 10e-2 is 0.1, while 1e-2 is 0.01. Of course to keep the meaning unchanged, 10e-x should be changed to 1e-(x-1)

I can make the changes 10e-x --> 1e-(x-1) in all the files containing 10e-x if we agree on that solution.

JeanKossaifi:
I agree, consistency is important and it doesn't make sense to use 10eX, it must be a relic from 10^X. 

> I can make the changes 10e-x --> 1e-(x-1) in all the files containing 10e-x if we agree on that solution.

Would be great, thanks!

JeanKossaifi:
Fixed by #180

PR #180, opened 2020-06-22
Changed all 10e-x to 1e-(x-1)

cohenjer:
A quickfix to #179 
Basically replacing all 10e-x by 1e-(x-1)  (e.g. 10e-2 by 1e-1) in all files of Tensorly, as suggested in #179.


JeanKossaifi:
Thanks @cohenjer! Merging.

PR #181, opened 2020-06-22
Add test where rank > min_dim.

aarmey:
I believe this is an error in PARAFAC2—I've added a test here to demonstrate what I'm seeing. Specifically when rank > min_dim, SVD initialization fails. It's not a problem with SVD but rather using the results to construct the PARAFAC2 decomposition. I'm not confident enough what is happening here to fix it. @MarieRoald, perhaps you could advise?

```
    def _compute_projections(tensor_slices, factors, svd_fun, out=None):
        A, B, C = factors
    
        if out is None:
            out = [T.zeros((tensor_slice.shape[0], C.shape[1]), **T.context(tensor_slice)) for tensor_slice in tensor_slices]
    
        slice_idxes = range(T.shape(A)[0])
        for projection, i, tensor_slice in zip(out, slice_idxes, tensor_slices):
            a_i = A[i]
>           lhs = T.dot(B, T.transpose(a_i*C))
E           ValueError: operands could not be broadcast together with shapes (3,) (30,4)
```

### DIFF
--- a/tensorly/decomposition/tests/test_parafac2.py
+++ b/tensorly/decomposition/tests/test_parafac2.py
@@ -124,11 +124,12 @@ def test_parafac2_normalize_factors():
     assert abs(tl.max(norms) - tl.max(normalized_rec.weights))/tl.max(norms) < 1e-2
     assert abs(tl.min(norms) - tl.min(normalized_rec.weights))/tl.min(norms) < 1e-2
 
-def test_parafac2_init_valid():
-    rng = check_random_state(1234)
-    rank = 3
 
-    random_parafac2_tensor = random_parafac2(shapes=[(15, 30)]*25, rank=rank, random_state=rng)
+@pytest.mark.parametrize("rank", [2, 4])
+def test_parafac2_init_valid(rank):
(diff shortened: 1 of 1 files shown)


MarieRoald:
Hi @aarmey! 

It looks like the problem is that since SVD init initialises the components for the third mode with a PCA of the tensor slice. Unfortunately, for an I-times-J-times-K tensor, we cannot get more than K columns in the factor matrix this way. If the rank is larger than K, we can not use SVD init.  I think we have to raise a ValueError here. To be sure, we tested SVD init on a similar problem with Rasmus Bro's PARAFAC2 code (available here: http://models.life.ku.dk/sites/default/files/parafac2.m ) published with the Kiers et al. (1999) paper and got the same kind of error.

Also, there we found two bugs in the SVD initialisation: We used SVD to initialise the A components, but it should be initialised with ones instead, and it used tuples instead of lists when creating the PARAFAC2 tensor object. We are submitting a pull request about that (and the value error) now.

aarmey:
Thanks @MarieRoald! I'll close this then.

Issue #74, opened 2018-09-10
Sign indeterminacy of SVD


JeanKossaifi:
The Singular Value Decomposition suffers of [sign indeterminacy](https://prod.sandia.gov/techlib-noauth/access-control.cgi/2007/076422.pdf). We can add a sign flip function and an additional flag on whether to use it to make output of SVD deterministic.

PR #182, opened 2020-06-24
Bugfix for PARAFAC2 SVD init

MarieRoald:
Hi,

Based on the test by @aarmey, we found some issues with SVD init of PARAFAC2

1. SVD init of A was incorrect, should be initialised with ones (Kiers et al. 1999)
    * We generated the decomposition with a tuple of factor matrices. Using tuples instead of lists made it fail later as there is no casting to a list in the PARAFAC2 class (or other decomposition classes for that sake).
1. Maybe this deserves an issue?
1. There was no sensible error when SVD init was used with an incompatible tensor (each slice must have at least `rank` columns)

After looking at the code by Rasmus Bro, we noticed that his implementation is slightly different from what is described in the paper. Bro's implementation adds normally distributed random noise (sigma = 0.1) on top of the all-ones components. This is maybe worth looking into?


MarieRoald:
It seems like we have some issues with the TensorFlow backend again. Though, this time it may be because the SVD init converges to a local minimum ([Kiers et al 1999](https://onlinelibrary.wiley.com/doi/abs/10.1002/(SICI)1099-128X(199905/08)13:3/4%3C275::AID-CEM543%3E3.0.CO;2-B)). I have experienced this myself a couple of times, where the SVD init converges to a suboptimal solution, which is why I usually use many random initialisations instead.

I'm not sure why TensorFlow obtains a suboptimal minimum while the others succeed. It may have something to do with the sign indeterminacy of the SVD causes a slightly different initialisation ([Bro et al. 2008](https://onlinelibrary.wiley.com/doi/abs/10.1002/cem.1122))?

JeanKossaifi:
Thanks @MarieRoald !
Not sure why TensorFlow is complaining -- it should be using numpy's SVD by default, I'll try to look into it in the coming days.

JeanKossaifi:
@MarieRoald I realized this is still open! What shall we do?

Are we sure the issue comes from the SVD? I guess we could always address the sign indeterminacy issue anyway, it's been long standing as it hadn't really been a big issue so far ( #74 ).

I checked and the code seems to indeed be using `numpy_svd` (i.e. converting to numpy to apply svd and back) so there shouldn't be a difference there. The difference in result seems high to be just due to numerical issues too. I guess we could skip the test for tensorflow or decrease the threshold of the test but these are less than ideal solutions.

MarieRoald:
I found a bug with the `_pad_by_zeros` function, which was only used by the SVD initialisation. We had the line `tl.index_update(padded, tl.index[i, :J_i], tensor_slice)` instead of `padded = tl.index_update(padded, tl.index[i, :J_i], tensor_slice)`. This worked well for the other backends, since for them, the `index_update` method is inplace. However, since it isn't inplace for TensorFlow,  the `_pad_by_zeros` function returned a tensor consisting of all zeros that was used for initialisation of the C-matrix.

After fixing this bug, all tests run well with TensorFlow on my local machine, but I cannot get them running on Travis for some reason. Does it have something to do with [Travis no longer offering unlimited credits for OpenSource projects](https://news.ycombinator.com/item?id=25338983)?

JeanKossaifi:
Awesome, thanks @MarieRoald -- yes indeed, Travis went through all my personal credits (without any warning!) and they haven't responded to my email -- I will switch to Github actions.

tensorly/decomposition/parafac2.py
```diff
@@ -40,13 +40,18 @@ def initialize_decomposition(tensor_slices, rank, init='random', svd='numpy_svd'
...
                     svd, tl.get_backend(), tl.SVD_FUNS)
             raise ValueError(message)
-        
         padded_tensor = _pad_by_zeros(tensor_slices)
-        A = svd_fun(unfold(padded_tensor, 0), n_eigenvecs=rank)[0]
+        A = T.ones((padded_tensor.shape[0], rank), **context)
+
+        unfolded_mode_2 = unfold(padded_tensor, 2)
+        if T.shape(unfolded_mode_2)[0] < rank:
+            raise ValueError("Cannot perform SVD init if rank ({}) is less than the number of columns in each tensor slice ({})".format(
```
aarmey, line 48:
```suggestion
            raise ValueError("Cannot perform SVD init if rank ({}) is greater than the number of columns in each tensor slice ({})".format(
```

Tiny issue—it appears the message here is backwards from the if statement immediately preceding it.

JeanKossaifi: Thanks for spotting @aarmey ! I accepted the change @MarieRoald to trigger the tests.

JeanKossaifi:
Thanks a lot @MarieRoald! Merging! :)

Issue #158, opened 2020-03-12
Line Search


JeanKossaifi:
It would be good to add enhanced Line Search [1], at least to CP, as proposed in #29.

[1] M. Rajih, et. al., "Enhanced Line Search: A Novel Method to Accelerate PARAFAC", SIAM J. Matrix Anal. Appl., Vol. 30, No. 3, pp. 1128-1147

aarmey:
I've been playing around with implementing a basic line search (copying it from Bro's implementation). To do this, it would be helpful to refactor out the [reconstruction error calculations from CP](https://github.com/tensorly/tensorly/blob/master/tensorly/decomposition/candecomp_parafac.py#L251-L263). I think the only downside to this is [`mttkrp`](https://github.com/tensorly/tensorly/blob/master/tensorly/decomposition/candecomp_parafac.py#L237) will be recalculated rather than reused. I figure this added cost is acceptable? If so I'll file a PR.

JeanKossaifi:
Added in #183

PR #183, opened 2020-07-01
PARAFAC line search

aarmey:
This implements a line search during PARAFAC decomposition. It is definitely unfinished and a very rudimentary implementation, but I'm interested in whether there are potential issues with the refactoring required here. It seems to vastly improve convergence, particularly for tensors with masking and/or PARAFAC2.

Related to #158


JeanKossaifi:
That's great @aarmey, it's been on the todo list for a while! I will have a more detailed look in the coming days. 
I guess we can switch between reusing the parts to compute the errors (no line search) or recomputing it from scratch (when using line-search). I'm curious to see the tradeoff compute-time/convergence.

aarmey:
This very rudimentary method is faster with missing values (probably because it repeatedly re-masks the tensor during line search). However, while it takes few iterations to converge, it takes about the same amount of time with complete tensors. Going to see if I can implement the enhanced line search from Rajih et al.

aarmey:
I found a slow spot by profiling, and the enhanced line search paper had a good suggestion (to only try every other iteration). Now this seems to consistently converge faster by line search, even without high collinearity or masking (where it'd do even better). Plot from the included notebook (wps.ipynb):

Unknown

The method here is actually, if anything, probably closer to Rajih et al than Bro's method, because the search distance is adjusted within-iteration. It's not using the polynomial calculation though—just brute force.

cohenjer:
@aarmey Hey, I'm curious to how you chose the "jump" values? Did you use the numbers given in a specific publication? Or in a particular implementation? I think the line-search from Comon et. al. [link](https://hal.archives-ouvertes.fr/hal-00327595/document) is quite different from the simple version your proposed, while the implementation of Bro et. al. (pp 95 of Bro's thesis describes the jump) is very close to yours but with a different rule for the step.

aarmey:
@cohenjer you're right that this is different from both these published strategies. My reasoning was to have an increasing step if they are repeatedly accepted. My application also has many missing values, and so I wanted to include mask updates during the line search. That led me to do this stepping, rather than the single jump in Bro or Rajih.

It would be easy to bring this in line with Bro—perhaps this is best, so that we can just provide a reference describing the implementation?

As I mentioned above the only similarity to Rajih/Comon et al is that they calculate a line search step size after each round of ALS. This seems better, but I'm not confident I could implement their method. As far as I could tell, they only define the 3-way case, so the method would need to be extended to N-way situations.

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -247,20 +264,46 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd',\
...
+            weights_diff = weights - weights_last
+            factors_diff = [factors[ii] - factors_last[ii] for ii in range(tl.ndim(tensor))]
+
+            while True:
+                # Keep stepping while we're successful.
+                new_weights = weights + jump * weights_diff
+                new_factors = [tl.copy(f) for f in factors]
+
+                for ii in modes_list:
+                    new_factors[ii] += factors_diff[ii]
```
cohenjer, line 283: So if I understand this section right, you don't use the jump variable for the factors update, rather only for the weights. Therefore you basically go several times in the direction given by the ALS update on the factors. This feels a little weird tbh, I would expect this line to be 
```
new_factors[ii] += jump * factors_diff[ii]
```

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -247,20 +264,46 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd',\
...
+                if mask is not None:
+                    tensor_new = tensor*mask + tl.kruskal_to_tensor((new_weights, new_factors), mask=1-mask)
+                else:
+                    tensor_new = tensor
 
-                # mttkrp and factor for the last mode. This is equivalent to the
-                # inner product <tensor, factorization>
-                iprod = tl.sum(tl.sum(mttkrp*factor, axis=0)*weights)
-                unnorml_rec_error = tl.sqrt(tl.abs(norm_tensor**2 + factors_norm**2 - 2*iprod))
+                new_rec_error = error_calc(tensor_new, new_weights, new_factors, sparsity)
```
cohenjer, line 290: With your approach you need to compute the error not once, but as many times as you look for a best jump and keep improving. Since computing the cost function for Parafac is as costly as an ALS update, this seems quite expensive? Am I missing something?

tensorly/decomposition/candecomp_parafac.py
aarmey: Oops. That's a typo from fixing an mxnet backend error.

aarmey:
Agreed. I'll turn this into Bro-ALS. After that, herALS seems more feasible to implement than Rajih.

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -247,21 +315,38 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd',\
...
-                
+        # Start line search if requested. Doesn't yet work for fixed modes.
+        if linesearch and iteration % 2 == 0:
+            jump = iteration ** (1.0/3.0)
+
+            weights_diff = weights - weights_last
+            factors_diff = [factors[ii] - factors_last[ii] for ii in range(tl.ndim(tensor))]
+
+            # Keep stepping while we're successful.
+            new_weights = weights + jump * weights_diff
```
aarmey, line 333: I think Bro jumps from `weights_last` instead of `weights`, because the jump here is essentially (1 + jump) per iteration. However, the error calculation for a jump of 1.0 is much cheaper since we already have `mttkrp`. I'm interested in what you think.

aarmey:
@cohenjer This now performs Bro-ALS.

aarmey:
@cohenjer I went through the N-way Toolbox code and updated this to match. When the extrapolation fails only the typical ALS update is used like here. The comments and code in the toolbox didn't match in one place—the comments indicate `n` is incremented by 1 after max_fail, but the code has it incrementing by 2. I think it's meant to be 1 so that's what I've done here.

JeanKossaifi:
So `n`'s value only increases over time: it is only incremented on failure, but never decremented on success?

aarmey:
True. I think the idea is that the acceleration is always increasing because it includes the iteration count, then the exponent effect reduces it if failing.

Great—I'll make a separate PR with the notebook code.

JeanKossaifi:
Yes, I was just wondering if there are some edge cases where we wouldn't necessarily want `n` to decrease with the iteration but it makes sense in general.

If everyone is happy I can merge. @cohenjer @aarmey

aarmey:
@JeanKossaifi after the changes in the last PR, something odd is happening where the ALS iterations after a line search jump are increasing the error. I'll need to look into this first...

aarmey:
Found it. Should be fixed in the last commit.

JeanKossaifi:
Great, thanks @aarmey!
Is it worth adding a small test checking that the error decreases (or at least, doesn't increase)?

aarmey:
Done.

cohenjer:
> Yes, I was just wondering if there are some edge cases where we wouldn't necessarily want `n` to decrease with the iteration but it makes sense in general.
> 
> If everyone is happy I can merge. @cohenjer @aarmey

@JeanKossaifi  Yeah well that's the Bro strategy. In other works the step can increase when we are doing going. In the context of CP decomposition, all these techniques are heuristics, but at least the Bro one has been used successfully for a long time :)

JeanKossaifi:
Great, thanks @aarmey for the new feature and @cohenjer for the help, merging.

PR #184, opened 2020-07-01
Fix testing errors

aarmey:
This fixes some sporadic and consistent testing errors:

- Rank of MPS cross decomposition reduced
- Fixed shape of mask in masked kr for Jax backend (not sure how this got past before)
- Set random seed on masked Tucker and PARAFAC tests


JeanKossaifi:
Thanks @aarmey, this is really helpful!
It seems the cross-decomposition is still problematic but this would probably require a larger effort to refactor that bit as it's due to the algo occasionally not converging..

PR #185, opened 2020-07-05
Using missing values with SVD initialization in PARAFAC

aarmey:
This provides a new option, `svd_mask_repeats`, which controls restarting SVD to handle masked values.

In the process of putting this together, I noticed that the SVD initialization only initializes using `U`, and so the initialization doesn't match the scale of `tensor` until one iteration through ALS. I made a change here to use `S` during initialization to match the scale.

Currently, it seems like the handling with `normalize_factors=True` was inconsistent before and still is. In `initialize_factors()`, `normalize_factors=True` leads to `factors` being scaled, but `weights` is still initialized to 1.0. Perhaps a more consistent interface would be to have `initialize_factors()` return `weights`, and to use the rescaling it does to set weights. I think this inconsistency quickly goes away as soon as you start the ALS iterations, but currently might be confusing if `n_iter_max` is small.

### DIFF
--- a/tensorly/decomposition/candecomp_parafac.py
+++ b/tensorly/decomposition/candecomp_parafac.py
@@ -35,9 +35,8 @@ def initialize_kruskal(tensor, rank, init='svd', svd='numpy_svd', random_state=N
 
     Returns
     -------
-    factors : ndarray list
-        List of initialized factors of the CP decomposition where element `i`
-        is of shape (tensor.shape[i], rank)
+    factors : KruskalTensor
+        An initial kruskal tensor.
 
     """
     rng = check_random_state(random_state)
--- a/tensorly/decomposition/candecomp_parafac.py
+++ b/tensorly/decomposition/candecomp_parafac.py
@@ -16,7 +16,7 @@ from collections.abc import Mapping
 
 # License: BSD 3 clause
 
-def initialize_factors(tensor, rank, init='svd', svd='numpy_svd', random_state=None, 
+def initialize_kruskal(tensor, rank, init='svd', svd='numpy_svd', random_state=None, 
                        non_negative=False, normalize_factors=False):
     r"""Initialize factors used in `parafac`.
 
@@ -44,14 +44,19 @@ def initialize_factors(tensor, rank, init='svd', svd='numpy_svd', random_state=N
     """
     rng = check_random_state(random_state)
     eps = tl.eps(tensor.dtype)
+    weights = tl.ones(rank, **tl.context(tensor))
 
     if init == 'random':
         factors = [tl.tensor(rng.random_sample((tensor.shape[i], rank)), **tl.context(tensor)) for i in range(tl.ndim(tensor))]
         if non_negative:
             factors = [tl.abs(f) for f in factors]
-        if normalize_factors: 
-            factors = [f/(tl.reshape(tl.norm(f, axis=0), (1, -1)) + eps) for f in factors]
-        return factors
+        if normalize_factors:
+            for ii, f in enumerate(factors):
+                norm = tl.reshape(tl.norm(f, axis=0), (1, -1)) + eps
+                factors[ii] /= norm
+                weights[ii] *= norm
+
+        return KruskalTensor((weights, factors))
 
     elif init == 'svd':
         try:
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
Thanks @aarmey -- good point about the weights. 

A better way to deal with this might be to completely rename `initialize_factors` to  `initialize_kruskal(...)` (or `initialize_CP`? ) which would now return an initialized kruskal tensor and also take care of the weight initialization there.

aarmey:
@JeanKossaifi this should be ready if it looks alright on review.

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -42,14 +42,19 @@ def initialize_factors(tensor, rank, init='svd', svd='numpy_svd', random_state=N
...
+    weights = tl.ones(rank, **tl.context(tensor))
 
     if init == 'random':
         factors = [tl.tensor(rng.random_sample((tensor.shape[i], rank)), **tl.context(tensor)) for i in range(tl.ndim(tensor))]
         if non_negative:
             factors = [tl.abs(f) for f in factors]
-        if normalize_factors: 
-            factors = [f/(tl.reshape(tl.norm(f, axis=0), (1, -1)) + eps) for f in factors]
-        return factors
+        if normalize_factors:
```
JeanKossaifi, line 50: Here we could just used `kruskal_normalize`: 
https://github.com/tensorly/tensorly/blob/cb099d9045c2ed5b40dc03d2d7c4887d7079c05a/tensorly/kruskal_tensor.py#L104

I remember having some issues with the normalization in some edge cases -- is there any strong case for just adding eps vs actually checking for zero as in `kruskal_normalize`?

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -75,14 +85,16 @@ def initialize_factors(tensor, rank, init='svd', svd='numpy_svd', random_state=N
             if non_negative:
                 factor = tl.abs(factor)
             if normalize_factors:
-                factor = factor / (tl.reshape(tl.norm(factor, axis=0), (1, -1)) + eps)
+                norm = tl.reshape(tl.norm(factor, axis=0), (1, -1)) + eps
```
JeanKossaifi, line 87: Same here I guess.

tensorly/decomposition/candecomp_parafac.py
```diff
@@ -307,6 +328,9 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd',\
                     pseudo_inverse = pseudo_inverse*tl.dot(tl.conj(tl.transpose(factor)), factor)
             pseudo_inverse += Id
 
+            if mask is not None:
+                tensor = tensor*mask + tl.kruskal_to_tensor((weights, factors), mask=1-mask)
```
JeanKossaifi, line 331: Good catch -- we weren't using the mask before??

aarmey: It's part of the error calculation step now. I'll remove this.

tensorly/decomposition/candecomp_parafac.py
aarmey: Maybe `eps` can just be removed? It seems like the issue would be if the factor norm is 0.0, but if that's the case I'm not sure what to do besides throw an error... I could add an assert to check for 0.0 so that the error is clearer?

JeanKossaifi:
Thanks a lot @aarmey, you're on a roll! :D

I just added a comment re. normalization: dividing by actual zeros seems safer whereas just adding eps is (slightly) faster but may lead to numerical inaccuracies, particularly with lower precision. @cohenjer do you have some insight on this?

I just made some tests and the difference seems to be negligible (the difference exists mostly for tiny factors but then it's in the order of the microsecond).

tensorly/decomposition/candecomp_parafac.py
JeanKossaifi: I meant that we could simply return tensorly.kruskal_tensor.normalize_kruskal((weights, factors)), this would normalize (by checking for zero values), but, I am not 100% sure whether that is better than just adding eps.
In any case, it seems more consistent to always use that to normalize, so if we change it there, it's updated for all functions.

aarmey:
Didn't know about `kruskal_normalise()`! Here's a more ambitious simplification using that.

JeanKossaifi:
Thanks - merging!

PR #186, opened 2020-07-10
Line search example script

aarmey:
Here is the example script (lightly polished).

Figure_1


JeanKossaifi:
Awesome, thanks @aarmey !

PR #191, opened 2020-08-28
Decomposition classes

JeanKossaifi:
Background
=========

This PR is to have a concrete draft to implement the long discussed new class API. The reference in the matter is probably [Scikit-Learn](https://scikit-learn.org/stable/modules/classes.html) which did an amazing job in designing a clear, simple and effective API. 

There are a few concepts behind it, namely:
	•	Consistency (of an interface with a small limited set of methods)
	•	Inspection
	•	Non-proliferation of classes
	•	Composition
	•	Sensible defaults
You can read more about in the [related paper](https://dtai.cs.kuleuven.be/events/lml2013/papers/lml2013_api_sklearn.pdf). These are all great concepts that I try to enforce in TensorLy.

Going forward, we want to add a class API for tensor decomposition. This is already the case for [tensor regression](http://tensorly.org/stable/modules/generated/tensorly.regression.tucker_regression.TuckerRegressor.html#tensorly.regression.tucker_regression.TuckerRegressor).

Proposed API
==========
Current API, updated to reflect the discussion

```python
class Decomposition:
    def  __init__(self, *args, **kwargs):
        # store all hyper-parameters
        # Only sets the values of the attributes, nothing more

    def fit(tensor, y=None, kruskal_init=None):
        # fit the decomposition on an input tensor
        return self

    def fit_transform(tensor, y=None, kruskal_init=None):
         # ...
         return decomposed_tensor

    def transform(self, tensor):
        # easy for a Tucker decomposition (return the core obtained by projecting with the factors)
        return projected_tensor
```

Principles to consider 
================
Key points discussed in the PR:

* Separation between data and algorithm:  the classes should not store the training data, only their own parameters. In particular, `KruskalTensor` is a separate class from the `Parafac` one. One issue with mixing the two is that it would introduces side effects makes is much harder to compose function (e.g. to create Pipelines).

* Optimization methods: we will have a separate class for each major optimization methods (e.g. `ParafacALS`). We may consider having a base class `Parafac` taking as parameter e.g. `method='als'`, which will specify which optimization method to use. This is slightly tricky as different methods may have different arguments and having a base class might be optimal for discoverability of these.

Questions to answer
===============

Fit_transform or decompose
----------------------------

The question is how to adapt the fit/transform for decompositions such as CP (where fit and transform cannot easily be separated), or Robust Tensor PCA, where we just express `X = L + R` but don't have an explicit way to project input data.

Once a CP decomposition is fit to the data, we get the factors of the decomposition. 
Transforming a new input tensor doesn’t make as much sense (the best I can think of would be to learn new weighting for the learned factors so as to minimise the rec error, this is more of another fit rather than a proper transform). 

The main question: do we decide CP will only have `fit_transform`? Or do we add a method `decompose` for tensor decomposition specifically, that just returns the decomposed tensor? 

Initialization
-------------
A good reference is scikit-learn’s non-negative factorisation (https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.NMF.html). The idea is to have an init param in the __init__ (a string specifying which method to use), and we can have another param in the `fit` and `fit_transform`, either as:

* Single argument: this could all be grouped in a parameter, e.g. `kruskal_tensor_init=None`

* or Individual arguments: e.g. for CP, `weights_init` and `factors_init` for instance.

-------------------------

Keen to get feedback on this! @MarieRoald @yngvem @aarmey @cohenjer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,19 @@ language: python
 matrix:
     include:
      - os: linux
-       python: 3.6
-       env: BACKEND=numpy
-     - os: linux
-       python: 3.7
+       python: 3.8
        env: BACKEND=mxnet
      - os: linux
-       python: 3.7
+       python: 3.8
+       env: BACKEND=jax
+     - os: linux
+       python: 3.8
        env: BACKEND=pytorch
      - os: linux
-       python: 3.7
+       python: 3.8
        env: BACKEND=tensorflow
      - os: linux
-       python: 3.7
+       python: 3.8
        env: BACKEND=numpy DEPLOY_DOCS=1 DEPLOY_PYPI=1
        addons:
          apt:
@@ -44,11 +44,13 @@ install:
     - if [[ "$BACKEND" == "numpy" ]]; then
         pip install sparse;
       elif [[ "$BACKEND" == "pytorch" ]]; then
-        conda install pytorch-cpu torchvision-cpu -c pytorch;
+        conda install pytorch cpuonly -c pytorch;
       elif [[ "$BACKEND" == "tensorflow" ]]; then
         conda install tensorflow;
       elif [[ "$BACKEND" == "mxnet" ]]; then
         pip install mxnet;
+      elif [[ "$BACKEND" == "jax" ]]; then
+        pip install jax jaxlib;
       fi
     - pip install -e .
 
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -3,6 +3,7 @@ from ..base import unfold
 from ..tenalg import multi_mode_dot, mode_dot
 from ..tucker_tensor import tucker_to_tensor
 from ..random import check_random_state
+import tensorly.tenalg as tlg
 from math import sqrt
 
 import warnings
@@ -269,3 +270,83 @@ def non_negative_tucker(tensor, rank, n_iter_max=10, init='svd', tol=10e-5,
             break
 
     return nn_core, nn_factors
+
+
+
+class Tucker:
+    def __init__(self, rank=None, n_iter_max=100, 
+                 init='svd', svd='numpy_svd', tol=10e-5, 
+                 random_state=None, mask=None, verbose=False):
+        """Tucker decomposition via Higher Order Orthogonal Iteration (HOI)
+
+            Decomposes `tensor` into a Tucker decomposition:
+            ``tensor = [| core; factors[0], ...factors[-1] |]`` [1]_
+
+        Parameters
+        ----------
+        tensor : ndarray
+        ranks : None or int list
+                size of the core tensor, ``(len(ranks) == tensor.ndim)``
+        rank : None or int
+                number of components
+        n_iter_max : int
+                    maximum number of iteration
+        init : {'svd', 'random'}, optional
+        svd : str, default is 'numpy_svd'
+            function to use to compute the SVD,
+            acceptable values in tensorly.SVD_FUNS
+        tol : float, optional
+            tolerance: the algorithm stops when the variation in
+            the reconstruction error is less than the tolerance
+        random_state : {None, int, np.random.RandomState}
+        verbose : int, optional
+            level of verbosity
+
+        Returns
+        -------
+        core : ndarray of size `ranks`
+                core tensor of the Tucker decomposition
+        factors : ndarray list
+                list of factors of the Tucker decomposition.
+                Its ``i``-th element is of shape ``(tensor.shape[i], ranks[i])``
+
+        References
+        ----------
+        .. [1] tl.G.Kolda and B.W.Bader, "Tensor Decompositions and Applications",
+        SIAM REVIEW, vol. 51, n. 3, pp. 455-500, 2009.
+        """
+        self.rank = rank
+        self.n_iter_max = n_iter_max
+        self.init = init
+        self.svd = svd
+        self.tol = tol
+        self.random_state = random_state
+        self.mask = mask
+        self.verbose = verbose
+
+    def fit(self, tensor):
+        self.fit_transform(tensor)
+        return self
+
+    def fit_transform(self, tensor):
+        tucker_tensor = tucker(tensor, rank=self.rank,
+                               n_iter_max=self.n_iter_max,
+                               init=self.init,
+                               svd=self.svd,
+                               tol=self.tol,
+                               random_state=self.random_state,
+                               mask=self.mask,
+                               verbose=self.verbose)
+        self.tucker_tensor_ = tucker_tensor
+        return tucker_tensor[0]
+
+    def transform(self, tensor):
+        _, factors = self.tucker_tensor_
+        return tlg.multi_mode_dot(tensor, factors, transpose=True)
+
+    def inverse_transform(self, tensor):
+        _, factors = self.tucker_tensor_
+        return tlg.multi_mode_dot(tensor, factors)
+
+    def __repr__(self):
+        return f'Rank-{self.rank} Tucker decomposition.'
--- a/tensorly/decomposition/candecomp_parafac.py
+++ b/tensorly/decomposition/candecomp_parafac.py
@@ -61,7 +61,7 @@ def initialize_kruskal(tensor, rank, init='svd', svd='numpy_svd', random_state=N
             # Put SVD initialization on the same scaling as the tensor in case normalize_factors=False
             if mode == 0:
                 idx = min(rank, tl.shape(S)[0])
-                tl.index_update(U, tl.index[:, :idx], U[:, :idx] * S[:idx])
+                U = tl.index_update(U, tl.index[:, :idx], U[:, :idx] * S[:idx])
 
             if tensor.shape[mode] < rank:
                 # TODO: this is a hack but it seems to do the job for now
@@ -705,3 +705,141 @@ def randomised_parafac(tensor, rank, n_samples, n_iter_max=100, init='random', s
                     break
 
     return KruskalTensor((weights, factors))
+
+
+class CP:
+    def __init__(self, rank, n_iter_max=100, tol=1e-08, 
+                 init='svd', svd='numpy_svd',
+                 l2_reg=0,
+                 linesearch=False,
+                 fixed_modes = [],
+                 normalize_factors=False, 
+                 orthogonalise=False, 
+                 non_negative=False,
+                 sparsity = None,
+                 mask=None, svd_mask_repeats = 5,
+                 cvg_criterion='abs_rec_error',
+                 random_state=None, 
+                 verbose=0):
+        """Candecomp-Parafac decomposition
+
+            Computes a rank-`rank` decomposition of `tensor` [1]_ such that,
+
+                ``tensor = [|weights; factors[0], ..., factors[-1] |]``.
+
+            Parameters
+            ----------
+            tensor : ndarray
+            rank  : int
+                Number of components.
+            n_iter_max : int
+                Maximum number of iteration
+            init : {'svd', 'random'}, optional
+                Type of factor matrix initialization. See `initialize_factors`.
+            svd : str, default is 'numpy_svd'
+                function to use to compute the SVD, acceptable values in tensorly.SVD_FUNS
+            normalize_factors : if True, aggregate the weights of each factor in a 1D-tensor
+                of shape (rank, ), which will contain the norms of the factors
+            tol : float, optional
+                (Default: 1e-6) Relative reconstruction error tolerance. The
+                algorithm is considered to have found the global minimum when the
+                reconstruction error is less than `tol`.
+            random_state : {None, int, np.random.RandomState}
+            verbose : int, optional
+                Level of verbosity
+            return_errors : bool, optional
+                Activate return of iteration errors
+            mask : ndarray
+                array of booleans with the same shape as ``tensor`` should be 0 where
+                the values are missing and 1 everywhere else. Note:  if tensor is
+                sparse, then mask should also be sparse with a fill value of 1 (or
+                True). Allows for missing values [2]_
+            cvg_criterion : {'abs_rec_error', 'rec_error'}, optional
+                Stopping criterion for ALS, works if `tol` is not None. 
+                If 'rec_error',  ALS stops at current iteration if (previous rec_error - current rec_error) < tol.
+                If 'abs_rec_error', ALS terminates when |previous rec_error - current rec_error| < tol.
+            sparsity : float or int
+                If `sparsity` is not None, we approximate tensor as a sum of low_rank_component and sparse_component, where low_rank_component = kruskal_to_tensor((weights, factors)). `sparsity` denotes desired fraction or number of non-zero elements in the sparse_component of the `tensor`.
+            fixed_modes : list, default is []
+                A list of modes for which the initial value is not modified.
+                The last mode cannot be fixed due to error computation.
+            svd_mask_repeats: int
+                If using a tensor with masked values, this initializes using SVD multiple times to
+                remove the effect of these missing values on the initialization.
+            linesearch : bool, default is False
+                Whether to perform line search as proposed by Bro [3].
+
+            Returns
+            -------
+            KruskalTensor : (weight, factors)
+                * weights : 1D array of shape (rank, )
+                    all ones if normalize_factors is False (default), 
+                    weights of the (normalized) factors otherwise
+                * factors : List of factors of the CP decomposition element `i` is of shape
+                    (tensor.shape[i], rank)
+                * sparse_component : nD array of shape tensor.shape. Returns only if `sparsity` is not None.
+
+            errors : list
+                A list of reconstruction errors at each iteration of the algorithms.
+
+            References
+            ----------
+            .. [1] T.G.Kolda and B.W.Bader, "Tensor Decompositions and Applications",
+            SIAM REVIEW, vol. 51, n. 3, pp. 455-500, 2009.
+
+            .. [2] Tomasi, Giorgio, and Rasmus Bro. "PARAFAC and missing values." 
+                    Chemometrics and Intelligent Laboratory Systems 75.2 (2005): 163-180.
+
+            .. [3] R. Bro, "Multi-Way Analysis in the Food Industry: Models, Algorithms, and 
+                    Applications", PhD., University of Amsterdam, 1998
+        """
+        self.rank = rank
+        self.n_iter_max = n_iter_max
+        self.tol = tol
+        self.l2_reg = l2_reg
+        self.init = init
+        self.linesearch = linesearch
+        self.svd = svd
+        self.normalize_factors = normalize_factors
+        self.orthogonalise = orthogonalise
+        self.mask = mask
+        self.svd_mask_repeats = svd_mask_repeats
+        self.cvg_criterion = cvg_criterion
+        self.random_state = random_state
+        self.verbose = verbose
+
+    
+    def fit_transform(self, tensor):
+        """Decompose an input tensor
+
+        Parameters
+        ----------
+        tensor : tensorly tensor
+            input tensor to decompose
+
+        Returns
+        -------
+        KruskalTensor
+            decomposed tensor
+        """
+        kruskal_tensor = parafac(tensor, rank=self.rank,
+                                 n_iter_max=self.n_iter_max,
+                                 tol=self.tol,
+                                 init=self.init,
+                                 svd=self.svd,
+                                 normalize_factors=self.normalize_factors,
+                                 orthogonalise=self.orthogonalise,
+                                 mask=self.mask,
+                                 cvg_criterion=self.cvg_criterion,
+                                 random_state=self.random_state,
+                                 verbose=self.verbose,
+                                 return_errors=True)
+        self.kruskal_tensor_ = kruskal_tensor 
+        return kruskal_tensor
+
+    def fit(self, tensor):
+        self.fit_transform(tensor)
+        return self
+
+    def __repr__(self):
+        return f'Rank-{self.rank} CP decomposition.'
--- a/tensorly/decomposition/mps_decomposition.py
+++ b/tensorly/decomposition/mps_decomposition.py
@@ -81,3 +81,47 @@ def matrix_product_state(input_tensor, rank, verbose=False):
         print("MPS factor " + str(n_dim-1) + " computed with shape " + str(factors[n_dim-1].shape))
 
     return factors
+
+
+class TensorTrain:
+    def __init__(self, rank, verbose=False):
+        """MPS decomposition via recursive SVD
+
+            Decomposes `input_tensor` into a sequence of order-3 tensors (factors)
+            -- also known as Tensor-Train decomposition [1]_.
+
+        Parameters
+        ----------
+        input_tensor : tensorly.tensor
+        rank : {int, int list}
+                maximum allowable MPS rank of the factors
+                if int, then this is the same for all the factors
+                if int list, then rank[k] is the rank of the kth factor
+        verbose : boolean, optional
+                level of verbosity
+
+        Returns
+        -------
+        factors : MPS factors
+                order-3 tensors of the MPS decomposition
+
+        References
+        ----------
+        .. [1] Ivan V. Oseledets. "Tensor-train decomposition", SIAM J. Scientific Computing, 33(5):2295–2317, 2011.
+        """
+        self.rank = rank
+        self.verbose = verbose
+
+    def fit_transform(self, tensor):
+        self.tensor_train_ = matrix_product_state(tensor, rank=self.rank, verbose=self.verbose)
+        return self.tensor_train_
+
+    def fit(self, tensor):
+        self.fit_transform(tensor)
+        return self
+
+    def __repr__(self):
+        return f'Rank-{self.rank} Tensor-Train decomposition.'
+
+
+    
--- a/tensorly/tests/test_backend.py
+++ b/tensorly/tests/test_backend.py
@@ -560,11 +560,11 @@ def test_index_update():
     insert = tl.tensor(np.copy(np_insert))
 
     np_tensor[:, 1:3] = np_insert
-    tl.index_update(tensor, tl.index[:, 1:3], insert)
+    tensor = tl.index_update(tensor, tl.index[:, 1:3], insert)
     assert_array_equal(np_tensor, tensor)
 
     np_tensor = np.random.random((3, 5)).astype(dtype=np.float32)
     tensor = tl.tensor(np.copy(np_tensor))
     np_tensor[2, :] = 2
-    tl.index_update(tensor, tl.index[2, :], 2)
+    tensor = tl.index_update(tensor, tl.index[2, :], 2)
     assert_array_equal(np_tensor, tensor)


aarmey:
Excellent! Just specifically regarding the functions to implement, it might be helpful to look at [`sklearn.decomposition.PCA`](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html). I'm not sure what `predict()` would be for CP or Tucker, but `fit()`, `fit_transform()`, and `inverse_transform()` would all be useful. I distinguish `fit()` and `fit_transform()` simply as the first doesn't return anything (but saves the result internally), while the second passes back the result. I'm less familiar with this use, but it would be helpful to think about what is passed during supervised regression when using CP or Tucker. `fit_transform()` is often chained with a regression method (the output feeds directly in to the supervised method) and it's quite valuable to have in that case.

cohenjer:
Thank you for launching that discussion @JeanKossaifi. I think several people (@MarieRoald for instance) including myself have started to work in that direction with various solutions. Here is what I got for the CP:
```python
class Parafac:
    """
    A class for defining a Parafac model and an optimization technique.
    This class allows to choose the optimization method, the initialization
    method, constraints on each mode, fixed modes, and various optimization
    parameters such as the number of iterations, gradient steps and so on.
    """

    def __init__(self, rank=1, method='ALS', init='random', constraints=[],
                 weights=None, fixed_modes=[], n_iter_max=100, tol=1e-8,
                 verbose=False, svd='numpy_svd', random_state=None):
 
        self.rank = rank
        self.init = init
        # and so on
        # Choosing the default method based on the constraints

    def initialize_parafac(self, data=0):
        """
        Compute initial factors for the Parafac model, stored in attribute
        self.init_factors
        """

    def fit(self, data):
        """
        Parafac.fit(data) for testing various initializations.
        """
        # Initialize factors if needed
        self.initialize_parafac(args)
        # Call the method

    def normalize(self):
        """
        Normalizes the components column-wise and returns the weights in self.weights.
        """

    def reconstruct(self):
        """
        returns the resulting tensor of the fitted Parafac model
        """

    def transform(self, data, factors=0, rank=0):
        """
        transforms an input tensor by projecting it on the span of the
        estimated factors using the current Parafac model.
        """
```

I am not very familiar tbh with scikit-learn, but I think while it is great to try and follow the principles listed by @JeanKossaifi, there are specificities of Tensor methods that mean we should part from some concepts.

A tentative (expert) user experience might be the following:

```python
    # Choosing the Alternating Nonnegative LS optimization method
algo = 'ANLS'
    # Choosing hyperpameters
alpha = 0.5 # just some random examples, there are not really hyperparameters in ANLS
delta = 1e-2
    # Fixed mode for instance
fixed = [0];
    # Constraints
constraints = ['NN','NN','NN']   # nonnegative parafac
    # creating the class instance
model = tl.Parafac(rank=cp_rank, verbose=True, init='random',
                     random_state=random_state, fixed_modes=fixed,
                     constraints=constraints, method=algo, tol=1e-5, alpha=alpha,
                     delta=delta)
    #  Modify init by hand for expert user for instance
model.init_factors = tl.abs(model2.init_factors)
    # Fitting
factor_fg, errors_fg = model.fit(some_data)
    # Reconstruct the data approximately from the factors
cp_reconstruction = model.reconstruct()
```
while someone doing his first CP might simply write

 ```python
# Creating the model and fitting
model = tl.Parafac(rank)
model.fit(some_data)
# getting the factors of the CP
facs = model.components
```


### Transform, reconstruct, predict and such.
I would say it is important to differentiate between (using my vocabulary):
- fit : the process of finding the CP decomposition from a given tensor, in a specific setting (initialization method, optimization algorithm, hyperparameters...). This should be the focus I think for CP.
- reconstruct : Given some Kruskal tensor, build the full tensor. That method is already implemented in tensorly, that could be a simple wrapper, but this is useful conceptually to have it as a method of the CP class I think. 
- transform: Given a CP model trained from a data set, do something to another data set. As far as I know, this is something that is not usually done with CP (but I simply may be unaware of that); while it is definitively relevant for Tucker. At least, we would require the user to specify what he wants to do with the trained CP (e.g. setting a factor matrix in the CP of another tensor to a pre-trained one? then on which mode...). 

As far as I am concerned, CP does not need transform() [or what @JeanKossaifi called predict()] because if does not seem very useful and there are only complicated possible definitions. But for Tucker it could be provided.

### Initialization
My opinion is that there should be, on the contrary to what @JeanKossaifi suggested, an initialization routine with many options (like choosing the initial distribution for random initialization). But that function would not need to be called by hand, and rather using a fit() call should automatically call the initialization with good pre-determined values if initialization has not been carried out already. This means that for people running a single decomposition, who are not familiar with CP and tensors, everything would be preset and simple. On the other hand, for expert users, they can easily initialize outside the fit() function which may be nice for running several experiments with the same code base.


To sum up, I propose to have a separate initialization method, a reconstruction method (that you guys may call transform() because of scikit-learn?) and also some other useful things like normalization of the factors estimated in the CP. Or is this already too many routines? Would be glad to have your feedback as well.

**Comment by MarieRoald** [comment]
Thank you for creating a place for this valuable discussion, @JeanKossaifi!

I think this is the correct path for TensorLy moving forward. I have, as @cohenjer noted, worked on an object-oriented tensor framework in Python for my personal use, and I have gained a fair share of experiences from working on it.


## Initialisation method

Regarding the single-versus-individual arguments, I think we should support a single input argument that specifies which initialisation scheme to use. This argument could be a string (e.g. `"random"` and `"SVD"`), a function that takes a data matrix as input and returns a decomposition or a precomputed decomposition.

Moreover, I think that we, to follow the scikit-learn API closely, should input the initialisation parameters to `__init__`, similar to how it is implemented in [`sklearn.cluster.KMeans`](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html). According to the [scikit-learn developer guide](https://scikit-learn.org/stable/developers/develop.html#apis-of-scikit-learn-objects), the `fit` method should only take the dataset and data-dependent arguments as input. 

I am not sure if I follow you completely, @cohenjer. If I understand @JeanKossaifi's example correctly, it still allows for a separate `_init_decomposition` method that is called by fit but not part of the public API. Then, if we have initialisation parameters as an argument for the `__init__` method, these parameters can be passed to `_init_decomposition` during the `fit` call, giving us full control. Moreover, if one wants to change the initialisation scheme between fits, it's enough to change the `init` attribute before calling `fit` the second time. Like how `GridSearchCV` work in scikit-learn (see [this](https://scikit-learn.org/stable/developers/develop.html#parameters-and-init) part of the developer guide).


## Input to the decomposition

I agree with @cohenjer's idea about a list of constraints for the different modes. However, I believe that we should consider a list of dictionaries, instead of strings. Handling constraints with dictionaries will widen the possibilities of new constraints, hard or soft (such as sparsity, which is requested in #169), without adding too many possible keyword arguments or complicated string parsing. I say this from experience, as I wish my codebase used a list of dictionaries. Below is an example of what I imagine

```python
constraints = [
   {'orthogonality': True},
   {'non_negativity': True, 'l2_penalty': 1e-3,},
   {'l1_penalty': 1e-2}
]
```

Also, using dictionaries for constraints makes it easier to ensure that they are valid; we can have a `check_constraints` method, which ensures that incompatible constraints aren't imposed (such as nonnegativity and orthogonality).


## Class structure

I have some thoughts on the class structure based on my experience with my codebase and discussions with @yngvem that may be useful to the discussion.

### Separate decomposition and data fitting classes

For the class structure in my implementation, I decided to separate between the decomposition and the class that computes the decomposition (so one KruskalTensor class and one Parafac class). My motivation was that I wanted to support multiple ways of computing the same decomposition (ALS, randomised ALS and weighted all-at-once optimisation). Especially all-at-once optimisation and ALS work very differently, so inheritance doesn't make much sense. Therefore, I decided to separate the fit-method and the decomposition in two separate classes.

Moreover, suppose we separate the fitting-procedure and the decomposition object in two classes. In that case, it is easier to have a set of decomposition-specific methods (e.g. fast computation of Frobenius norm with KruskalTensors). 

A drawback of the above class structure is that it does not follow the Scikit-learn principle of non-proliferation of classes. After all, it would mean creating a class that represents something other than a learning algorithm. However, I think it's worth at least considering for two reasons:

1. Scikit-learn was likely designed for analysis of one mode in two-way data. This is apparent in the PCA class, where the `PCA.components_` represent the loadings, and the scores aren't stored (or at least not in the documented API). Moreover, the only methods I am familiar in Scikit-learn that performs labelling of both rows and columns is the spectral biclustering and spectral coclustering classes. However, these do not have a transform method. 
2. There are many decomposition-specific methods. For example, it may be useful to have a seamless way to compute scores like the core consistency for both Kruskal tensors and PARAFAC2 tensors, and it may therefore make sense to have a `core_consistency` method as part of the decomposition class API. Then we would get `PARAFAC2.core_consistency` and `KruskalTensor.core_consistency` instead of separate `kruskal_core_consistency` and `parafac2_core_consistency` functions (I'm not sure if we want to implement core consistency, but the same argument holds for other decomposition-specific methods)

Alternatively, we could have one class that takes care of both fitting and storing the decomposition, and let that class have a wide range of possible methods. In that case, it may be worthwile to look at the [Prince](https://github.com/MaxHalford/prince) library, which is a library for two-way factor analysis that follows the scikit-learn API.

### Choice of optimisation algorithms

One way to structure the decomposition class to allow for different optimisation methods is to have one decomposition superclass and implement the fit-method separately in, for example, two sub-classes, one block-coordinate-descent class and one all-at-once class.

Now, having separate classes for block-coordinate-descent and all-at-once optimisation may not be the easiest for the user. How would they choose which PARAFAC class to use? We can avoid this problem by, for example, having the optimisation-specific classes hidden from the user and then automatically selecting the correct class based on the inputs to the public API. This would be similar to how the `scipy.optimize.minimize` function works, which will decide which function in `scipy.optimize` to call based on the input arguments.

However, I believe @cohenjer has been working on a Python project for tensor learning with different optimisation algorithms, so he may have some valuable insight here.

yngvem:
I just have a few things to tag onto @MarieRoald's notes.

## Decomposition dependent methods

We may also want to discuss which methods we want to implement in our decomposition classes. Do we want to implement decomposition-specific methods (such as Frobenius norm) as methods of the decomposition classes? That way, when `tensorly.norm` is called, it can check if a fast norm-implementation is available (e.g. as a `Decomposition.__tensorly_norm__(...)` method or a `Decomposition.__tensorly_function__`) . Here, we may be able to get some inspiration from the dispatch mechanism in NumPy ([NEP-19](https://numpy.org/neps/nep-0018-array-function-protocol.html)). 

## Scikit-learn rules

Scikit-learn classes follow some, at first glance, odd choices. For example, in Scikit-learn, the `__init__` function should not perform any logic other than storing the arguments as attributes with the same name. All input checking happens in the `fit` function. Moreover, we should be able to set all parameters in `__init__`. This class structure is chosen to make it easier to have a grid search without creating new class instances.

Another interesting part of the Scikit-learn API is that all attributes computed in `fit` has a trailing underscore. This makes it easy to separate between the model parameters and the computed parameters! 

## The transform/predict method

Finding a single API that fit all tensor decomposition models will be tough, so I think that we probably should have a list of methods that each represent different things. Then, for a given decomposition, we can implement the relevant subset of these methods.

I agree with @aarmey's comment that `fit_transform` should be informed by what is passed during supervised regression. Following this thought, the only way I can imagine implementing a transform method that fits in the scikit-learn ecosystem is with a method that returns the factor matrix for the first mode.

Let us consider a 3-way CP tensor,   
*X = [A, B, C]*,   
where *A, B* and *C* are factor matrices. *A* represents the sample-mode. Next, we and we have the following code:

```python
X = ...
X_train = X[:50]
X_test = X[50:]
parafac = tensorly.Parafac(...)
A_train = parafac.fit_transform(X_train)
A_test = parafac.transform(X_test)

# use A_train and A_test for a classification or regression problem
```

With the above code, `A_train` should (hopefully) correspond to the first 50 rows of *A* and `A_test` should correspond to all subsequent rows of *A*.

For the `transform` method to work this way, the `parafac.transform(X_test)` call will need to compute the solution to
*min\_{A_test} ||X_test - [A_test, B, C]||_F^2 ,*
keeping the factors for *B* and *C* kept constant equal to their value when fitted to the training dataset.

By structuring the transform method this way, we ensure full compatibility with the scikit-learn ecosystem, and can create regression objects using the pipeline, like this

```python
pipe = Pipeline(
    memory=None,
    steps=[
        ("parafac", Parafac(rank=4)),
        ("regressor", Ridge(alpha=0.1)),
    ],
)
pipe.fit(X, y)
```

to fit a PARAFAC-model where the sample-mode (first mode) factor matrix is used as input to a ridge regressor.

Unfortunately, this way of implementing the transform method is not easy to implement for all decomposition types. For PARAFAC2, the order of the modes affects the decomposition. Consequently, the transform method would need to know which mode it is that represents samples and select the correct transformation method accordingly.

The `predict` method is reserved for regression and classification algorithms, so I don't think we should have a `predict` method on the classes that compute tensor decompositions.

JeanKossaifi:
Thanks all for the great discussion! :D

Just summarizing some of the main points: 

* Following scikit-learn API: @cohenjer, one reason I took scikit-learn as an example is that it is well established and most users are familiar with its API (40K+ stars and 20K+ forks). I agree that we do not need to blindly follow their API but they have a set of very nice abstractions (e.g. the Pipelines, @yngvem's example is great!), fine-tuned through several years of experience. They also have some nice [enhancement proposals](https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/).

* Separation between data and algorithm (@MarieRoald ): I 100% agree, it is a good programming principle in general. Here, this means that the classes should not store the training data, only their own parameters. In particular, KruskalTensor is a separate class from the `Parafac` one. One issue with mixing the two is that it introduces lots of side effects makes is much harder to compose function (e.g. to create Pipelines as @aarmey  mentioned). In that way the reconstruction from the decomposed tensor could be left outside the actual decomposition `Parafac` class.

* core_consistency (@MarieRoald ) : that sounds sensible — what would be the signature? What exactly would this compute?

* Initialization: I really like @MarieRoald's approach. A good reference is scikit-learn’s non-negative factorisation (https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.NMF.html). The idea is to have an init param in the __init__ (a string specifying which method to use), and we can have another param in the fit and fit_transform that provides an optional actual init for the decomposed tensor (either `kruskal_tensor_init=None` or `factors=None` and `weights=None` separately).

* Choice of optimisation algorithms (@MarieRoald ): I agree with your subdivision: the main class (e.g. `Parafac`) could have an argument (`method`?) specifying which optimisation method should be used (e.g. `ALS`, etc). This main class will then dispatched to sub-classes, specialised in one such optimisation method (e.g. `ParafacALS`).

* About `predict` (@cohenjer and @yngvem): you are absolutely right, we shouldn’t have a `predict` for decompositions, I meant to write `transform`!

* Updated glossary of the methods:
    * `fit(tensor, y=None, kruskal_init=None)`: returns `self` —  learning  the parameters of the *model* 
    * `fit_transform(tensor, y=None, kruskal_init=None)` : returns `decomposed_tensor` — fit the model and return the resulting decomposition 
    * `transform(tensor, y=None, kruskal_init=None)` : returns the projection onto the decomposition space — useful for e.g. Higher-Order PCA, etc when we can project an input tensor to a learnt subspace (see @yngvem's example for CP, for Tucker this would project a tensor using the factors and return a core tensor)
    * `inverse_transform(projected_tensor)` : returns back a full tensor — reconstructs the original tensor from the decomposition, see `tranform` above.



Note that the PR includes a rough draft of classes, e.g. for Tucker:
https://github.com/tensorly/tensorly/blob/be41f195cd0600b70ec66ac48ca0c4c083b1e42d/tensorly/decomposition/_tucker.py#L276

Since you have already implemented a class API, @marie and @yngvem and so have experience (your feedback is really on point, thanks!), wondering whether you want to add to the PR / or even create a separate one (once we have converged to a finalise API I guess)?

We can also have a separate PR for the Pipeline?   I will also update the main comment of the PR to reflect this discussion.  What do you all think?

JeanKossaifi:
Making a separate comment for important considerations that are not directly tied into the Class API — I think we can create an issue for each, they are all important API considerations.

* Constraints (@MarieRoald, @cohenjer) — this is a hard one!
I completely agree: we need something flexible. I am not certain that a single dictionary for all (including l1 and l2 regularisation) is the best way to expose this, mostly as it is not easily discoverable and the resulting dict could become quite complex — I would tend to prefer keyword arguments, which could each be a dictionary ? (e.g. keys being the modes and values being the corresponding value, such as `orthogonality = {0:True, 2:True}`, and, as suggested by @cohenjer  we could have an additional `constraint = {0:’NN’, 1:’NN’}` to add e.g. non-negativity only to some modes). We can add this later onto the class API though, so this is not blocking. It also involves quite a bit of new code (@cohenjer  has already done a lot there, and is leading an exciting project with INRIA to push the CP further and adding lots of cool new features!).

* Decomposition dependents methods (@yngvem )— that is a really good point. I wonder whether this should be done automatically or whether each algo should call the appropriate one (e.g. kruskal_tensor.norm)?

JeanKossaifi:
I guess the big question that remains: should we use `fit_transform` or add a separate `decompose` function specific to tensor decompositions?

cohenjer:
The discussion is already getting a bit complex :) Let me try to structure/summarize things a little bit and share my opinion when relevant (hopefully). I apologize for the long message.

## Pure API
This section discusses the API design only, not the addition of new methods/features to tensorly.

### Classes for representation vs computation
Having two sets of classes is a great idea and probably a must-have. Even among researchers, people often confuse the ''representations'' (i.e. a tensor in a kruskal format for instance) with an algorithm that computes a (possibly approximate) decomposition of some data tensor. For Parafac, this means basically having:
- a kruskal_tensor class (which is already implemented)
- a parafac_compute class that creates an instance of a (problem definition+solver) object

This is also a great way to make tensorly almost pedagogic for new users. 

I think this needs to be however careful designed so that it is crystal clear which class is which. I am not sure we can take inspiration from Scikit-learn on that however; they often do not make that distinction. Probably one of the most important points is to choose good, clear names for these classes and use a consistent terminology for all tensor decompositions. For instance
- ``parafac_tensor``, ``tucker_tensor``, ``mps_tensor``, ``parafac2_tensor`` ... would all be representation classes
- ``parafac``, ``tucker``, ``mps``, ``parafac2`` ... could all be computation classes.

I am not sure my terminology is the best, but probably it would be nice to settle for something consensual. I think ''representation'' or ''format'' are quite well-understood concepts. We may find better names for computation class (problem instance class ?). 

### Initialization
I think we all agree that @MarieRoald's solution is the way to go, and this is also what I had in mind. I was probably not being very clear, but to me what matters is that a user can easily access, monitor and change the initialization (like changin the method, extracting the initial factors or changing them manually). 

### fit, fit_transform and transform
Not being used to scikit-learn, I am not sure I understand why this seems to be such an important point. For Parafac, I really don't think transform (as defined in the last message of @JeanKossaifi ) would be needed as it makes little sense. As @yngvem noted, at least it would be required to add an additional argument specifying which mode uses the result of the previous fit() call.

### Representation specific methods 
This is a great idea @yngvem. We would need to work on a list of already existing methods in tensorly that are easier to compute with tensors in a specific format, such as mode_dot (kruskal tensors already have their own mode dot in fact in tensorly !)

### Subclasses for groups of algorithms
This is tighly linked to the discuss below, but on the particular matter of dividing ''computation'' classes into subclasses for families of methods, I am not yet convinced this is necessary. We might as well work with nested functions, like a ``alternating`` function  which would be a general wrapper; in fact mostly an outer-loop calling mode-wise updates such as ``least_squares``, ``gradient step`` or such mode-wise update methods. 

At least this is how I have designed the first bits of codes for my personnal branch of tensorly, and I have not found this organisation of methods to be limiting. But maybe subclasses have advantages compared to nested functions that I am missing? 


### Pipelines
Nothing to add here.

## API with enhancements along the line 
This discusses improvements and additions to tensorly of various importance that we may tackle simultaneously to the development of the class API. (this is where the [project](https://cohenjer.gitlab.io/tensoptly-website/) @JeanKossaifi mentionned comes into play)

### Constraints formulation
This is indeed a quite complicated story. There are a few very important constraints that I meet all the time when considering tensor decompositions:
- nonnegativity,
- sparsity (l_p penalty with p<=1 for instance, but there are others),
- orthogonality,
- unit norm
- Structured factors (Hankel, Toplitz and such).

For these constraints in particular, there should be dedicated algorithms, and low-level optimization routnies such as a nonnegative least squares solver(s? there exist several methods with different benefits!) or a Lasso solver(s? same remark).

But there are many other possible constraints, and we cannot have dedicated algorithms for any constraints, or even for combinations of usual constraints. Also constraints may sometimes be imposed on only a subset of modes as @JeanKossaifi mentionned. Thus, we may implement an algorithm such as alternating proximal gradient along with a great variety of proximal operators (projection-like operators that allow to enforce constraints). We can also allow any user to use his own proximal operator to enforce a constraint of his choice.

All things considered, this means that we need to write a good parser, and think carefully about a data structure to store the links between constraints and available decomposition algorithms. I am not sure a bunch of keyword will do the trick.

For instance, suppose a user wants to compute a parafac decomposition with nonnegativity constraints and sparsity.  We then should be able to list the available techniques for that combination of constraints (probably something like proximal gradient and ANLS, maybe other variants) and provide a default choice. Now if the user also wants to add orthogonality (which contrary to what @MarieRoald suggested are not imcompatible, they simply impose that the factors are 1-sparse), this means changing the algorithm completely, actually ignore the sparsity constraints (which is useless since the only solutions arte 1-sparse anyway) and propose a list of available methods (probably none unless someone takes time to adapt one of the few methods for othogonal NMF in the tensor case).

### Algorithms
I tried to list the various algorithms used to compute parafac in tensorly at the moment:
- ALS, with Bro's line search. Has a few things for sparsity and orthogonality but I am not confident these are properly justified in terms of convex optimization.
- CPRAND (T. Kolda and collaborators recent work): the only randomized algorithm in tensorly? Also current implementation does not authorise constraints.
- Multiplicative updates for nonnegative parafac.

I think these algorithms are all over the place, which was fine for early developments stages of tensorly. But I hope we can improve things. A first step is to properly characterize which algorithms are available for implementation. I would divide algorithm features in several, non-exclusive categories:
1/ Algorithm type
There are basically three families of algorithms for computing tensor decompositions:
   - Direct algorithms (often based on SVDs). These are not represented in tensorly at the moment.
   - Alternating algoritms, such as ALS, CPRAND, AOADMM and such.
   - All-at-once algorithms, such as Gauss-Newton, Levemberg Marcquart or conjuguate gradient on all modes. Again these are not presently implemented.

2/ Extrapolation
Most of fast methods for convex optimization nowdays feature extrapolation. It can be used in alternating algorithms and all-at-once algorithms alike. New results are starting to appear in that direction, but discussing it here would take too much space. Let's just say we can do better than having only Bro's line search for ALS.

3/ Randomization
CPRAND is just one of many recent randomized algorithms; some of which allow for constraints of various types (while CPRAND allows only constraints leading to solvable constrained least squares problems). I don't see any reason to only have CPRAND.

4/ Constraints-specific design
Some constraints such as nonnegativity, Hankel factors or orthogonality may call for specific algorithms

Part of the project I will work on with a dedicated engineer will be to sort out these methods and implement them in tensorly, using the class API. I just wanted to list these things here so that we have all those aspects in mind when designing the API. 

### Algorithms hyperparameters
A difficult point in terms of API design for us is that different optimization methods have different hyperparameters. We may want to have a dictionary of specifications. For instance (examples are incomplete, just for the sake of discussion) for extrapolated gradient descent mode-wise,
```python
params = [ {'stepsize' : 0.1 }, {'extrapolation_step' : 0 }]
```
while for ANLS,
```python
params = [ {'inner_iterations' : 20 }, {'inner_stop' : 1e-6 }]
```
Some parameters are shared across all methods like a stopping criterion (max iterations, tolerance ...), which could also be part of such a dictionary? This matter is gonna get more important if several methods are added to tensorly with possibly many hidden hyperparameters.

### Other decompositions
We should also keep in mind that some methods are nested. For instance, a Tucker tensor is a particular MPS tensor, and a Parafac tensor is a particular Tucker tensor. I wonder how we could use this fact to our benefit?

### Other format-specific methods
Core-consistency is one of those methods that help with chosing the rank of the approximate tensor when performing parafac. There are many such ''utility'' functions, that would belong to the ''computation''classes I think. They are not a priority in my opinion.

yngvem:
Thank you @ JeanKossaifi and @cohenjer for the comprehensive summaries. I still have a few questions I think are worth to discuss, so I will try structure them as possible choices with pros and cons. 



I firmly believe that we should do our best to follow the scikit-learn API. That way, we encourage people to learn tensor methods, and otherwise, it may be challenging to promote widespread use of TensorLy. Just about everyone doing data science in Python are familiar with the init->fit philosophy of scikit-learn, and I think we should cater to that audience. 

Now, where I am uncertain, is on the behind-the-scenes part. It may be worthwhile to implement a separate decomposition class too. I am not sure what's the best here, and I honestly think that depends on what exactly we want TensorLy to provide. *TensorLy's functionality should inform the exact class structure we want to provide!*

# What do we want to provide

## Helper functions

I think both of you sort of misunderstood Marie's point in her previous post. In my understanding, she wanted to know if we plan on implementing decomposition-specific helper functions, *such as* the Core Consistency. Another possible helper-function is the leverage score. Are we planning on supporting such methods? If so how?

We can either have all helper functions implemented as methods for the `KruskalTensor` class (whether that class also takes care of the fitting procedure or not). Then, the function signature would be akin to

```python
ktensor.core_consistency(data_tensor)
```

Alternatively, we can have a module that provides this functionality, in that case, the signature would look like

```python
tensorly.utils.kruskal_tensor.core_consistency(decomposition, data_tensor)
```

where decomposition could either be a Kruskal tensor object or possibly a tuple containing factor matrices.

### Possible choices:

- **Don't implement helper functions.**

  - Pros:
  - Less bloated codebase
    - More maintainable code
    - Easier to understand the scope of TensorLy if it only provides decomposition functions

  - Cons:

    - Keeping all (or much of the) multi-way functionality in one project can be beneficial for the users.

- **Provide functionality through methods of the decomposition classes (e.g. `KruskalTensor.core_consistency(X)`:**

  - Pros:
  - All decomposition-based functionality is kept in one place, possibly making it easier to find for end-users.
    - Decomposition-independent API for metrics that represent the same thing, but have different implementation details based on which decomposition-type it's used for

  - Cons:

    - Bloated classes can make the code more difficult to maintain
    - If we provide too much functionality directly in the class, then it may be challenging to navigate the documentation.

- **Provide functionality in separate modules**

  - Pros:
  - Straightforward API, if I want a metric for a Kruskal tensor, then that method is found in `tensorly.metrics.kruskal_tensor`
    - Possibly more maintainable code, since we can avoid having a 1000+ lines class definition.
    - More clear API, everyone understands that `tensorly.metrics.kruskal_tensor.core_consistency` is implemented different from `tensorly.metrics.parafac2_tensor.core_consistency`.

  - Cons:

    - We end up with decomposition-specific functions for metrics that represent the same quantities.
    - Lesser degree of abstraction, but that is not necessarily negative.

## Fast computations?

As mentioned earlier, we can implement specific fast algorithms for standard tensor functionality. E.g. the norm. How do we want to implement it? Do we want to implement it?

### Options

- **Don't implement it**

  - Pros:

    - Less bloated codebase
    - More maintainable code
  - Cons:

    - Cumersome to implement oneself if it is ever needed.

- **`KruskalTensor.norm()`**

  - Pros: 
  - Very clear that it is a custom norm method for the Kruskal tensor.

  - Cons:

    - May be confusing if we don't support this functionality for all decomposition classes.

- **`tensorly.fast_algorithms.kruskal_tensor.norm(kruskal_tensor)`**

  - Pros:

    - Very explicit code, whenever we want to use it, we can use a from-import.
  - e.g. from `tensorly.fast_algorithms.kruskal_tensor import norm as ktensor_norm`.
  - Cons:

    - It may be difficult to remember what all the different modules are.

- **`tensorly.norm(kruskal_tensor)`, interfaced through a `KruskalTensor.__tensorly_norm__` method.**

  - Pros:
  - We can write seamless code without worrying about what decomposition we are working with
    
  - Cons:
  - May be confusing if we don't support this functionality for all decomposition classes.
    - Explicitly saying that we want to compute the norm fast may lead to more. 



# Deciding public-facing API

- **Only one class, both for fitting and for storing the decomposition**
  - Possible implementation details:

    - Essentially implemented as a single class that contain the decomposition
    - `__init__` stores options for fitting the model
    - `fit` takes tensor as input and decomposes it
  - Pros

    - Strictly follows the scikit-learn API

  - Cons

    - May lead to unmaintainable code
  - Notes

    - If we go with this option, then we should likely not have decomposition specific tensor-methods (such as norms) or decomposition specific helper function (such as core consistency) as methods of the class, but functions in a separate submodule.

- **Two classes, one for fitting and one for storing the decomposition**
  - Implementation details:

    - To decompose a tensor, first form a `Parafac` object.
    - When `Parafac.fit` is called, the `decomposition_` attribute, which contain the decomposition is set
  - Pros

    - Clear separation of fitting procedure and decomposition

  - Cons

    - Breaking the scikit-learn API (but only slightly)
  - Notes:

    - Although this goes against the scikit-learn philosophy of non-proliferation of classes, I don't think it is a big problem. Because with scikit-learn, it is still common to get attributes that end with an underscore.



# Optimisation specific details

How the optimisation procedure is implemented should perhaps be decided after (or seperately from) we have agreed on a public-facing API. It should be easy to use TensorLy without knowing anything about optimisation algorithms!



The optimisation algorithm API is essential. And @cohenjer's points are very interesting. However, I believe that we should have a separate discussion where the optimisation backend is in focus.

MarieRoald:
Thank you for the overview @yngvem! You are right that I didn't intend to discuss whether we should implement core consistency specifically. That was just the first example I could think of for a decomposition specific method. 



Regarding @JeanKossaifi's question

>  I guess the big question that remains: should we use `fit_transform` or add a separate `decompose` function specific to tensor decompositions?

I think that depends on how and if we want to be compatible with the scikit-learn API. If we want to be compatible with the scikit-learn pipeline, then we should implement the `transform` and `fit_transform` methods so their output can be the input of the following estimator. Thus, if the transform method of a Tucker model returns the core tensor, then Tucker decompositions cannot be chained directly before scikit-learn estimators (but they can be chained directly before other tensor decomposition methods). We should probably make this clear in the documentation, and ensure that it is easy to see which TensorLy classes fit with scikit-learn estimators and which fit with other tensor decomposition models. Also, I agree that an eventual pipeline should have its own issue/discussion, but we should probably have pipelines in mind when we are discussing the decomposition API.



Also, I believe that fit_transform should yield the same output as first calling fit, then transform on the same data, if we want to follow the scikit-learn API ([see guidelines](https://scikit-learn.org/stable/developers/develop.html#apis-of-scikit-learn-objects)). If my understanding is correct, then

```python
decomposition.fit(X)
out = decomposition.transform(X)
```

should do the same as

```python
out = decomposition.fit_transform(X)
```



Regarding a separate `decompose` method, we need to consider how it would be different from a call to `fit`. One possible way would be to have the `transform` method work as above. Then, possible behaviour of a `decompose` method could be: it has the same signature as `transform`, but it returns a full decomposition instead of only a single factor-matrix/core tensor/etc. 



An alternative is not to have a `decompose` method. Instead,  the `fit`-method can create a `decomposition_` attribute that contains the fitted decomposition. Then, whenever someone wants to use that decomposition, they can write

```python
decomposer.fit(X)
decomposition = decomposer.decomposition_

# Do something with the decomposition object
```

aarmey:
@yngvem's summary is excellent here. I think that scikit-learn has made some good choices regarding these tradeoffs. 

Implementing one class that both stores the decomposition and is capable of creating it means that the user does not have to make this distinction. They only have to inspect one object to find all of tensorly's capabilities.

The way I think of the scikit-learn classes is that they are a central organizing object for a variety of methods implemented elsewhere. The fast performance, decomposition options, and helper methods are important features of tensorly. Putting these in separate modules would help readability, testing, and continuing to develop new capabilities. Each is exposed through the class interface. Having this separation between interface and implementation would make it easier to incorporate other library's implementations. For example, many of scikit-learn's classes can optionally use well-developed external implementations such as `libsvm`, `liblinear`, etc. Finally, this choice would make it easier to incrementally develop this new interface without breaking the old one until we choose to.

@MarieRoald brings up some good points above. I struggle to see the useful distinction between `fit()` and `decompose()`, but it is unclear what `transform()` should return to make it useful for the Pipeline API. I've had this issue with scikit-learn's methods as well where, for example, we wanted to cluster the rows instead of columns of a dataset, so we had to make a transpose method to insert between KMeans and regression in a pipeline. Perhaps a solution would be to have `transform()` always return the full decomposition, but then make a separate class that allows you to select a portion of that decomposition for other estimators? One would then write a tensor regression pipeline as PARAFAC.transform() → SelectMethod → Regression.

JeanKossaifi:
Thanks all for the invaluable discussion, it's really great to see the TensorLy community coming together! :D


--------------------------------


> I firmly believe that we should do our best to follow the scikit-learn API. 

@yngvem @aarmey -- I completely agree, having come from scikit-learn myself, I tried to do this from the start (e.g. for tensor regression). For decomposition it was not as straightforward so I started with the functional interface but adding classes was always in the roadmap, great to finally add it!


# What do we want to provide

These are great points, thanks for putting all the options and pros/cons there @yngvem ! Let me start this by saying I absolutely think we should provide both helper functions and efficient operations on decomposed tensors (fast computation). Actually @yngvem, @MarieRoald I think we share a very similar vision of what the API look like :) 

## Current API

Just to summarize, here is the current API 
* `tensorly.backend` : all array functions, making tensorly compatible with most array libraries (PyTorch, NumPy, ...)
* `tensorly.XXX_tensor` (e.g. kruskal_tensor, tucker_tensor, mps_tensor, etc): functions for manipulating decomposed arrays. This includes efficient operations.
* `tensorly.metric` (.regression and .classification) : metrics between pairs of tensors (currently doesn't handle decomposed tensors)
* `tensorly.decomposition` : all the decompositions
* `tensorly.regression` : regression models (these need to be extended to classification as well)
* `tensorly.backend` : all the tensor algebra functions
* `tensorly.random` : random tensor creation
* `tensorly.testing` : utils for (unit)-tests.

## Helper functions 

I think they definitely should be in TensorLy, and currently could live in each tensor decomposition's module (e.g. `tensorly.kruskal_tensor`, etc).

Do you think it needs to be in a separate submodule (e.g. `metrics`) or shall we just leave it in `xxx_tensor`?


## Fast computation
These should definitely be in TensorLy, and some are already implemented in their respective modules. In the current API, they are in the `tensorly.xxx_tensor` modules. We also have classes for the decompositions, which can then directly use these functions (e.g. we implement `kruskal_tensor.kruskal_norm`, and this is used by `KruskalTensor.norm`).

## Bigger picture

I think both previous points come down to the following choices: 

1. Do we want to 
i. handle all the decompositions with a single (unified) interface (e.g. Tensor ) or
ii. explicitly handle each decomposition as a separate, distinct entity (e.g. KruskalTensor)?

So far, I added classes for the decompositions as very thin wrapper that behave like tuples that can actually be ignored by the users, as a test: users can do these two things equivalently:
```
ktensor = parafac(tensor, rank)
weights, factors = parafac(tensor, rank)

tl.kruskal_tensor.kruskal_normalise(ktensor)
tl.kruskal_tensor.kruskal_normalise((weights, factors))
```

I think at some point we may want to decide more explicitly which should be the way. What do you think?

#  Back to the class API

Based on the above discussions, I'd say we've more or less reached a consensus. What this would look like:

```python
class BaseDecomposition:
    def  __init__(self, *args, **kwargs):
        # store all hyper-parameters
        # Only sets the values of the attributes, nothing more

    def fit(tensor, y=None, kruskal_init=None):
        # fit the decomposition on an input tensor
        return self

    def fit_transform(tensor, y=None, kruskal_init=None):
         # ...
         return decomposed_tensor

    def transform(self, tensor):
        # easy for a Tucker decomposition (return the core obtained by projecting with the factors)
        return projected_tensor

class CPDecomposition(BaseDecomposition):
    def __init__(self, rank, init='random', ...):
        self.rank = rank
        self.init = init
        ...

    def fit_transform(self, X, y=None, decomposition_init):
         self.decomposition_ = parafac(X, rank=self.rank, init=self.init, ...)


class Tucker(BaseDecomposition):
    def __init__(self, rank=None, n_iter_max=100, 
                 init='svd', svd='numpy_svd', tol=10e-5, 
                 random_state=None, mask=None, verbose=False):
        # register the params

    def fit_transform(self, tensor):
        tucker_tensor = tucker(tensor, rank=self.rank, n_iter_max=self.n_iter_max, init=self.init, svd=self.svd,  tol=self.tol,
                                              random_state=self.random_state, mask=self.mask, verbose=self.verbose)
        self.decomposition_ = tucker_tensor
        return tucker_tensor[0] # Return the core

    def transform(self, tensor):
        _, factors = self.tucker_tensor_
        # The above is annoying to write, should we 
        # i) store explicitly core and factors or
        # ii) explicitly use the class API, so decomposition_ is a TuckerTensor and we can do decomposition.core?
        return tlg.multi_mode_dot(tensor, factors, transpose=True)

    def inverse_transform(self, tensor):
        _, factors = self.decomposition_
        return tensorly.tenalg.multi_mode_dot(tensor, factors)
```

Based on the above, only one question remains: if `init='custom'`, do we pass `kruskal_init` in the `fit` or the `fit_transform`? In scikit-learn, it mostly is passed in the `fit`, though I found one exception for the [NMF](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.NMF.html). I would tend to pass it in the fit_transform as it is dependent on the data in general.

Is everyone happy with the overall API? Any comments?

MarieRoald:
This API looks good! Though, I think @aarmey's point on separating the implementation and public API is valid. Maybe we could have a `CP` class that calls, for example, a `CPALS` class behind the scenes based on a solver argument for the `__init__` method (e.g. `solver="ALS"`, `solver="PowerIterations"`). This would be similar to how NMF takes a `solver` argument, and KMeans takes an `algorithm` argument. 

I also like @aarmey's idea of always returning a decomposition object (whether that is a tuple or a separate decomposition class). Then, we could have an estimator in the pipeline that selects the interesting part of the decomposition object.

Regarding the final question, in my mind, it makes sense to pass the initial decomposition to either the fit method or the `__init__` method. However, I mostly found it in the `__init__` method for scikit-learn classes (at least it is that way for KMeans, FastICA, SparsePCA and DictionaryLearning), so I think we should go with that (though, I am curious why it is different for the NMF class).

yngvem:
Regarding the bigger picture, I like the way you have used these minimal classes, @JeanKossaifi. 

I think that we can handle the decompositions in such minimal classes and possibly even just store them as tuples in our estimator classes (i.e. the classes whose `fit` function we call). Then we can have a kruskaltensor folder or kruskaltensor.py file that contains all the Kruskal tensor specific helper functions.

However, it is important that even if the different decomposition classes don't inherit from a common Tensor class, that they still implement the same API.

JeanKossaifi:
I accidentally merged this while pushing another commit. I reverted the main branch but it seems Github does not allow  'unmerging' / reopening a merged PR so I will reopen a PR.

JeanKossaifi:
Side note: I accidentally merged this while pushing another commit. I reverted the main branch but it seems Github does not allow 'unmerging' / reopening a merged PR so I will reopen a PR.

PR #197, opened 2020-10-02
Decomposition

JeanKossaifi:
This is a replacement of #191 and #195  (third time's the charm!) -- please see the detailed discussion in #191.

Background
=========

This PR is to have a concrete draft to implement the long discussed new class API for tensor decomposition. This is already the case for [tensor regression](http://tensorly.org/stable/modules/generated/tensorly.regression.tucker_regression.TuckerRegressor.html#tensorly.regression.tucker_regression.TuckerRegressor).
The reference in the matter is probably [Scikit-Learn](https://scikit-learn.org/stable/modules/classes.html) which did an amazing job in designing a clear, simple and effective API. 

There are a few concepts behind it, namely:
	•	Consistency (of an interface with a small limited set of methods)
	•	Inspection
	•	Non-proliferation of classes
	•	Composition
	•	Sensible defaults
You can read more about in the [related paper](https://dtai.cs.kuleuven.be/events/lml2013/papers/lml2013_api_sklearn.pdf). These are all great concepts that we try to enforce in TensorLy.

API Summary
==========

## Fast computation and helper functions

Each tensor decomposition's module (e.g. `tensorly.kruskal_tensor`, etc) should include methods to efficiently manipulate decomposed tensors, in a unified API shared between decompositions

Helper functions for particular decompositions will also be there.

## Class API

The consensus is a scikit-learn compatible API, which we can build on to add Pipeline, etc. 

What this would look like:

```python
class BaseDecomposition:
    def  __init__(self, *args, **kwargs):
        # store all hyper-parameters
        # Only sets the values of the attributes, nothing more

    def fit(tensor, y=None, kruskal_init=None):
        # fit the decomposition on an input tensor
        return self

    def fit_transform(tensor, y=None, kruskal_init=None):
         # ...
         return decomposed_tensor

    def transform(self, tensor):
        # easy for a Tucker decomposition (return the core obtained by projecting with the factors)
        return projected_tensor

class CPALS(BaseDecomposition):
    def __init__(self, rank, init='random', ...):
        self.rank = rank
        self.init = init
        ...

    def fit_transform(self, X, y=None, decomposition_init):
         self.decomposition_ = parafac(X, rank=self.rank, init=self.init, ...)

class CPPowerIteration(BaseDecomposition):
    ...

class TuckerALS(BaseDecomposition):
    def __init__(self, rank=None, n_iter_max=100, 
                 init='svd', svd='numpy_svd', tol=10e-5, 
                 random_state=None, mask=None, verbose=False):
        # register the params

    def fit_transform(self, tensor):
        tucker_tensor = tucker(tensor, rank=self.rank, n_iter_max=self.n_iter_max, init=self.init, svd=self.svd,  tol=self.tol,
                                              random_state=self.random_state, mask=self.mask, verbose=self.verbose)
        self.decomposition_ = tucker_tensor
        return tucker_tensor[0] # Return the core

    def transform(self, tensor):
        _, factors = self.tucker_tensor_
        # The above is annoying to write, should we 
        # i) store explicitly core and factors or
        # ii) explicitly use the class API, so decomposition_ is a TuckerTensor and we can do decomposition.core?
        return tlg.multi_mode_dot(tensor, factors, transpose=True)

    def inverse_transform(self, tensor):
        _, factors = self.decomposition_
        return tensorly.tenalg.multi_mode_dot(tensor, factors)
```

Principles to consider 
================
Key points discussed in the PR:

* Separation between data and algorithm:  the classes should not store the training data, only their own parameters. In particular, `KruskalTensor` is a separate class from the `Parafac` one. One issue with mixing the two is that it would introduces side effects makes is much harder to compose function (e.g. to create Pipelines).

* Optimization methods: we will have a separate class for each major optimization methods (e.g. `ParafacALS`). We may consider having a base class `Parafac` taking as parameter e.g. `method='als'`, which will specify which optimization method to use. This is slightly tricky as different methods may have different arguments and having a base class might be optimal for discoverability of these.

* Discoverability: Ideally, we want the methods to clear have easily discoverable (through auto-complete, docstrings, etc) parameters.


JeanKossaifi:
Thanks so much all @cohenjer @aarmey @MarieRoald @yngvem for the great discussions and for being part of TensorLy!
 I updated the PR based on our discussion:

* We now have scikit-learn compatible classes for all tensor decompositions.
* I have adopted the philosophy we discussed -> all tensor decomposition should return the decomposed tensor (internally stored as `decomposition_`. 
* I added a Mixin classes inherited by all tensor decomposition classes. This can be expanded in the future. 
* I also added a shared base class for all Factorized tensors.
* Based on suggestion from @cohenjer  and others, I will also deprecate kruskal_tensor and similar in favour of cp_tensor to be consistent (working on this right now).

------------------------

There is one addition that I am not sure about: classes for tensors in factorized form.  I added these for all decompositions, and they behave like tuple. They also implement efficient operations on decomposed tensors as we discussed.
I am not sure about having these efficient operations as part of the class though as we cannot ensure a consistent API. 

Imagine the following scenario:
i) create a tensor (decomposed or not)
ii) use it with the nice API (`tensor.mode_dot(matrix)`, `tensor.norm()`, etc).
iii) do an operation that modifies the tensor (e.g. `tensor = torch.relu(tensor)` or even just `tensor = tensor.to_unfolded()`)
iv) try `tensor.fold()` -> FAILS as the result is no longer a TensorLy tensor. 
This may also happen if a use gets a NumPy (or Pytorch) tensor and expects it to have these methods as well.

Should we 1) accept this shortcoming and document this behaviour or 2) encourage instead using functions *on* those tensors, e.g. `cp_mode_dot(tensor, matrix, mode)`, `cp_norm(tensor)`, etc? 
I would say I prefer 1 as a user but 2 as a developer catering to many different kind of users. 

--------------------------

Is everyone happy with the other changes and what do you think regarding the factorized tensor classes?

aarmey:
Regarding option #1 vs. #2, this seems to be varied in both NumPy and Pandas, so I'm guessing there's not a clear answer. I can see the argument for #2, but also do like to be able to type `tensor.` and then get all the autocomplete options. Perhaps the answer is to provide both? For example, there's `np.copy()` and `ndarray.copy()`, `np.sum()` and `ndarray.sum()`, etc. One can just call the other form.

yngvem:
Based on @JeanKossaifi's scenario, I believe that the best solution to support both options, like @aarmey proposed. Implementation-wise, I think it will be simple.

```python
def norm(cp_tensor):
    ...
    return norm

...

class CPTensor:
    ...
    def norm(self):
        return norm(self)
```

However, if we have to choose on, then maybe it's be best to have the functionality as function rather than methods of the decomposition classes. I prefer consistency over convenience. Also, having functions that can take tuples as input makes it possible to support the TensorLy functionality without requiring TensorLy as a dependency.

JeanKossaifi:
Thanks @aarmey and @yngvem -- I agree in principle, my worry is inconsistent behaviour: in NumPy you can use either `numpy.method`, `array.method()` or both but the *output* is consistent and *always* an `ndarray`. 

However, when using "our" objects, users will get nice methods, e.g. `unfolding = tensor.unfold(mode=1)` but then, `unfolding.fold()` will not exist. Unless we also create classes for regular tensors, unfolded tensors and vectorized tensors (which we can do). But then, calling a regular function from the underlying backend will still return a tensor without those methods (e.g. `res = torch.matmul(unfolding, matrix)` will return a new tensor for which the call `res.fold()` will fail. 

The only way to ensure consistency would be to subclass each and every backend base tensor class so that all operations on Tensors (whether with a tensorly method or backend (torch, numpy) method) return a tensor augmented with our methods (even then there may be edge cases potentially breaking that behaviour). 

---------------------------

This consideration aside, is there any blocking issue with this PR? If not, I will go ahead and merge after refactoring kruskal_tensor to cp_tensor (which will changes a lot of files and make the PR hard to review).

JeanKossaifi:
Everyone happy with merging this? 
@aarmey @cohenjer @yngvem @MarieRoald

cohenjer:
Hi all,
The PR looks great. I think there will be room for modifications further down the line, but it's a perfect starting point as far as I am concerned. A few very minor comments:

- Since my goal is to add a few algorithms for CP and Tucker in the coming months, maybe we should discuss a bit further the naming convention. I see there is a randomisedCP class and a CP class. CP class is fine (it's like the default class, implementing ALS at the moment), but 
  - we could split CP into two classes, CP and NNCP. As we discussed, nonnegative CP typically requires different methods than CP, and is arguably a pretty different model than CP when implementation is considered. I also want to add soon a more general constrained CP class, something like CCP? 
  - randomisedCP may be renamed as CPRAND? it's also the official name of the algorithm by the way. This way other algorithms (e.g. CPGN (gauss newton), CPADMM, ...) have consistent names. Also chaining my previous remark, for NNCP we may as well have algorithms such as NNCPHALS, NNCPADMM, ...
  - seing how capital letters are piling up, maybe we can change a little bit the names, using small letters for algorithm, like CPrand, CPadmm, CPgn, CPals, NNCPhals and so on?

- Then regarding the matter of class methods vs functions, supposing we want to chose, I also prefer to have function defined outside each class which support tensors in specific formats.

JeanKossaifi:
Awesome, thank you all, merging! @aarmey @yngvem @MarieRoald @cohenjer 

Looking forward to having new algorithms, pipelines, plus many more new features and refining the API further!

PR #202, opened 2020-11-22
ALS with missingness patterns

aarmey:
This removes chords from the unfolded tensor for cases when there are patterns to missing data. There is no lstsq backend operation, and so I am wondering whether this should be added, or I should transform this to work with the existing `tl.solve`.

### DIFF
--- a/tensorly/decomposition/_cp.py
+++ b/tensorly/decomposition/_cp.py
@@ -7,7 +7,7 @@ from ..random import check_random_state, random_cp
 from ..base import unfold
 from ..cp_tensor import (cp_to_tensor, CPTensor,
                          unfolding_dot_khatri_rao, cp_norm,
-                         cp_normalize, validate_cp_rank)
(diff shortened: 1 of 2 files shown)


cohenjer:
After you remove rows from the unfolded tensor, the problem is actually the same as in the CPRAND algorithm (the randomized_parafac in tensorly). Maybe you could try to follow the same syntax (or update the syntax in that function if yours is better)?

You can always call the solve backend routine by writing the normal equations (y = Ax --> Aty = AtAx). This is typically crucial in ALS because the inner product of khatri-rao matrices is efficiently computed. Nevertheless with missing rows in the unfolding, you can still compute this normal equations, it simply won't be ``better'' than calling lstsq because you lost the khatri rao structure essentially. 
Since it's the same in the end, I think we can avoid adding lstsq to the supported backend operations.

aarmey:
Note to self: This needs to use banded missingness thresholds, not all the indices of non-missing values.

Issue #203, opened 2020-11-23
Improved non-negative SVD


aarmey:
#### Is your feature request related to a problem? Please describe.

Currently it looks like we simply take the standard SVD and then the absolute value of every element, for non-negative CP and Tucker.

#### Describe the solution you'd like

There are [implementations](https://github.com/mims-harvard/nimfa/blob/a37335f249ba74e86b08fef06a63460b02230bb2/nimfa/methods/seeding/nndsvd.py) of non-negative SVD which show [better convergence with non-negative matrix factorization](https://www.sciencedirect.com/science/article/abs/pii/S0031320307004359). I wonder if one of these would be beneficial to implement for non-negative CP and Tucker.

#### Additional context

These methods are also available in sci-kit learn.

cohenjer:
I think that would be great :)

JeanKossaifi:
I agree that would be great! In addition to better SVD init, I also wanted to add better init by averaging columns/rows/fibers of the input.

aarmey:
@cohenjer would you mind if I see that code? Figure it'd help me get it right the first time.

cohenjer:
@aarmey In fact we used the implementation for the nimfa toolbox:
https://github.com/mims-harvard/nimfa/blob/master/nimfa/methods/seeding/nndsvd.py

Not sure we want nimfa as a dependancy however. For an exemple usage,
https://gitlab.inria.fr/amarmore/nonnegative-factorization/-/blob/master/nn_fac/nmf.py

aarmey:
Ok, thanks! This looks similar to scikit-learn's implementation. I don't think we need new dependencies. I have been trying to decide whether this fits more easily into the backend or frontend—it looks like everything will work with the backend operations we have

aarmey:
#204

PR #204, opened 2020-12-10
Refactor non-negative CP and provide nndsvd initialization

aarmey:
- tl.clip seems to be broken for the MXnet backend when only `a_min` or `a_max` are used alone.
- Perhaps it's worth re-factoring the non-negative methods into a separate file? They seem to be almost entirely separate.


JeanKossaifi:
Thanks @aarmey  -- will check mxnet's clip. 

> Perhaps it's worth re-factoring the non-negative methods into a separate file? They seem to be almost entirely separate.

I agree -- let's separate them!

aarmey:
Great. I'll make those changes here.

tensorly/decomposition/_nn_cp.py
```diff
@@ -0,0 +1,413 @@
...
+
+            factors.append(U[:, :rank])
+
+        kt = CPTensor((None, factors))
+
+    elif isinstance(init, (tuple, list, CPTensor)):
+        # TODO: Test this
+        try:
+            kt = CPTensor(init)
+        except ValueError:
```
aarmey, line 125: Should this exception be within CPTensor?

tensorly/decomposition/_cp.py
```diff
@@ -44,9 +43,7 @@ def initialize_cp(tensor, rank, init='svd', svd='numpy_svd', random_state=None,
     rng = check_random_state(random_state)
 
     if init == 'random':
-        # factors = [tl.tensor(rng.random_sample((tensor.shape[i], rank)), **tl.context(tensor)) for i in range(tl.ndim(tensor))]
-        # kt = CPTensor((None, factors))
-        return random_cp(tl.shape(tensor), rank, normalise_factors=False, **tl.context(tensor))
+        kt = random_cp(tl.shape(tensor), rank, normalise_factors=False, **tl.context(tensor))
```
aarmey, line 46: This can't immediately return because it potentially has to be normalized later.

aarmey:
Addresses #203.

tensorly/decomposition/_nn_cp.py
```diff
@@ -0,0 +1,413 @@
...
+    if tl.ndim(tensor)-1 in fixed_modes:
+        warnings.warn('You asked for fixing the last mode, which is not supported while tol is fixed.\n The last mode will not be fixed. Consider using tl.moveaxis()')
+        fixed_modes.remove(tl.ndim(tensor)-1)
+    modes_list = [mode for mode in range(tl.ndim(tensor)) if mode not in fixed_modes]
+
+    for iteration in range(n_iter_max):
+        if orthogonalise and iteration <= orthogonalise:
+            for i, f in enumerate(factors):
+                if min(tl.shape(f)) >= rank:
+                    factors[i] = tl.abs(tl.qr(f)[0])
```
aarmey, line 210: This seems like a really bad approach to orthogonalization. An alternative might be [pivoting](http://proceedings.mlr.press/v97/zhang19r/zhang19r.pdf)? I'd suggest removing it for now...

aarmey:
This should be ready for review @JeanKossaifi. The only remaining test failure is from tensorflow, which doesn't seem to accept boolean indexing? Could use your advice on how to resolve that.

tensorly/decomposition/_nn_cp.py
JeanKossaifi: Possibly but in this case we also want to alert the user about what he did wrong (e.g. they should get an error message explaining what is wrong -- not a CP init-- and hinting at how to fix it. If the error is raised in CPTensor it might be harder for them to understand and fix the issue?

tensorly/decomposition/_nn_cp.py
```diff
@@ -0,0 +1,413 @@
...
+
+                lbd = tl.sqrt(S[j] * sigma)
+                W = tl.index_update(W, tl.index[:, j], lbd * u)
+                H = tl.index_update(H, tl.index[j, :], lbd * v)
+
+            # After this point we no longer need H
+            idx = W < np.finfo(np.float32).eps
+
+            if nntype == "nndsvd":
+                W = tl.index_update(W, tl.index[idx], 0.0)
```
JeanKossaifi, line 101: For this we can use proximal.soft_thresholding with esp as threshold.

tensorly/decomposition/_nn_cp.py
```diff
@@ -0,0 +1,413 @@
...
+            elif nntype == "nndsvda":
+                avg = tl.mean(tensor)
+                W = tl.index_update(W, tl.index[idx], avg)
```
JeanKossaifi, line 104: A less-efficient way to do this is using something like this
```python
tl.where(W < eps, tl.ones(tl.shape(scales), **tl.context(factor))*avg, W)
```
But this should work for all backends.

tensorly/decomposition/_nn_cp.py
```diff
@@ -0,0 +1,413 @@
...
+                    u = x_n / x_n_nrm
+                    v = y_n / y_n_nrm
+                    sigma = m_n
+
+                lbd = tl.sqrt(S[j] * sigma)
+                W = tl.index_update(W, tl.index[:, j], lbd * u)
+                H = tl.index_update(H, tl.index[j, :], lbd * v)
+
+            # After this point we no longer need H
+            idx = W < np.finfo(np.float32).eps
```
JeanKossaifi, line 98: This could be just `eps = tl.eps(tensor.dtype)` and then using it as in the comments below.

JeanKossaifi:
Thanks for the PR @aarmey !

I'm wondering whether it would be worth having a separate function for the non-negative SVD and calling it in the CP-init as it might be useful in other places (e.g. non-negative Tucker).

tensorly/decomposition/_nn_cp.py
JeanKossaifi: This orthogonalization was taken from the paper "[Orthgonalized ALS for Tensor Decomposition](http://proceedings.mlr.press/v70/sharan17a/sharan17a.pdf)", Sharan and Valiant, ICML 2017.

It was added to the original CP and was never intended to really be used with the non-negative version and might be problematic indeed. Pivoting would be great to have instead! Happy to remove in the meantime.

aarmey:
Thanks @JeanKossaifi. I agree about separating nndsvd into a function so that is done here. I've applied your other suggestions, and it looks like everything is building (aside from that flakey test).

JeanKossaifi:
Thanks @aarmey -- will review asap. It seems there is an issue with Travis that was going through my personal credit for testing. As I ran out, testing has currently been disabled, I'll try to fix the issue or switch to Github's workflows instead.

JeanKossaifi:
I switched to Github actions for the tests -- any commit should trigger it.

Philosophical question: should `nnsvd` take an SVD as input and make it non-negative as currently done, or should it directly take a tensor and apply both steps (we can then have a intermediate function, e.g. `make_svd_non_negative`). What do you think?

aarmey:
I like having *a* function that just takes the SVD as input because it doesn't depend in any way on the manner by which the SVD was constructed. A function that offers `make_svd_non_negative` seems natural, but we don't have a `make_svd` on the other hand. Would a `make_svd`/`make_svd_non_negative` help to simplify `initialize_nn_cp`, `initialize_cp`...?

JeanKossaifi:
> I like having _a_ function that just takes the SVD as input because it doesn't depend in any way on the manner by which the SVD was constructed. 

I agree. 

The non-negative init is a two step process i) apply regular SVD: `U, S, V = svd(...)` and ii) (only in the non-negative case) `make_svd_non_negative(U, S, V)` -- I suggested this name to make explicit that it takes an _existing_ SVD and *makes* it non-negative (and then we could optionally have a function `nnsvd` that directly takes as input a tensor and applies both steps). The reasoning was that a user might expect any kind of SVD to take a matrix as input, but I am also happy to just keep the current naming! What do you think?

aarmey:
Oh! Completely agree. The function is renamed here now.

I believe I've fixed `test_parafac` now. Previously it was making a tensor filled with random numbers of undetermined rank, then decomposing with rank 6. I changed this to make sure it's using a tensor that actually has corresponding low-rank structure. I also used `pytest.parameterize` a bit to clean up the various combination of tested options.

JeanKossaifi:
Awesome, thanks a lot @aarmey! 
You may want to add yourself to the author list for the nn-svd? Otherwise happy to merge! :) 

I thought I had fix that pesky tucker validation issue -- will dig into it.

aarmey:
Done. Thanks @JeanKossaifi

PR #205, opened 2020-12-13
Feature/mpca

james-oldfield:
Hey folks!

I thought an implementation of Haiping Lu's Multilinear PCA could go nicely into `tensorly.decomposition`--what do you think? (I can also add a notebook demo using it for identity classification if that suits. Where is a good place to put this?)

I've added a unit test to confirm we can reconstruct the original tensors from the inverse projection matrices (inspired by the robustPCA's unit tests), and one to confirm the numerical equivalence of the eigendecomp+SVD for the total scatter matrices (with `tl.partial_svd` preferred for its being backend-agnostic?). Unfortunately, I can't run cupy on my machine... but the tests for all the other backends pass!

Any suggestions? or more ideas for unit tests?

Looking forward to hearing your feedback :-)!

### DIFF
--- a/doc/modules/api.rst
+++ b/doc/modules/api.rst
@@ -241,6 +241,7 @@ Functions
     partial_tucker
     non_negative_tucker
     robust_pca
+    mpca
     tensor_train
     tensor_train_matrix
     parafac2
--- a/tensorly/decomposition/__init__.py
+++ b/tensorly/decomposition/__init__.py
@@ -7,6 +7,7 @@ from ._cp import (parafac, non_negative_parafac, CP, RandomizedCP,
                                 randomised_parafac, sample_khatri_rao)
 from ._tucker import tucker, partial_tucker, non_negative_tucker, Tucker
 from .robust_decomposition import robust_pca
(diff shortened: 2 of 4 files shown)


JeanKossaifi:
Thanks @james-oldfield !

Can't we directly use partial_svd on the unfoldings without explicitly computing the mode-k total scatter, since we're using svd anyway? Is there a fundamental difference with the Tucker decomposition, aside from not explicitly computing the core?

I can see a potential use-case to having a separate `TensorPCA` class for users not familiar with tensor methods and just looking for multi-linear version of PCA rather than having to build it from Tucker. In any case the version with the scatter is great to have as a tutorial!

james-oldfield:
Agreed--this is probably more suited left as a tutorial than as a method in the main library (especially now I think that the tucker decomp already lives here!). Closing this PR for now, and we can pick up with implementation details at a later date if it ever fits :-).

JeanKossaifi:
Hi @james-oldfield -- I meant having this as a TensorPCA, separate from Tucker can be helpful for new users that want to just do dimensionality reduction. At the very least we can turn this into a notebook style example for the gallery?

PR #206, opened 2020-12-14
Feature/mda

james-oldfield:
Also opening a PR for Li's constrained multilinear discriminant analysis [1]. This could also perhaps suit `tensorly/decomposition` best, given the current structure? Along with MPCA, there's also a notebook walkthrough that could potentially fit into the examples!

---

- [1] Q. Li and D. Schonfeld, "Multilinear Discriminant Analysis for Higher-Order Tensor Data Classification," in IEEE Transactions on Pattern Analysis and Machine Intelligence, vol. 36, no. 12, pp. 2524-2537, 1 Dec. 2014, doi: 10.1109/TPAMI.2014.2342214.

### DIFF
--- a/tensorly/decomposition/mda.py
+++ b/tensorly/decomposition/mda.py
@@ -1,0 +1,238 @@
+import numpy as np
+from .. import backend as T
+from ..base import unfold
+from ..tenalg import multi_mode_dot
+
+
+# Author: James Oldfield
+
+# License: BSD 3 clause
+
+
+def mda(X, y, ranks, n_iters=5):
+    """Multilinear Linear Discriminant Analysis (MDA).
+
+        Learns a projection matrix for each mode of the data tensor
+        to project the input tensor into a low-dimensional tensor subspace
+        where the *scatter ratio criterion* is maximised.
+
+    Parameters
+    ----------
+    X : ndarray
+        tensor data of shape (n_samples, N1, ..., NS). Note: the first dimension is the sample dimension.
+    y : ndarray
+        list of length (n_samples) containing the integer-valued class label of tensor sample i.
+    ranks : list
+        list of integers determining the dimension of the subspace for mode-k.
+    n_iters : int, optional, default is 5
+        number of steps to employ the APP scheme for.
+
+    Returns
+    -------
+    factors : list
+        list of the learnt projection matrices for each mode
+
+    Notes
+    -----
+
+    This implementation computes the *Constrained Multilinear Discriminant Analysis* (CMDA) solution as presented in [1].
+
+    Given the learnt factor matrices, one can then compute the projection along all modes with::
+
+      factors = mda(X_train, y_train, ranks, n)
+      Z = tl.tenalg.multi_mode_dot(X_train, factors, modes=[1, 2], transpose=True)
+
+
+    The so-called **scatter ratio objective function** (a higher-order extension of the Fisher criterion) for each mode n is maximised. Concretely, both the mode-n between- and within-class scatter matrices are computed. First a global mean tensor :math:`\\mathcal{M}` (the higher-order analogue of the mean vector :math:`\\mathbf{m}` in the standard LDA setting), and a set of class-specific mean tensors :math:`\\mathcal{M}_i` for :math:`i=1,\\dots, c` are computed. The class-specific mean tensors are computed as
+
+    .. math::
+       :nowrap:
+
+        \\begin{equation*}
+        \\begin{aligned}
+            \\mathcal{M}_i= \\frac{1}{n_i} \\sum_{j=1}^{n_i} \\mathcal{X}_{ij},
+        \\end{aligned}
+        \\end{equation*}
+
+    and the global mean tensor is computed as
+
+    .. math::
+       :nowrap:
+
+        \\begin{equation*}
+        \\begin{aligned}
+            \\mathcal{M}= \\frac{1}{N} \\sum_{i=1}^{c} \\sum_{j=1}^{n_i} \\mathcal{X}_{ij}.
+        \\end{aligned}
+        \\end{equation*}
+
+    Following this, the mode-n between- and within-class scatter matrices are computed (following [1]) as:
+
+    .. math::
+       :nowrap:
+
+        \\begin{equation*}
+        \\begin{aligned}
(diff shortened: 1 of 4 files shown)


JeanKossaifi:
Thanks @james-oldfield!
This is a great new feature! 

On a high-level, would it make sense to also have an additional solver based on SVD (e.g. like for the [linear case](https://scikit-learn.org/stable/modules/generated/sklearn.discriminant_analysis.LinearDiscriminantAnalysis.html) in scikit-learn)? This would avoid building the covariance matrix (it seems we make the assumption that all classes have the same covariance anyway here?). However, it's not clear whether it would be computationally advantageous in the end. Looking at the update rule, an online version might be nice to add down the line.

james-oldfield:
Definitely! Although off the top of my head I'm not quite sure how one would extend the SVD solver for the multilinear case (i.e. how we'd solve for the factor matrices with the SVD directly on the batch of mode-n unfolding (a 3rd-order tensor), without computing the scatter matrices explicitly?) It certainly would make a nice optional feature though if it were possible.

That said, I guess materialising the mode-n scatter matrices is less of a worry in the multilinear case, given fewer features than if we'd vectorised the whole thing.

One thing that would be nice in the future could be to provide a flag to instead take the solution via the eigendecomposition of `W^{-1}B` (for the [DATER](https://ieeexplore.ieee.org/abstract/document/1467312/) solution--rather than the CMDA solution via the SVD).

Good idea with the docstring, I've removed a lot of the bulk :)

tensorly/decomposition/mda.py
```diff
@@ -0,0 +1,165 @@
...
+      Z = tl.tenalg.multi_mode_dot(X_train, factors, modes=[1, 2], transpose=True)
+
+    - [1] Q. Li and D. Schonfeld, "Multilinear Discriminant Analysis for Higher-Order Tensor Data Classification," in IEEE Transactions on Pattern Analysis and Machine Intelligence, vol. 36, no. 12, pp. 2524-2537, 1 Dec. 2014, doi: 10.1109/TPAMI.2014.2342214.
+
+    """
+
+    ###############
+    # check correct # of ranks have been supplied
+    ###############
+    assert len(ranks) == len(T.shape(X)[1:]), 'Expected number of ranks: {}. \
```
JeanKossaifi, line 52: It's good for end users to make the error messages as clear as possible. Something like:

```suggestion
    if len(ranks) != (T.ndim(X) - 1):
        msg = f'Got as input a tensor of shape {T.shape(X)} corresponding to {T.shape(X)[0]} samples of shape {T.shape(X)[1:]} of order {T.ndim(X) -1}.'
        msg += f'But got {len(ranks)} ranks != {T.ndim(X) -1}.'
        raise ValueError(msg)
```

tensorly/decomposition/mda.py
```diff
@@ -0,0 +1,165 @@
...
+    class_idx = [[] for _ in range(len(set(y)))]
+
+    # store the training data's class label at class index
+    for i, label in enumerate(y):
+        class_idx[label] += [i]
+
+    # store the mean of all tensor samples with label i
+    for i in range(len(set(y))):
+        # tensorflow is only backend to not support indexing into tensor with a list
+        if backend == 'tensorflow':
```
JeanKossaifi, line 70: I really don't like/want to have these in the code (it should be backend agnostic). However tensorflow really does not make it easy.. Would be ideal to find another way if it doesn't slow down the other backends.

tensorly/decomposition/mda.py
```diff
@@ -0,0 +1,165 @@
...
+            class_means += [T.mean(T.tensor([X[j] for j in class_idx[i]]), axis=0)]
+        else:
+            class_means += [T.mean(X[class_idx[i], ...], axis=0)]
```
JeanKossaifi, line 73: We can directly get the means without creating `class_idx`: 
```suggestion
            class_means += [T.mean(X[y == i, ...], axis=0)]
```

tensorly/decomposition/mda.py
```diff
@@ -0,0 +1,165 @@
...
+            # i.e. solve SX=I for X
+            W_scat_inv = T.solve(W_scat, T.eye(T.shape(W_scat)[0]))
+
+            ###################################################
+            # *Constrained Multilinear Discriminant Analysis* (CMDA) [1] solution for factor matrix U_k is given by
+            # top `rank' number of left-singular vectors of W^{_1}B.
+            # --
+            # [1] Q. Li et al. "Multilinear Discriminant Analysis for Higher-Order Tensor Data Classification"
+            ###################################################
+            U, _, _ = T.partial_svd(T.dot(W_scat_inv,  B_scat))
```
JeanKossaifi, line 98: We should probably be using SVD_FUNS but this is being refactored in #217 and should be easier to use when that gets merged.

tensorly/decomposition/mda.py
```diff
@@ -0,0 +1,165 @@
...
+            factors[k] = U[:, :ranks[k]]
+
+    return factors
+
+
+def compute_modek_wb_scatters(X, mode, factors, global_mean, class_means, class_idx):
```
JeanKossaifi, line 104: What about naming it `mode_scatter_matrices`? That would be consistent with `mode_dot`. 
Or even just `scatter_matrices` and we document well that this corresponds to mode-n between and within class scatter matrices?

tensorly/decomposition/tests/test_mda.py
```diff
@@ -0,0 +1,37 @@
...
+    ###########################################
+    factors = mda(X, y, ranks=[5, 5, 5], n_iters=5)
+
+    # project onto MDA tensor subspace
+    Z = multi_mode_dot(X, factors, modes=[1, 2, 3], transpose=True)
+
+    # recover X, using the inverse projection matrices
+    X_hat = multi_mode_dot(Z, [tl.tensor(inv(tl.to_numpy(f))) for f in factors], modes=[1, 2, 3], transpose=True)
+
+    assert_array_almost_equal(X, X_hat, decimal=tol)
```
JeanKossaifi, line 37: Isn't this always true as long as the factors are not singular? Also the factors are obtain through SVD so orthogonal, we can just take their transpose instead of converting to NumPy and using `inv`. It would be good to have some specific tests.

PR #207, opened 2020-12-17
Fix mxnet clip and add simple test

aarmey:
The problem turned out to be a simple typo. 🎉


aarmey:
The test failures look to be unrelated.

JeanKossaifi:
Awesome, thanks @aarmey !

Indeed, the failure was with the line-search test, it seems to often fail, we may want to relax it a little.

PR #209, opened 2020-12-21
Fix random functions to use NumPy's global seed

### LINKED ISSUE #208: Random tensors are not reproducible with global seed set
Currently setting numpy.seed, torch.manual_seed, etc., is not sufficient for reproducible results when a seed is not explicitly set for the random tensor functions. 
Inspired by scipy and scikit-learn`check_random_state` function, 
https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/utils/validation.py#L862
https://github.com/scipy/scipy/blob/624fd766e2886c458012d98e3c6ec85551ab2916/scipy/_lib/_util.py#L201
we can change the `check_random_state` function to return the global numpy RNG if the seed is `None`. This will make sure setting `np.random.seed(seed)` is enough for reproducible results.

merajhashemi:
This resolves #208.

### DIFF
--- a/tensorly/random/base.py
+++ b/tensorly/random/base.py
@@ -25,7 +25,10 @@ def check_random_state(seed):
     -----
     Inspired by the scikit-learn eponymous function
     """
-    if seed is None or isinstance(seed, int):
+    if seed is None:
+        return np.random.mtrand._rand
+
+    elif isinstance(seed, int):
         return np.random.RandomState(seed)
 
     elif isinstance(seed, np.random.RandomState):


JeanKossaifi:
Thanks @merajhashemi, this is very useful, merging.

As a side note, the issue you had previously was due to cyclic imports -- all tensorly modules import the backend.

merajhashemi:
Thank you @JeanKossaifi. I was trying to modify the `partial_svd` function, but there was no need.
I thought that when `v0` is `None`, ARPACK will use its own internal RNG and mess up reproducibility. But the seed is actually fixed to 1357 in the Fortran code!
https://github.com/scipy/scipy/blob/master/scipy/sparse/linalg/eigen/arpack/ARPACK/SRC/cgetv0.f#L200-L203

JeanKossaifi:
@merajhashemi thanks. It might be good to explicitly use the global seed there too. One way would be some refactoring to move `check_random_state` to the backend. I have been considering doing this: on the one side, this is an easy way to use native random number generators for each backend. On the flip side, the backend might become a messy blob of functions.

The issue you created was closed automatically when this was merged -- we can re-open until this is fixed too. We can at least document this in the docstring of `check_random_state` for future users/maintainers.

PR #212, opened 2020-12-23
Move check_random_state to core backend

merajhashemi:
Following our discussion with @JeanKossaifi here #209. This pr moves the `check_random_state` function from `tensorly.random` to the backend.


JeanKossaifi:
Ah I was actually debating whether and how to move this to the backend -- the risk is the backend becoming a long mess. Also I was wondering whether it is more logical to have all random related functions living in `tensorly.random`.

This would be fine though if we don't add too many such fun directly in the backend -- perhaps we can then also add a function to sample random tensors, e.g.
```python
def randn(size, random_state=None, **context):
    ...
```
where context would contain at least dtype but also for pytorch device, etc.

PR #213, opened 2020-12-23
Add conj to pytorch backed

merajhashemi:
Since PyTorch 1.4.0 `torch.conj` has been added. This pr adds this function to the pytorch backend.

PR #215, opened 2020-12-23
Add randomized SVD

### LINKED ISSUE #214: Add feature: Randomized SVD
As introduced in [Halko et al (2009)](https://arxiv.org/abs/0909.4061) we can implement the randomized SVD algorithm.
Implementations in scikit-learn and PyTorch:

https://github.com/scikit-learn/scikit-learn/blob/42aff4e2e/sklearn/utils/extmath.py#L245
https://github.com/pytorch/pytorch/blob/55b431b17aba504ae7b75f6f97b4437101e50f38/torch/_lowrank.py#L85

merajhashemi:
This pr resolves #214.


merajhashemi:
@JeanKossaifi this pr requires #212 and #213 (for PyTorch only) to merge. I can change it to be merged before those.

tensorly/backend/core.py
```diff
@@ -833,8 +833,110 @@ def truncated_svd(self, matrix, n_eigenvecs=None, **kwargs):
...
+
+        Notes
+        -----
+        This function is implemented based on Algorith 5.1 in `Finding structure with randomness:
+        Probabilistic algorithms for constructing approximate matrix decompositions`
+        - Halko et al (2009)
+        """
+
+        # hack to avoid the warning in partial svd
+        def _partial_svd(matrix, n_eigenvecs=n_eigenvecs, random_state=None):
```
JeanKossaifi, line 873: Why did you use this? This will also only work for the NumPy backend if we use scipy.linalg.svd.

JeanKossaifi:
Thanks @merajhashemi, you're on a roll! :D 
Perhaps we can finalize  #217  and use it here once it's merged -- currently it is limited to the NumPy backend as you used scipy.

merajhashemi:
I'll add tests for `randomized_range_finder` and then I think it's ready.

tensorly/backend/core.py
```diff
@@ -941,45 +941,144 @@ def symeig_svd(self, matrix, n_eigenvecs=None, **kwargs):
...
+            warnings.warn('Trying to compute SVD with n_eigenvecs={0}, which '
+                          'is larger than max(matrix.shape)={1}. Setting '
+                          'n_eigenvecs to {1}'.format(n_eigenvecs, max_dim))
+            n_eigenvecs = max_dim
 
-        if dim_1 < dim_2:
+        if dim_1 > dim_2:
             S, U = self.eigh(self.dot(matrix, self.transpose(matrix)))
-            S = self.sqrt(S)
+            S = self.sqrt(self.clip(S, self.eps(S.dtype)))
```
JeanKossaifi, line 957: Is that better than the other approach we have so far?
```python
 tl.where(tl.abs(S) <= tl.finfo(S.dtype).eps, 0, 1/S)[None, :]
```
Would it create any issue by replacing something essentially 0 by something that is not? 

If there's no impact I'm happy to keep this as it results in a cleaner code :)

JeanKossaifi:
Looks good to me, just missing the tests as you said. 

It's a great addition! Now we're just missing online/incremental SVD :D

JeanKossaifi (commented):
Looks good to me, just missing the tests as you said.

It's a great addition! Now we're just missing online/incremental SVD :D

merajhashemi:
This PR will also resolve #233 when merged.

JeanKossaifi:
Can you add a small test for `randomized_range_finder` @merajhashemi ?

tensorly/backend/core.py
merajhashemi: There were two issues:
1. Some of the negative values were smaller than `eps`, and taking the square root would result in nans. 
2. We cannot divide U by S in the next line if S has zeros.

PR #216, opened 2020-12-24
Fix svd sign indeterminacy

### LINKED ISSUE #74: Sign indeterminacy of SVD

merajhashemi:
This pr builds on @aarmey work in #174 and resolves #74.


JeanKossaifi:
Great feature that's been on the backlog for a while, thanks @merajhashemi and @aarmey!

Currently this is NumPy specific though and not all backends support __array_function__ so this won't work in all cases. Shall we move it inside the backend class and replace the `np.` by `self.`?

JeanKossaifi:
Hi @merajhashemi -- are you still working on this?

tensorly/backend/core.py
```diff
@@ -738,7 +738,51 @@ def kr(self, matrices, weights=None, mask=None):
...
+            decision on is generally algorithm dependent.
+        Returns
+        -------
+        u_adjusted, v_adjusted : arrays with the same dimensions as the input.
+        """
+        if u_based_decision:
+            # columns of u, rows of v
+            is_numpy = isinstance(u, np.ndarray)
+            max_abs_cols = self.argmax(self.abs(u), axis=0)
+            signs = np.sign(self.to_numpy(u)[max_abs_cols, range(self.shape(u)[1])])
```
merajhashemi, line 764: I haven’t worked with Tensorflow, but it seems this kind of advanced indexing isn't supported. Is it possible to do this more elegantly for all backends?

JeanKossaifi: Thanks @merajhashemi! To do advanced indexing with all backends, you can use [tl.index_update](http://tensorly.org/dev/modules/generated/tensorly.index_update.html#tensorly.index_update).

merajhashemi: `tl.index_update` is helpful for assigning values to immutable tensors (JAX and TensorFlow) but doesn't support advanced indexing. 
Example:
```python
>>> x = tl.tensor([[1, 2], [3, 4], [5, 6]])
>>> # Supports all backends except tensorflow
>>> tl.index_update(x, tl.index[[0, 1, 2], [0, 1, 0]], 100)
array([[100,   2],
       [  3, 100],
       [100,   6]])
```
To support this kind of advanced indexing in TensorFlow we need to use`tf.gather_nd` like they do [here](https://github.com/jonasrauber/eagerpy/blob/master/eagerpy/tensor/tensorflow.py#L535-L577).

merajhashemi: We can add a `get_index` method to the backend:?

JeanKossaifi: Hmm I guess we could but it really adds another cumbersome methods just for tensorflow..
For this particular case, we could also just use:
```python
signs = tl.sign(tl.tensor([M[i, j] for (i, j) in zip(max_abs_cols, range(rank))], **tl.context(M)))
```

tensorly/backend/core.py
```diff
@@ -792,7 +792,51 @@ def kr(self, matrices, weights=None, mask=None):
...
+                signs = self.tensor(signs, **self.context(u))
+            u = u * signs
+            if self.shape(v)[0] > self.shape(u)[1]:
+                signs = self.concatenate((signs, self.ones(self.shape(v)[0] - self.shape(u)[1])))
+            v = v * signs[:self.shape(v)[0]][:, None]
+        else:
+            # rows of v, columns of u
+            is_numpy = isinstance(v, np.ndarray)
+            max_abs_rows = self.argmax(self.abs(v), axis=1)
+            signs = np.sign(self.to_numpy(v)[range(self.shape(v)[0]), max_abs_rows])
```
JeanKossaifi, line 829: Same as above, generally it's better not to convert in NumPy if possible.

JeanKossaifi (changes requested):
I left a comment in the code for the NumPy conversion, otherwise, looks good to me, thanks @merajhashemi!

JeanKossaifi:
Thanks @merajhashemi - it would be good to have a small unit-test.

Wondering whether it would make sense to unify with `cp_sign_flip` @cohenjer @aarmey

aarmey:
I see how they address similar needs, but I don't see a way to unify this with `cp_sign_flip`, unless I'm missing something.

cohenjer:
Unless you want to have a weird, general sign-flip function, I would also keep these separated? But maybe I am simply not understanding what you have in mind @JeanKossaifi

JeanKossaifi:
Was just a thought as SVD can be seen as a 2D CP with weights = diag(S) and factors = [U, V], so it would be possible to resolve the sign ambiguity for all cases. Happy to leave the two separate.

JeanKossaifi (dismissed):
Looks good to me, just missing a small unit-test and I think it's ready to merge.

aarmey:
If this is just missing a test, I'm happy to write one as part of #227, since that will likely require some refactoring of the tests as well.

JeanKossaifi:
Awesome, thanks @aarmey, in that case I'll go ahead and merge!

PR #217, opened 2020-12-24
Refactoring across backends

aarmey:
- Various SVD forms are backend-agnostic.
- Norm method is backend-agnostic.
- CuPy tests now pass.

I know there's some discussion below about changing the SVD interface now that it can be handled as a frontend feature. Perhaps we can merge this as a first step?


tensorly/backend/mxnet_backend.py
```diff
@@ -57,12 +57,20 @@ def reshape(tensor, shape):
         return nd.reshape(tensor, shape)
 
     def solve(self, matrix1, matrix2):
+        # TODO: Replace this as soon as mxnet.np is stable.
```
aarmey, line 60: I don't use MXNet so I'm less familiar with its state, but `MXNet.np` looks promising. If it's feasible to switch over from the `ndarray` interface it would also remove most of the backend-specific code.

tensorly/backend/core.py
```diff
@@ -48,10 +48,6 @@ def float64(self):
     def float32(self):
         raise NotImplementedError
 
-    @property
-    def SVD_FUNS(self):
```
merajhashemi, line 52: SVD_FUNS is used in the tests.

tensorly/backend/core.py
```diff
@@ -644,56 +640,20 @@ def kron(self, a, b):
...
-                raise ValueError('All matrices should have the same number of columns.')
-
-            a = self.reshape(res, (s1, 1, s2))
-            b = self.reshape(e, (1, s3, s4))
-            res = self.reshape(a * b, (-1, n_col))
-
-        m = self.reshape(mask, (-1, 1)) if mask is not None else 1
-
-        return res*m
+        m = mask.reshape((-1, 1)) if mask is not None else 1
```
merajhashemi, line 655: tl.kr wasn’t NumPy specific:?, rewriting it with einsum will make it slower in some backends.

tensorly/backend/core.py
aarmey: `SVD_FUNS` is still included in every backend. The change here is that core provides an implementation of it, which is then optionally overwritten in backends that provide a third SVD method. I thought this might be cleaner given that the majority of backends provide the same set?

tensorly/backend/core.py
aarmey: My thinking was the `mxnet.np` interface will provide SVD and einsum methods, and so `kr` might not need to be backend-defined at all after that change. We've had a backend-specific bug before with the masking feature and so, absent other priorities, I figure it's better to have everything as unified as possible.

You're right that the einsum performance is likely very different for each backend. I'll revert the `kr` changes. It might be interesting to benchmark the two implemenetations if/when we can change `mxnet` over to the new interface.

aarmey: Thanks for the input!

tensorly/backend/pytorch_backend.py
```diff
@@ -201,43 +201,9 @@ def _reverse(tensor, axis=0):
...
-
-        if n_eigenvecs is None or n_eigenvecs > min_dim:
-            full_matrices = True
-        else:
-            full_matrices = False
-
-        U, S, V = torch.svd(matrix, some=full_matrices)
-        U, S, V = U[:, :n_eigenvecs], S[:n_eigenvecs], V[:n_eigenvecs, :]
-        return U, S, V
+    def svd(matrix, full_matrices=True):
```
aarmey, line 204: This might have to switch back to numpy, as torch's result seems to be lower accuracy, even with float64.

JeanKossaifi:
Thanks @aarmey! I like the idea of the refactoring - SVD_FUNS was a hacky solution to a problem that appeared when introducing the backend system. So, how about adding a parameter (e.g. `svd_fun`) to `truncated_svd` and making that common to all implementations, which will then just call the various svd funs to actually perform the computation? We can pass an additional parameter to `svd_fun`, n_eigenvecs for instance, which can be ignored when not used (scipy's sparse eigendecomposition uses it efficiently while standard svd does not in general). 

Regarding mxnet.np, indeed, I was waiting for it to be finalized to switched, we can open a separate PR for that!

Regarding `kr`, this implementation was much faster than other ones when I wrote it. Since we now have various tenalg backends, one `core` one using "regular" methods (e.g. reshapings + broadcasting and matmuls) and an `einsum` one that dispatches all operations to `einsum`, so we shouldn't force one or the other directly. Instead, all functions where we have a choice, such as `khatri-rao`, can be put in `tensorly.tenalg` (both `einsum_backend` and `core_backend`).

As a side note, I would like to enable dispatching tenalg calls directly to the backend in case they implement a particular function but I'm worried it can become a big mess, it's already not the most intuitive :/

tensorly/backend/core.py
JeanKossaifi: Indeed, whether we implement it using einsum or the current version, it will still have to be defined in the backend unless each backend implements it. That implementation was also much faster when I wrote it than other alternatives.

In any case, I also wrote an [`einsum` version](https://github.com/tensorly/tensorly/blob/master/tensorly/tenalg/einsum_tenalg/_khatri_rao.py), which is automatically used if you set the tenalg backend to "einsum" (`tensorly.tenalg.set_tenalg_backend("einsum")`).

tensorly/backend/core.py
```diff
@@ -729,82 +726,67 @@ def partial_svd(self, matrix, n_eigenvecs=None, random_state=None, **kwargs):
...
 
         if not is_numpy:
             U = self.tensor(U, **ctx)
             S = self.tensor(S, **ctx)
             V = self.tensor(V, **ctx)
         return U, S, V
 
-    def truncated_svd(self, matrix, n_eigenvecs=None, **kwargs):
-        """Computes a truncated SVD on `matrix` using pytorch's SVD
+    def truncated_svd(self, matrix, n_eigenvecs=None):
```
JeanKossaifi, line 782: we could add a parameter `svd_fun="svd"` and carefully document it? This could then accept "numpy_svd", and other when applicable. That would probably be cleaner than what we have now.

tensorly/backend/core.py
```diff
@@ -818,22 +800,32 @@ def truncated_svd(self, matrix, n_eigenvecs=None, **kwargs):
...
+        full_matrices = (n_eigenvecs is None) or (n_eigenvecs > min_dim)
 
-        U, S, V = self.svd(matrix)
+        U, S, V = self.svd(matrix, full_matrices=full_matrices)
         U, S, V = U[:, :n_eigenvecs], S[:n_eigenvecs], V[:n_eigenvecs, :]
         return U, S, V
 
     index = Index()
+
+    @property
```
JeanKossaifi, line 825: If we go for that solution we could have a function `available_svd_funs` instead of this? What do you think?

tensorly/backend/core.py
aarmey: If kr's implementation is set in the frontend, perhaps we can just remove T.kr from the backend altogether?

JeanKossaifi:
Thanks a lot @aarmey, this is a large but necessary change to make the behaviour of SVD clearer to users. 

Overally looks good to me, the change from `_reverse` to `flip` is great. 
I think we may want to use the opportunity to add some documentation, perhaps in the user_guide, either in the [backend](https://github.com/tensorly/tensorly/blob/main/doc/user_guide/backend.rst) section or in a new page? We can there explain the difference between the three provided interfaces. 

I'd really rather avoid converting to NumPy as much as possible as it breaks the computational graph in some backends and causes surprising issues (e.g. if running the method on GPU).

On the API side, I'm wondering whether we want to provide a function `partial_svd` (or even just `svd`) and an additional `implementation` (or `method`?) :

```python
def partial_svd(matrix, n_eigenvecs, implementation='truncate')
```
where 
```
implementation : {'truncate', 'numpy', 'eish'}

          - truncate : uses the backend's native SVD function and truncates it
          - numpy : uses the efficient Sparse SVD from Scipy to only compute the required eigenvalues
          - symeig : uses the native backend's symeig to leverage the symmetry in the covariance matrix when computing the SVD
```

What do you think?

aarmey:
I think this SVD interface looks great. Indeed, I partly wanted these changes to move more of the SVD methods to the proper backend as opposed to silently going back to NumPy. I like that your suggestion will make it clear whether the method is using the indicated backend or Scipy.

Perhaps I can open an issue with your notes here? I expect that changing the SVD interface is going to be another large, complicated change.

JeanKossaifi:
Yes, that's a good idea. There are some conversions to NumPy in the reconstruction errors, would be great to remove those before merging if possible!

aarmey:
@JeanKossaifi removed.

JeanKossaifi (approved):
Thanks @aarmey for the all great changes! 🥇 
Merging.

tensorly/backend/core.py
merajhashemi: Why is `**kwargs` removed?

JeanKossaifi: I don't think it was actually used anywhere but we can add it back if it breaks things downstream.

tensorly/backend/pytorch_backend.py
```diff
@@ -174,133 +184,22 @@ def argmax(input, axis=None):
...
-        -------
-        reversed_tensor : for a 1-D tensor, returns the equivalent of
-                        tensor[::-1] in NumPy
-        """
-        indices = torch.arange(tensor.shape[axis] - 1, -1, -1, dtype=torch.int64)
-        return tensor.index_select(axis, indices)
+    def svd(self, X, full_matrices=True):
+        # The torch SVD has accuracy issues. Try again when torch.linalg is stable.
+        ctx = self.context(X)
+        X = self.to_numpy(X)
```
merajhashemi, line 190: Pytorch 1.8 is now the stable release, should we consider replacing this with `torch.linalg.svd`? 
As a side note, the linear algebra functions in torch will be deprecated with their `linalg` counterparts in 1.9. For example, `torch.qr` is already deprecated in 1.8 we should replace it with its `linalg` equivalent.

JeanKossaifi: Yes, I was actually looking forward to switching `torch.linalg.svd`! Thanks @merajhashemi !

aarmey: Thanks for the note @merajhashemi. I'll open an issue to make sure this is done.

PR #221, opened 2021-01-01
Fix #220 - Improving naming consistency

### LINKED ISSUE #220: Enhancement: Improve naming consistency
Almost all decomposition implementations are stored in a file with the name `_{{decomposition_name}}`. The exceptions are PARAFAC2 and Robust PCA. Should we rename these modules to `_parafac2.py` and `_robust_decomposition.py`, or is there a reason for these exceptions?

MarieRoald:
This pull request fixes the naming inconsistencies mentioned in #220. Also, I moved the docstring of the PARAFAC2 class from the `__init__` function to the class itself, which makes it accessible for Sphinx.


JeanKossaifi:
Thanks @MarieRoald, merging!

PR #224, opened 2021-01-25
Hals feature

caglayantuna:
# HALS for Tensorly
This pull request adds Hierarchical Alternating Least Square (HALS) algorithm and non-negative cp (NNCP) with HALS class. We adapted this code from  a matlab code (https://sites.google.com/site/nicolasgillis/) to tensorly library with @cohenjer.

# Motivation
First motivation is to have better NNCP algorithm in tensorly. In tensorly, NNCP uses multiplicative update (MU) algorithm. According to the literature (Gillis, 2012), HALS is more robust than MU. Our second motivation is to have sparse NNCP model using the same algorithm. 

We have tested our code with different publicly available image data (hyperspectral, multitemporal satellite images and a multispectral image) and we have obtained better RMSE and SSIM metric results with our HALS code compared to MU algorithm in tensorly. If it is necessary, we could share a notebook or we can submit another pull request to tensorly examples.

# Source Modification and Implementation Details
* **tenalg/proximal.py:** 

Here, we added *"hals_nnls_approx"* and *"hals_nnls_exact"* functions as nonnegative least square (nnls) algorithms. 

First one is an approximate solution. It has three stopping criteria and 500 iteration as default. One of the stopping criteria relies on *"delta"* value in the function which can be changed by the user. 

Second function is an exact solution. It has less stopping criteria (2) and more iteration compared to approximate solution. Although exact solution gives more precise result, process takes longer.

* **_nn_cp.py:**

Here, we added *"non_negative_parafac_hals"* function and *"CPNN_Hals"* class. 

First function calls nnls functions from proximal.py and it has an option to choose one of them. This option is 'approx' as default. There is a sparsity coefficient option in order to tackle with sparse data. To use this option, one should have a list of which length should equal to number of modes.

We have created a new class *"CPNN_Hals"* which is very similar to existing *"CPNN"* class in this file. Our class includes *"non_negative_parafac_hals"* function as a difference. Additionally, we have added *"cp_to_tensor"* function to this class in order to reconstruct tensor from extracted factors. For instance, this class can be used as:

```python
nncp=CPNN_Hals(rank=rank,init=init)
cptensor=nncp.fit_transform(tensor)
cp_reconstruction_hals = nncp.cp_to_tensor() 
``` 

* **test_cp.py:**

Lastly, we have added a test function *"test_non_negative_parafac_hals"* for *"non_negative_parafac_hals"* function. Simply, we have copied *"test_non_negative_parafac"* and changed related function. We have tested our function with this test function as well.

# Possible Improvements
At the moment, this code works with numpy, pytorch and tensorflow backends. However, in order to use our code with pytorch and tensorflow backends, we had to add some lines to nnls functions (such as lines 214 to 226 in proximal.py) which can be improved in the future. Here, we had to use *torch.maximum* function because we wanted to implement *tl.max* with other backends in order to be consistent with tensorly. However, pytorch backend doesn't allow to use *'axis'* option in *tl.max*. 

Other problem is item assignment with tensorflow. We need to assign values to each row V[k,:] in hals algorithm, however tensorflow doesn't allow us to implement such an assignment. Therefore, we have used 'tf. Variable' and '.assign' commands with tensorflow backend. 

Besides, code has failed with Mxnet backend in the same item assignment part and we haven't tried our code with other backends (JAX, Cupy) yet.

# Reference

1- Gillis, N., & Glineur, F. (2012). Accelerated multiplicative updates and hierarchical ALS algorithms for nonnegative matrix factorization. Neural computation, 24(4), 1085-1105.  [Link](https://www.mitpressjournals.org/doi/abs/10.1162/NECO_a_00256)

### DIFF
--- a/tensorly/decomposition/_nn_cp.py
+++ b/tensorly/decomposition/_nn_cp.py
@@ -297,7 +297,7 @@ def non_negative_parafac_hals(tensor, rank, n_iter_max=100, init="svd", svd='num
     """
     Non-negative CP decomposition via HALS
 
-    Uses HALS which updates each factor columnwise, fixing every other columns, see [1]_
+    Uses Hierarchical ALS (Alternating Least Squares) which updates each factor column-wise (one column at a time while keeping all other columns fixed), see [1]_
 
     Parameters
     ----------
--- a/tensorly/decomposition/_nn_cp.py
+++ b/tensorly/decomposition/_nn_cp.py
@@ -356,8 +356,8 @@ def non_negative_parafac_hals(tensor, rank, n_iter_max=100, init="svd", svd='num
     norm_tensor = tl.norm(tensor, 2)
 
     n_modes = tl.ndim(tensor)
-    if sparsity_coefficient is None or isinstance(sparsity_coefficient, float):
-       sparsity_coefficient = [sparsity_coefficient]*n_modes
+    if sparsity_coefficients is None or isinstance(sparsity_coefficients, float):
+       sparsity_coefficients = [sparsity_coefficients]*n_modes
     
     if fixed_modes is None:
         fixed_modes = []
--- a/tensorly/decomposition/_nn_cp.py
+++ b/tensorly/decomposition/_nn_cp.py
@@ -400,7 +400,7 @@ def non_negative_parafac_hals(tensor, rank, n_iter_max=100, init="svd", svd='num
                 factors[mode] = tl.transpose(
                     hals_nnls_exact(tl.transpose(mttkrp), pseudo_inverse, tl.transpose(factors[mode]),
                                      maxiter=5000)[0])
- 
+
         if tol:
             factors_norm = cp_norm((weights, factors))
             iprod = tl.sum(tl.sum(mttkrp*factor, axis=0)*weights)
--- a/tensorly/decomposition/_nn_cp.py
+++ b/tensorly/decomposition/_nn_cp.py
@@ -20,7 +20,7 @@ from ..cp_tensor import (cp_to_tensor, CPTensor,
 
 
 def make_svd_non_negative(tensor, U, S, V, nntype):
-    """ Use NNDSVD method to transform SVD results into a non-negative form. This 
+    """ Use NNDSVD method to transform SVD results into a non-negative form. This
     method leads to more efficient solving with NNMF [1].
 
     Parameters
@@ -313,8 +313,8 @@ def non_negative_parafac_hals(tensor, rank, n_iter_max=100, init="svd", svd='num
           tolerance: the algorithm stops when the variation in
           the reconstruction error is less than the tolerance
         Default: 1e-8
-    sparsity_coefficients: array of float (as much as the number of modes)
-        The sparsity coefficients on U and V respectively.
+    sparsity_coefficients: array of float (of length the number of modes)
+        The sparsity coefficients on each factor.
         If set to None, the algorithm is computed without sparsity
         Default: [],
     fixed_modes: array of integers (between 0 and the number of modes)
@@ -681,7 +681,27 @@ class CPNN_Hals(DecompositionMixin):
         self.errors_ = errors
         return self.decomposition_
     def cp_to_tensor(self):
-        """
+        """Turns the Khatri-product of matrices into a full tensor
+
+            ``factor_matrices = [|U_1, ... U_n|]`` becomes
+            a tensor shape ``(U[1].shape[0], U[2].shape[0], ... U[-1].shape[0])``
+
+        Parameters
+        ----------
+        cp_tensor : CPTensor = (weight, factors)
+            factors is a list of factor matrices, all with the same number of columns
+            i.e. for all matrix U in factor_matrices:
+            U has shape ``(s_i, R)``, where R is fixed and s_i varies with i
+
+        mask : ndarray a mask to be applied to the final tensor. It should be
+            broadcastable to the shape of the final tensor, that is
+            ``(U[1].shape[0], ... U[-1].shape[0])``.
+
+        Returns
+        -------
+        ndarray
+            full tensor of shape ``(U[1].shape[0], ... U[-1].shape[0])``
+
         """
         return tl.cp_to_tensor(self.decomposition_)
 
--- a/tensorly/decomposition/_nn_cp.py
+++ b/tensorly/decomposition/_nn_cp.py
@@ -1,6 +1,6 @@
 import numpy as np
 import warnings
-
+import time
 import tensorly as tl
 from ._base_decomposition import DecompositionMixin
 from ..random import random_cp
@@ -291,6 +291,142 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
         return cp_tensor
 
 
+def non_negative_parafac_hals(tensor, rank, n_iter_max=100, init="svd", svd='numpy_svd', tol=1e-7,
+                              sparsity_coefficients=[], fixed_modes=[],
+                              verbose=False, return_errors=False):
+    """
+    Non-negative CP decomposition
+
+    Uses HALS which updates each factor columnwise, fixing every other columns, see [1]_
+
+    Parameters
+    ----------
+    tensor : ndarray
+    rank   : int
+            number of components
+    n_iter_max : int
+                 maximum number of iteration
+    init : {'svd', 'random'}, optional
+    svd : str, default is 'numpy_svd'
+        function to use to compute the SVD, acceptable values in tensorly.SVD_FUNS
+    tol : float, optional
+          tolerance: the algorithm stops when the variation in
+          the reconstruction error is less than the tolerance
+        Default: 1e-8
+    sparsity_coefficients: array of float (as much as the number of modes)
+        The sparsity coefficients on U and V respectively.
+        If set to None, the algorithm is computed without sparsity
+        Default: [],
+    fixed_modes: array of integers (between 0 and the number of modes)
+        Has to be set not to update a factor, 0 and 1 for U and V respectively
+        Default: []
+    verbose: boolean
+        Indicates whether the algorithm prints the successive
+        reconstruction errors or not
+        Default: False
+    return_errors: boolean
+        Indicates whether the algorithm should return all reconstruction errors
+        and computation time of each iteration or not
+        Default: False
+
+    Returns
+    -------
+    factors : ndarray list
+            list of positive factors of the CP decomposition
+            element `i` is of shape ``(tensor.shape[i], rank)``
+    errors: list
+        A list of reconstruction errors at each iteration of the algorithm.
+    toc: list
+        A list with accumulated time at each iterations
+
+   fixed_modes = [], normalize = [False, False, False],
+               verbose = True, return_errors = False)
+
+    References
+    ----------
+    [1]: N. Gillis and F. Glineur, Accelerated Multiplicative Updates and
+    Hierarchical ALS Algorithms for Nonnegative Matrix Factorization,
+    Neural Computation 24 (4): 1085-1105, 2012.
+    """
+
+    weights, factors = initialize_nn_cp(tensor, rank, init=init, svd=svd,
+                                     random_state=None,
+                                     normalize_factors=False)
+
+    norm_tensor = tl.norm(tensor, 2)
+
+    # set init if problem
+    nb_modes = len(tensor.shape)
+    if sparsity_coefficients == None or len(sparsity_coefficients) != nb_modes:
+        print(
+            "Irrelevant number of sparsity coefficient (different from the number of modes), they have been set to None.")
+        sparsity_coefficients = [None for i in range(nb_modes)]
+    if fixed_modes == None:
+        fixed_modes = []
+
+    # Avoiding errors
+    for fixed_value in fixed_modes:
+        sparsity_coefficients[fixed_value] = None
+
+    # Generating the mode update sequence
+    modes_list = [mode for mode in range(tl.ndim(tensor)) if mode not in fixed_modes]
+
+    # initialisation - declare local varaibles
+    rec_errors = []
+    toc = []
+
+    # Iterate over one step of NTF
+    for iteration in range(n_iter_max):
+        # One pass of least squares on each updated mode
+        for mode in modes_list:
+
+            # Computing Hadamard of cross-products
+            pseude_inverse = tl.tensor(tl.ones((rank, rank)), **tl.context(tensor))
+            for i, factor in enumerate(factors):
+                if i != mode:
+                    pseude_inverse *= tl.dot(tl.transpose(factor), factor)
+
+            if not iteration and weights is not None:
+                # Take into account init weights
+                mttkrp = unfolding_dot_khatri_rao(tensor, (weights, factors), mode)
+            else:
+                mttkrp = unfolding_dot_khatri_rao(tensor, (None, factors), mode)
+
+
+            # Call the hals resolution with nnls, optimizing the current mode
+
+            factors[mode] = tl.transpose(
+                hals_nnls_acc(tl.transpose(mttkrp), pseude_inverse, tl.transpose(factors[mode]),
+                             maxiter=100,sparsity_coefficient=sparsity_coefficients[mode])[0])
+
+            factors_norm = tl.sum(tl.sum(pseude_inverse * tl.dot(tl.conj(tl.transpose(factors[mode])), factors[mode])))
+            rec_error = norm_tensor ** 2 + factors_norm - 2 * tl.dot(tl.tensor_to_vec(factors[mode]),
+                                                                     tl.tensor_to_vec(mttkrp))
+            rec_error = rec_error ** (1 / 2) / norm_tensor
+
+        if tol:
+            rec_errors.append(rec_error)
+
+            if iteration > 1:
+                if verbose:
+                    print('reconstruction error={}, variation={}.'.format(
+                        rec_errors[-1], rec_errors[-2] - rec_errors[-1]))
+
+                if tol and abs(rec_errors[-2] - rec_errors[-1]) < tol:
+                    if verbose:
+                        print('converged in {} iterations.'.format(iteration))
+                    break
+
+    cp_tensor = CPTensor((weights, factors))
+    if return_errors:
+        return cp_tensor, rec_errors
+    else:
+        return cp_tensor
+
+
+
+
+
 class CPNN(DecompositionMixin):
     """Non-Negative Candecomp-Parafac decomposition via Alternating-Least Square
 
@@ -399,6 +535,7 @@ class CPNN(DecompositionMixin):
         CPTensor
             decomposed tensor
         """
+
         cp_tensor, errors = non_negative_parafac(tensor, rank=self.rank,
                                                  n_iter_max=self.n_iter_max,
                                                  tol=self.tol,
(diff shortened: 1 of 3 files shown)


tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+    cnt = 1
+    eps = 1
+
+
+    while eps >= delta * eps0 and cnt <= 1 + 0.5* rho and cnt <= maxiter:
+        nodelta = 0
+        for k in range(r):
+
+            if UtU[k, k] != 0:
+                if sparsity_coefficient != None: # Modifying the objective function for sparsification
```
JeanKossaifi, line 213:
```suggestion
                if sparsity_coefficient is not None: # Modifying the objective function for sparsification
```

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+                    if tl.get_backend() == 'pytorch':
```
JeanKossaifi, line 214: It's best to avoid backend specific code.

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+                        import torch
+                        deltaV = torch.maximum((UtM[k, :] - UtU[k, :] @ V - sparsity_coefficient * tl.ones(n)) / UtU[k, k],
+                                        -V[k, :])
+                    else:
+                        deltaV = tl.max([(UtM[k, :] - UtU[k, :] @ V - sparsity_coefficient * tl.ones(n)) / UtU[k, k],
+                                        -V[k, :]],axis=0)
```
JeanKossaifi, line 220: We could probably update the backend to include both `max` and `maximum`, following numpy's convention that max does a reduce while maximum is elementwise.

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+                    if tl.get_backend()=='tensorflow':
+                        import tensorflow as tf
+                        V=tf.Variable(V,dtype='float')
+                        V[k, :].assign(V[k, :] + deltaV)
```
JeanKossaifi, line 224: You can use `tl.index_update` for all backends.

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+                    else:
+                        V[k, :] = V[k, :] + deltaV
+
+                nodelta = nodelta + tl.dot(deltaV, tl.transpose(deltaV))
+
+                # Safety procedure, if columns aren't allow to be zero
+                if nonzero_row and (V[k, :] == 0).all():
+                    V[k, :] = 1e-16 * tl.max(V)
```
JeanKossaifi, line 248: You could use `tl.eps(V.dtype)`, or check how we checked for zero values in cp_normalize.

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+                    sqrt_n = 1/n ** (1/2)
+                    V[k,:] = [sqrt_n for i in range(n)]
+        if cnt == 1:
+            eps0 = nodelta
+
+        rho_up=tl.shape(V)[0]*tl.shape(V)[1]+tl.shape(V)[1]*r
+        rho_down=tl.shape(V)[0]*r+tl.shape(V)[0]
+        rho=1+(rho_up/rho_down)
+        eps = nodelta
+        cnt += 1
```
JeanKossaifi, line 267: cnt could be directly defined in the for loop as `for iter in n_iter:`, then there is no need for this update. Would be good to come up with more explicit names for all variables.

The `eps0 = nodetla` and `eps = nodelta` is a little confusing too.

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+
+    return V, eps, cnt, rho
```
JeanKossaifi, line 269: New line.

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+def hals_nnls_exact(UtM, UtU, in_V, maxiter,delta=10e-12,sparsity_coefficient=None):
+
```
JeanKossaifi, line 271:
```suggestion
```

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+    """
+    =================================
```
JeanKossaifi, line 273: We don't typically wrap the doc in a title, we just put the description straight after the quotes.

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+    M is m by n, U is m by r, V is r by n.
+    All matrices are nonnegative componentwise.
+
+    The NNLS unconstrained problem, as defined in [1], solve the following problem:
+
+            min_{V >= 0} ||M-UV||_F^2
+
+    The matrix V is updated linewise.
+
+    The update rule of the k-th line of V (V[k,:]) for this resolution is:
```
JeanKossaifi, line 288:
```suggestion
    The update rule of the k-th line of V (V[k,:]) for this resolution is::
```

JeanKossaifi (commented):
Thanks a lot @caglayantuna and @cohenjer, it's a great PR!

I made a first pass throughout and left comments directly in the code. Would be good to follow pep8 where applicable.

**Naming of the parameters**
It's good (but hard!) to try and have a clear names as possibles for the parameters.  
- sparsity_coefficient: n_sparse_elements?
- CPNN_Hals: CP_NN_HALS ?
- CPNN : CP_NN

Similarly, instead of following necessarily the paper, we can try to have explicit names for the variables such as `esp0`, etc. This is especially helpful, combined with the documentation, for users who may not have read the paper or know the method at all (or even us when we re-read this in the future :) )

Generally, I like to focus on readability and making the code easily maintainable.

Would it make sense to merge `hals_update_approx` and `exact` and add an optional parameter, e.g. `exact=False`? 

**For convergence**, we have two criteria for convergence in CP, we may want to add this but we can do in a subsequent PR.

cohenjer:
Hey @JeanKossaifi , I think I agree with all your comments, we will make answers and modifications accordingly, thanks for the careful review. You are right we should pay more attention to variables names in later PR and improve them for this PR. This code is an adaptation of another code, ``passed down in generations'' from my tutor to me and then to my PhD student and now this PR, so some names date from back then.

About the exact vs approx HALS, I think we can merge both but we need to improve the doc to explain why there are two different versions.

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,6 +291,142 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
         return cp_tensor
 
 
+def non_negative_parafac_hals(tensor, rank, n_iter_max=100, init="svd", svd='numpy_svd', tol=1e-7,
+                              sparsity_coefficients=[], fixed_modes=[],hals='approx',
```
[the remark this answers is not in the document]
cohenjer: Actually we used the same syntax as in _cp.py `parafac` function, which input `fixed_modes` has default [] (my bad since this was also from my older PR). We can change both `parafac` and `non_negative_parafac_hals` to have defaults `None` wherever we used `[]`.

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,6 +291,142 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
...
+    """
+
+    weights, factors = initialize_nn_cp(tensor, rank, init=init, svd=svd,
+                                     random_state=None,
+                                     normalize_factors=False)
+
+    norm_tensor = tl.norm(tensor, 2)
+
+    nb_modes = len(tensor.shape)
+    if sparsity_coefficients == None or len(sparsity_coefficients) != nb_modes:
```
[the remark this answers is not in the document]
caglayantuna: I have changed these lines, thanks

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,6 +291,142 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
...
+    # Generating the mode update sequence
+    modes_list = [mode for mode in range(tl.ndim(tensor)) if mode not in fixed_modes]
+
+    # initialisation - declare local varaibles
+    rec_errors = []
+
+    # Iteratation
+    for iteration in range(n_iter_max):
+        # One pass of least squares on each updated mode
+        for mode in modes_list:
```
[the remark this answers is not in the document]
caglayantuna: Yes, it is updated

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,6 +291,142 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
...
+        sparsity_coefficients = [None for i in range(nb_modes)]
+    if fixed_modes == None:
+        fixed_modes = []
+
+    # Avoiding errors
+    for fixed_value in fixed_modes:
+        sparsity_coefficients[fixed_value] = None
+
+    # Generating the mode update sequence
+    modes_list = [mode for mode in range(tl.ndim(tensor)) if mode not in fixed_modes]
```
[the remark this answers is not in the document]
caglayantuna: I understand 'modes' is enough but it is 'modes_list' in non_negative_parafac function. We may open a new pull request in future, in order to make all these names and signatures consistent.

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,6 +291,142 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
...
+        # One pass of least squares on each updated mode
+        for mode in modes_list:
+
+            # Computing Hadamard of cross-products
+            pseudo_inverse = tl.tensor(tl.ones((rank, rank)), **tl.context(tensor))
+            for i, factor in enumerate(factors):
+                if i != mode:
+                    pseudo_inverse = pseudo_inverse*tl.dot(tl.transpose(factor), factor)
+
+            if not iteration and weights is not None:
```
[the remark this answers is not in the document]
caglayantuna: We are planning to submit a PR related to this issue

tensorly/decomposition/_nn_cp.py
```diff
@@ -410,9 +547,163 @@ def fit_transform(self, tensor):
...
+                                                      init=self.init,
+                                                      svd=self.svd,
+                                                      hals=self.hals,
+                                                      verbose=self.verbose,
+                                                      return_errors=True)
+
         self.decomposition_ = cp_tensor
         self.errors_ = errors
         return self.decomposition_
+    def cp_to_tensor(self):
```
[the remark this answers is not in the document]
caglayantuna: Ok, I removed this function.

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
     """
     U, _, V = tl.partial_svd(matrix, n_eigenvecs=min(matrix.shape))
     return tl.dot(U, V)
+def hals_nnls_approx(UtM, UtU, in_V, maxiter=500,delta=10e-8,
```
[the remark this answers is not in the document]
caglayantuna: I changed the names according to your suggestion

tensorly/tenalg/proximal.py
```diff
@@ -91,4 +91,316 @@ def procrustes(matrix):
...
+    Hierarchical ALS Algorithms for Nonnegative Matrix Factorization,
+    Neural Computation 24 (4): 1085-1105, 2012.
+
+    [2] J. Eggert, and E. Korner. "Sparse coding and NMF."
+    2004 IEEE International Joint Conference on Neural Networks
+    (IEEE Cat. No. 04CH37541). Vol. 4. IEEE, 2004.
+
+    """
+
+    r, n = tl.shape(UtM)
```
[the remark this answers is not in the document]
caglayantuna: Yes, I changed the names.

tensorly/tenalg/proximal.py
caglayantuna: I removed all the backend options

tensorly/tenalg/proximal.py
caglayantuna: Here, I have a problem with pytorch. Yes, if tensorly includes elementwise for each backend, problem will be solved.

tensorly/tenalg/proximal.py
caglayantuna: Thanks, I didn't know about this function. Actually, it is very important but not presented in the website.

tensorly/tenalg/proximal.py
caglayantuna: As I understand, you mean `(V[k, :] == 0).all()` part. Now, I changed it to  `tl.all(V[k, :] == 0)`.

tensorly/tenalg/proximal.py
caglayantuna: Now, there is no `while` in the code. All the names are changed. `eps0=> rec_error0` and `eps=>rec_error`. There is no `nodelta`.

tensorly/decomposition/_nn_cp.py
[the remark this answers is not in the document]
caglayantuna: Thanks for the suggestion, I updated this part of the code.

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,6 +291,142 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
...
+            iprod = tl.sum(tl.sum(mttkrp*factor, axis=0)*weights)
+            rec_error = tl.sqrt(tl.abs(norm_tensor**2 + factors_norm**2 - 2*iprod)) / norm_tensor
+            rec_errors.append(rec_error)
+
+            if iteration > 1:
+                if verbose:
+                    print('reconstruction error={}, variation={}.'.format(
+                        rec_errors[-1], rec_errors[-2] - rec_errors[-1]))
+
+                if tol and abs(rec_errors[-2] - rec_errors[-1]) < tol:
```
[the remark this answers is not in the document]
caglayantuna: As I understand, you mean convergence criterions (`abs_rec_error` and `rec_error`) in parafac functions. I have updated our code with these options.

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,6 +291,142 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
...
+                mttkrp = unfolding_dot_khatri_rao(tensor, (weights, factors), mode)
+            else:
+                mttkrp = unfolding_dot_khatri_rao(tensor, (None, factors), mode)
+
+
+            # Call the hals resolution with nnls, optimizing the current mode
+            if hals=='approx':
+                factors[mode] = tl.transpose(
+                      hals_nnls_approx(tl.transpose(mttkrp), pseudo_inverse, tl.transpose(factors[mode]),
+                             maxiter=100,sparsity_coefficient=sparsity_coefficients[mode])[0])
```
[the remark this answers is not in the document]
caglayantuna: I agree but I couldn't find another option for now.

cohenjer:
Also I am not sure what is the best way to do this, but it would be nice to credit also Axel Marmoret (https://gitlab.inria.fr/amarmore) who made the base version of that code, and has a package nn_fac (https://pypi.org/project/nn-fac/) which basically contains the same algorithms we implement in this PR.

tensorly/tenalg/proximal.py
cohenjer: Actually @JeanKossaifi we have no experience with the backend system, what would it take to support both max and maximum?

JeanKossaifi:
I agree, sounds like he should be acknowledged-- you could sync with him and add him in the author list in the files changed?

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,6 +291,142 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
...
+    -------
+    factors : ndarray list
+            list of positive factors of the CP decomposition
+            element `i` is of shape ``(tensor.shape[i], rank)``
+    errors: list
+        A list of reconstruction errors at each iteration of the algorithm.
+    toc: list
+        A list with accumulated time at each iterations
+
+   fixed_modes = [], normalize = [False, False, False],
```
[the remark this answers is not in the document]
caglayantuna: yes, that was a mistake. I removed these lines

caglayantuna:
Hey @JeanKossaifi, thank you for your comments to make our PR completely compatible with Tensorly. Since we are planning to submit many PRs in the future, these comments will be useful for me to not to do same mistakes again.  

We have updated our PR according to your suggestions and I replied to your comments in case it is needed. I would like to add some notes for specific topics:

* I haven't changed the name sparsity_coefficients to n_sparse_elements because sparsity_coefficients are not the number of sparse elements. If this name is not good for you, we may think about it.
* For convergence, I have copied the `cvg_criterion` part from the non_negative_parafac. If you didn't mean that, could you give me more details about it?
* I haven't changed the things which are not related to this PR such as `fixed_modes=[]` and `modes_list` in other functions. I am planning to do another PR for these compatibility issues.

I hope the code is clear now. Besides, you can find a notebook here  [Link](https://github.com/caglayantuna/Tensorly_notebook/blob/main/nncp_Notebook%20for%20Pull%20Request.ipynb). In this notebook, I have used the same image in tensorly experiments and compared the results of `non_negative_parafac` and `non_negative_parafac_hals` according to some metrics and computational time. I am looking forward to hear your comments about the last version of our PR.

tensorly/tenalg/proximal.py
JeanKossaifi: I also was talking about the 1e-16: you could use `tl.eps(V.dtype)` instead of that hard-coded value so it matches the dtype of V.

tensorly/tenalg/proximal.py
```diff
@@ -92,3 +95,159 @@ def procrustes(matrix):
...
+                norm = tl.norm(V[k,:])
+                if norm != 0:
+                    V[k,:] /= norm
+                else:
+                    sqrt_n = 1/n_col_M ** (1/2)
+                    V[k,:] = [sqrt_n for i in range(n_col_M)]
+        if iteration == 1:
+            rec_error0 = rec_error
+
+        rho_up=tl.shape(V)[0]*tl.shape(V)[1]+tl.shape(V)[1]*rank
```
JeanKossaifi, line 243: Could we find a better name for rho*?

tensorly/tenalg/proximal.py
JeanKossaifi: It should just be a matter of adding them in each backend, that shouldn't be an issue, I'll have a look. 
On this line, @ is not supported by all backends, it's (more verbose but) better to use `tl.dot()`

Are you using tl.max over a list of two elements? I wonder if just using Python's builtin `max` would work. 
I also don't think the axis is needed here since it's a list withh just two elements. 
Finally, I also don't think you need the tl.ones, you can directly do array - int and the int is broadcast to the whole array.

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,7 +294,145 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
...
+    factors : ndarray list
+            list of positive factors of the CP decomposition
+            element `i` is of shape ``(tensor.shape[i], rank)``
+    errors: list
+        A list of reconstruction errors at each iteration of the algorithm.
+
+    References
+    ----------
+    [1]: N. Gillis and F. Glineur, Accelerated Multiplicative Updates and
+    Hierarchical ALS Algorithms for Nonnegative Matrix Factorization,
```
JeanKossaifi, line 353: I _think_ this would have to be indented to match with the previous line (`N. Gillis`) so it's parsed correctly as part of the same citation.

tensorly/decomposition/_nn_cp.py
```diff
@@ -291,7 +294,145 @@ def non_negative_parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_sv
...
+                    pseudo_inverse = pseudo_inverse*tl.dot(tl.transpose(factor), factor)
+
+            if not iteration and weights is not None:
+                # Take into account init weights
+                mttkrp = unfolding_dot_khatri_rao(tensor, (weights, factors), mode)
+            else:
+                mttkrp = unfolding_dot_khatri_rao(tensor, (None, factors), mode)
+
+            # Call the hals resolution with nnls, optimizing the current mode
+            factors[mode] = tl.transpose(
```
JeanKossaifi, line 398: I think it would be clearer to have this in two lines, to avoid the potentially confusing indexing of the tuple returned by hals_nnls:
```python
nn_factor, _, _ = hals_nnls(...)
factors[mode] = tl.transpose(nn_factor)
```

tensorly/decomposition/_nn_cp.py
```diff
@@ -410,6 +552,139 @@ def fit_transform(self, tensor):
...
+
+    def __repr__(self):
+        return f'Rank-{self.rank} Non-Negative CP decomposition.'
+
+
+
+class CP_NN_HALS(DecompositionMixin):
+    """Non-Negative Candecomp-Parafac decomposition via Alternating-Least Square
+
+        Computes a rank-`rank` decomposition of `tensor` [1]_ such that,
```
JeanKossaifi, line 569: This should be de-indented to match the beginning of the previous line.
Then either use:
```rst 
Computes a rank-`rank` decomposition of `tensor` [1]_ such that:: 

      ``tensor = [|weights; factors[0], ..., factors[-1] |]``.
``` 
if you want an explicit bloc quote or simply
```rst
Computes a rank-`rank` decomposition of `tensor` [1]_ such that ``tensor = [|weights; factors[0], ..., factors[-1] |]``.
```

I've been trying to harmonize the docstrings and iron these quirks out so it all renders nicely in the html/pdf documentation.

tensorly/decomposition/_nn_cp.py
```diff
@@ -410,6 +552,139 @@ def fit_transform(self, tensor):
...
+                (tensor.shape[i], rank)
+            * sparse_component : nD array of shape tensor.shape. Returns only if `sparsity` is not None.
+
+        errors : list
+            A list of reconstruction errors at each iteration of the algorithms.
+
+        References
+        ----------
+        .. [1] T.G.Kolda and B.W.Bader, "Tensor Decompositions and Applications",
+        SIAM REVIEW, vol. 51, n. 3, pp. 455-500, 2009.
```
JeanKossaifi, line 629: The SIAM should probably be aligned with the beginning of the previous line (T.G.Kolda)

tensorly/tenalg/proximal.py
```diff
@@ -92,3 +95,159 @@ def procrustes(matrix):
...
+    Non Negative Least Squares (NNLS)
+
+    Computes an approximate solution of a nonnegative least
+    squares problem (NNLS) with an exact block-coordinate descent scheme.
+    M is m by n, U is m by r, V is r by n.
+    All matrices are nonnegative componentwise.
+
+    The NNLS unconstrained problem, as defined in [1], solve the following problem:
+
+            min_{V >= 0} ||M-UV||_F^2
```
JeanKossaifi, line 112: Either 
```rst 
following problem::

    # 4 space indent etc...
```

Generally, in the docstring, it should be readable rather than correct LaTex.
If it is helpful for comprehension, you can add more detail in the 
```rst
Notes
------
``` 
section, using the \`math\`:formula: direction or
```rst
.. math::

   formula
``` 
for blocks of math (see https://github.com/tensorly/tensorly/blob/main/tensorly/decomposition/robust_decomposition.py for an example of math in the doc, note the escaped backslash in the formula) 

Finally, a longer, more detailed and more technical explanation is perfect for the [user-guide](https://github.com/tensorly/tensorly/tree/main/doc/user_guide)!

tensorly/tenalg/proximal.py
```diff
@@ -92,3 +95,159 @@ def procrustes(matrix):
...
+        number of loops authorized by the error stop criterion
+    iteration: integer
+        final number of update iteration performed
+    rho: float
+        number of loops authorized by the time stop criterion
+
+    References
+    ----------
+    [1]: N. Gillis and F. Glineur, Accelerated Multiplicative Updates and
+    Hierarchical ALS Algorithms for Nonnegative Matrix Factorization,
```
JeanKossaifi, line 184: Ref. alignment.

JeanKossaifi:
Hi @caglayantuna,

Thanks for making the changes, it looks much cleaner! When reviewing this PR I actually realized there was a small issue in the Github action running the test which mean it only got tested with the NumPy backend, this should be fixed now. 
There are a couple of formatting issues, checkout [PEP8](https://www.python.org/dev/peps/pep-0008/) for the guidelines. You can install and run `flake8` to check for pep8 issues.

The notebook is great, we can adapt it (e.g. just use RMSE and relative rec error) and add it as an [example](https://github.com/tensorly/tensorly/tree/main/examples) for the [gallery](http://tensorly.org/dev/auto_examples/index.html).

tensorly/tenalg/proximal.py
caglayantuna: I removed tl.ones and @. 

I am using tl.max over two lists. The target is to find elementwise maximum for each location in kth row. So, the result is also a list which has `len(V(k,:))` elements.

tensorly/tenalg/proximal.py
caglayantuna: ok, I changed that part as well

tensorly/tenalg/proximal.py
caglayantuna: New names are nominator, denominator and cvg_criterion.

tensorly/tenalg/proximal.py
caglayantuna: I have prepared the mathematical equations according to your second suggestion. I hope it is ok now.

caglayantuna:
Hi @JeanKossaifi, thanks for your comments. Now, there is no PEP8 issue except longer lines which could be acceptable and no alignment problem in the doc. Pytorch and jax backends are failing because of the `tl.max` line. I have to use 'axis' as I explained in the related conversation. I couldn't understand why mxnet backend fails.

I will submit a PR to tensorly examples.

JeanKossaifi:
Thanks @caglayantuna! In most cases it PEP8 makes the code more readable but not always, no need to always adhere to it at all cost! :)

This week is really busy for me but as soon but I'll review as soon as I can and fix the max issue (as a side note we could also use something like `tl.where(U > V, U, V)` ) and check why the MXNet test fails.

JeanKossaifi:
About the errors with `check_random_state`: in #212 it was moved to the backend to allow for better reproducibility (and let backends use their own RNG).

JeanKossaifi:
Awesome, thanks for fixing the issues! Looks good to me and seems like all the tests are passing.
Is there anything else you wanted to add @caglayantuna @cohenjer? Otherwise, happy to merge!

cohenjer:
Thanks for the great reviews and ideas @JeanKossaifi. We don't have anything left we wish to merge right now, we are working on some nice examples to showcase the new algorithm but they can wait until we PR nonnegative Tucker using the same techniques.

JeanKossaifi:
Thanks for this great new feature @caglayantuna and @cohenjer ! 🚀 
@caglayantuna that's an amazing first PR, great to have you onboard! 

Merging.

PR #225, opened 2021-02-06
Upgrade MXNet backend to MXNet.np interface

aarmey:
This also removes the need for some of the MXNet-specific workarounds.


aarmey:
@JeanKossaifi this should be ready to review when you have a chance.

JeanKossaifi:
Thanks @aarmey, it's great to have this!
Can you merge in the main branch? I fixed an error in the github action script that caused the tests to always be run using TensorLy.

aarmey:
Done @JeanKossaifi. MXNet seems to be extra picky about types, so I needed to adjust a few places where ints were being passed in. Everything is working aside from the robust PCA test, which I can't seem to fix. Has something changed about it recently?

tensorly/contrib/decomposition/tests/test_mps_decomposition_cross.py
```diff
@@ -9,14 +9,11 @@
 from ....random import check_random_state
 from tensorly.testing import assert_
 
-skip_if_tensorflow = pytest.mark.skipif(tl.get_backend() == "tensorflow",
-                                        reason="Operation not supported in TensorFlow")
-skip_if_jax = pytest.mark.skipif(tl.get_backend() == "jax",
-                                        reason="Operation not supported in JAX")
+skip_if_backend = pytest.mark.skipif(tl.get_backend() in ("tensorflow", "jax", "mxnet"),
+                                     reason=f"Operation not supported in {tl.get_backend()}")
```
JeanKossaifi, line 13: Why are we skipping for MXNet? I am planning to fix this for JAX and tensorflow too, should be mostly a matter of doing the index assignment with index_update.

JeanKossaifi:
The issue with the Robust PCA must be due to the changes in the MXNet backend: the test still passes for the other backends and the last changes where made almost a year ago (same for the test, only the `random_kruskal` was replaced by `random_cp`). Could it be due to the fact that the dtype is no longer set to float32 by default?

aarmey:
Ok, thanks. I'll try and track down the change...

tensorly/backend/mxnet_backend.py
```diff
@@ -1,41 +1,41 @@
...
+        return {'dtype': tensor.dtype}
 
     @staticmethod
-    def tensor(data, ctx=mx.cpu(), dtype=numpy.float32):
+    def tensor(data, dtype=None):
         if dtype is None and isinstance(data, numpy.ndarray):
             dtype = data.dtype
-        return nd.array(data, ctx=ctx, dtype=dtype)
+        elif dtype is None:
+            dtype = np.float64
```
JeanKossaifi, line 29: Should we just not set anything? In NumPy the [default is None](https://numpy.org/doc/stable/reference/generated/numpy.array.html#numpy.array) and in that case the best dtype is used, it seems that [MXNet](https://mxnet.apache.org/versions/master/api/python/docs/api/np/generated/mxnet.np.array.html#mxnet.np.array) also has a default to None.

JeanKossaifi:
If the issue just comes from a dtype issue, then setting the default to None for the dtype as in NumPy might solve the issue. We can then let NumPy/MXNet choose the best dtype based on the input.

tensorly/contrib/decomposition/_tt_cross.py
```diff
@@ -426,7 +426,7 @@ def maxvol(A):
         A_new = A_new - A_new * tl.reshape(projection, (tl.shape(A_new)[0], 1))
 
         # Delete the selected row
-        mask.pop(max_row_idx)
+        mask.pop(tl.to_numpy(max_row_idx))
```
aarmey, line 429: MXNet does not automatically make scalars standard Python types, and so this is required to work with pop.

tensorly/contrib/decomposition/tests/test_mps_decomposition_cross.py
```diff
@@ -69,8 +66,9 @@ def test_tensor_train_cross_2():
         first_error_message += str(r_k) + " > " + str(rank[k+1])
         assert(r_k<=rank[k+1]), first_error_message
 
-@skip_if_jax
-@skip_if_tensorflow
+
+@skip_if_backend
+@pytest.mark.skipif(tl.get_backend() in ("mxnet"), reason=f"MXNet bug in advanced indexing (Issue 18919).")
```
aarmey, line 71: I've gotten the other TT methods to work, but this one has an indexing problem where `core` is first created. I believe it's the same issue as this Issue on the MXNet repository.

tensorly/decomposition/robust_decomposition.py
```diff
@@ -71,6 +71,8 @@ def robust_pca(X, mask=None, tol=10e-7, reg_E=1, reg_J=1,
     """
     if mask is None:
         mask = 1
+    else:
+        mask = T.tensor(mask, dtype=float)
```
aarmey, line 75: MXNet is not properly promoting a boolean/integer-type array when multiplied by a scalar float. The threshold was being converted to 0 below.

tensorly/contrib/decomposition/tests/test_mps_decomposition_cross.py
aarmey: Fixed (mostly). See below.

aarmey:
Everything should be resolved @JeanKossaifi

tensorly/backend/mxnet_backend.py
```diff
@@ -51,110 +49,27 @@ def ndim(tensor):
...
-        if res.shape == (1,):
-            return res.astype('int32').asscalar()
-        else:
-            return res
-
-    @staticmethod
-    def argmin(data=None, axis=None):
-        res = nd.argmin(data, axis)
-        if res.shape == (1,):
-            return res.astype('int32').asscalar()
```
JeanKossaifi, line 92: I had added these `asscalar` and typecasting where appropriate to compensate for the issue with 0-order tensors. Do you think it's better to remove them all?

tensorly/decomposition/robust_decomposition.py
JeanKossaifi: That was a tricky one, kuddos for finding/fixing it! 
That's a very odd behavior from MXNet -- we could report in the main repo.

tensorly/contrib/decomposition/_tt_cross.py
JeanKossaifi: Yes, I was explicitely typecasting in argmax to make up for that strange behaviour.

JeanKossaifi:
Looks good to me overall -- do you think it's better to not do any typecasting or `asscalar()` in the backend?

tensorly/decomposition/robust_decomposition.py
aarmey: Definitely was tricky—I got lucky by seeing `rec_X` was always 0 with a mask.

https://github.com/apache/incubator-mxnet/issues/19891

aarmey:
Hmm... I'm happy to go with your preference, but think the casting makes more sense this way. I *think* MXNet is often doing this automatically (e.g., Python's `min()` works in RPCA) and the only place where an issue arises is with `list.pop()`.

In TT, the `mask` list could also be implemented as an array, if we made `np.delete` or its equivalent available as an alternative to `list.pop()`.

It seems more surprising that operations would fundamentally change the type of a variable depending upon the shape you end up with. My mental model of Tensorly is that it provides high-level abstractions of operations that mostly remain within the backend.

JeanKossaifi:
I agree - best to let the backend decide as long as it doesn't break things for us. 
Yes, the tt-cross needs some work.

Thanks for the PR, merging!

PR #226, opened 2021-02-19
Sign correction method for CP tensors

aarmey:
`cp_fix_sign` fixes the sign indeterminacy of CP-formatted tensors by forcing all but one factor to be positive. I keep implementing this in different projects, and so figure it would be helpful to include.

### DIFF
--- a/tensorly/cp_tensor.py
+++ b/tensorly/cp_tensor.py
@@ -334,12 +334,16 @@ def cp_fix_sign(cp_tensor, mode=0, func=None):
         func = T.mean
 
     for jj in range(0, len(factors)):
+        # Skip the target mode
+        if jj == mode:
+            continue
+
         # Calculate the sign of the current factor in each component
         means = T.sign(func(factors[jj], axis=0))
 
         # Update both the current and receiving factor
-        factors[mode] *= means[np.newaxis, :]
-        factors[jj] *= means[np.newaxis, :]
+        factors[mode] = factors[mode]*means[np.newaxis, :]
+        factors[jj] = factors[jj]*means[np.newaxis, :]
 
     return CPTensor((weights, factors))
 
--- a/tensorly/cp_tensor.py
+++ b/tensorly/cp_tensor.py
@@ -339,11 +339,11 @@ def cp_fix_sign(cp_tensor, mode=0, func=None):
             continue
 
         # Calculate the sign of the current factor in each component
-        means = T.sign(func(factors[jj], axis=0))
+        column_signs = T.sign(func(factors[jj], axis=0))
 
         # Update both the current and receiving factor
-        factors[mode] = factors[mode]*means[np.newaxis, :]
-        factors[jj] = factors[jj]*means[np.newaxis, :]
+        factors[mode] = factors[mode]*column_signs[np.newaxis, :]
+        factors[jj] = factors[jj]*column_signs[np.newaxis, :]
 
     return CPTensor((weights, factors))
 
--- a/tensorly/cp_tensor.py
+++ b/tensorly/cp_tensor.py
@@ -345,6 +345,11 @@ def cp_flip_sign(cp_tensor, mode=0, func=None):
         factors[mode] = factors[mode]*column_signs[np.newaxis, :]
         factors[jj] = factors[jj]*column_signs[np.newaxis, :]
 
+    # Check the weight signs
+    weight_signs = T.sign(weights)
+    factors[mode] = factors[mode]*weight_signs[np.newaxis, :]
+    weights = T.abs(weights)
+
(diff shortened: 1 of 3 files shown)


aarmey:
@JeanKossaifi should be ready for your review. You might have thoughts about the interface.

tensorly/cp_tensor.py
```diff
@@ -300,6 +300,50 @@ def cp_normalize(cp_tensor):
...
+    if func is None:
+        func = T.mean
+
+    for jj in range(0, len(factors)):
+        # Calculate the sign of the current factor in each component
+        means = T.sign(func(factors[jj], axis=0))
+
+        # Update both the current and receiving factor
+        factors[mode] *= means[np.newaxis, :]
+        factors[jj] *= means[np.newaxis, :]
```
JeanKossaifi, line 342: I wonder if we may want to do factors[mode] = factors[mode]*means[np.newaxis, :] in case there are gradients associated -- doing the operation inplace may break the graph. Perhaps we can add an option `copy=True`? 

As a side note I didn't realize indexing with np.newaxis/None would work with all backends, that's great! :)

tensorly/cp_tensor.py
```diff
@@ -300,6 +300,50 @@ def cp_normalize(cp_tensor):
...
+    -------
+    CPTensor = (normalisation_weights, normalised_factors)
+    """
+    _validate_cp_tensor(cp_tensor)
+    weights, factors = cp_tensor
+
+    if func is None:
+        func = T.mean
+
+    for jj in range(0, len(factors)):
```
JeanKossaifi, line 336: We could avoid the computation on the target mode as it currently just multiplies it by its sign twice
```python
if jj == mode: 
    continue
```

tensorly/cp_tensor.py
```diff
@@ -300,6 +300,50 @@ def cp_normalize(cp_tensor):
...
+        # Calculate the sign of the current factor in each component
+        means = T.sign(func(factors[jj], axis=0))
```
JeanKossaifi, line 338: Super minor but perhaps we can call that `column_aggregated_signs` instead of means just to be more explicit? Or something similar?

JeanKossaifi:
I left a few comments, overall looks great to me. Curious when you need to do this flipping? Do you not mind the sign of the weights of the decomposition? 

Are there other kinds of sign flipping we may want to add down the line @cohenjer ?

tensorly/cp_tensor.py
```diff
@@ -300,6 +300,50 @@ def cp_normalize(cp_tensor):
     return CPTensor((weights, normalized_factors))
 
 
+def cp_fix_sign(cp_tensor, mode=0, func=None):
```
JeanKossaifi, line 303: Maybe `cp_flip_sign` to echo the SVD version? Just a thought, happy to keep the current name.

cohenjer:
@JeanKossaifi As far as I know sign flipping is mostly useful as a post-processing for unconstrained CP; I don't have any other kind of sign flip in mind :) For instance it's not really an issue in (unconstrained) Tucker because the factors should not be interpreted directly (lack of uniqueness).

Also, using the mean to find the sign seems reasonable, but with the interface by @aarmey one can use any function he wants so that's flexible enough for any case I am aware of.

aarmey:
@JeanKossaifi thanks for the input. I've applied your suggestions.

I end up doing this to help with interpretation of the resulting factors. For example, if I'm looking at the results of a patients x cytokines x time tensor, I can ensure the cytokines/time modes are predominantly positive and the patients mode exclusively indicates the sign of the effect. It also helps to ensure positive effects don't appear negative due to two negative signs cancelling each other.

JeanKossaifi:
Awesome, thanks a lot @aarmey, merging!



> I end up doing this to help with interpretation of the resulting factors. 

Is it not an issue if the weights have negative signs (assuming `cp = (weights, factors)` )? Do you also absorb them into the factors? Or are they all one as using CP without normalization?

aarmey:
The weights are set from the norm, so I believe they should always be positive?

cohenjer:
I guess that's only true for weights which are estimated with the parafac function for instance, but not if some user has deliberately used negative weights when creating an instance of a CP tensor.

JeanKossaifi:
Exactly - in most cases they are positive, but in general there might be cases where they could be negative, e.g. from user input or from tensor power iteration.

aarmey:
Ok, I'll add a quick check for the weights right now.

aarmey:
@JeanKossaifi @cohenjer should be handled now.

JeanKossaifi:
Awesome, thanks @aarmey! 😃 
Merging.

PR #229, opened 2021-03-01
Tensor Ring implementation

merajhashemi:
This PR implements Tensor Ring decomposition using recursive SVD; see [Algorithm 1](https://arxiv.org/abs/1606.05535).

### DIFF
--- a/tensorly/decomposition/_tr.py
+++ b/tensorly/decomposition/_tr.py
@@ -10,7 +10,7 @@ def tensor_ring(input_tensor, rank, mode=0, verbose=False):
     Parameters
     ----------
     input_tensor : tensorly.tensor
-    rank : {int, int list}
+    rank : Union[int, List[int]]
             maximum allowable TR rank of the factors
             if int, then this is the same for all the factors
             if int list, then rank[k] is the rank of the kth factor
@@ -45,12 +45,11 @@ def tensor_ring(input_tensor, rank, mode=0, verbose=False):
     unfolding = tl.reshape(input_tensor, (tensor_size[0], -1))
 
     n_row, n_column = unfolding.shape
-    # TODO: Proper warning message
     if rank[0] * rank[1] > min(n_row, n_column):
-        raise ValueError(f'rank[{mode}] * rank[{mode + 1}] = {rank[0] * rank[1]} is larger than'
-                         'first matricization dimensions.\n'
-                         'Failed to compute first factor with specified rank'
-                         'Try with another mode or reduce ranks')
+        raise ValueError(f'rank[{mode}] * rank[{mode + 1}] = {rank[0] * rank[1]} is larger than '
+                         f'first matricization dimension {n_row}×{n_column}.\n'
+                         'Failed to compute first factor with specified rank. '
+                         'Reduce specified ranks or change first matricization `mode`.')
 
     # SVD of unfolding matrix
     U, S, V = tl.partial_svd(unfolding, rank[0] * rank[1])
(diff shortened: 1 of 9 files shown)


tensorly/decomposition/_tr.py
```diff
@@ -0,0 +1,102 @@
...
+    factors = [None] * n_dim
+
+    # Getting the first factor
+    unfolding = tl.reshape(input_tensor, (tensor_size[0], -1))
+
+    n_row, n_column = unfolding.shape
+    # TODO: Proper warning message
+    if rank[0] * rank[1] > min(n_row, n_column):
+        raise ValueError(f'rank[{mode}] * rank[{mode + 1}] = {rank[0] * rank[1]} is larger than'
+                         'first matricization dimensions.\n'
```
JeanKossaifi, line 51: Would be good to have a clearer message, e.g.
The mismatch between rank[mode] and rank[0],

The best is to have a message that points precisely the error to the users and helps to fix it. 
We could for instance say the first unfolding is of size (x, y) so will have rank at most min(x, y) and yet got rank[0]*rank[1]=z > min(x, y).

tensorly/decomposition/_tr.py
```diff
@@ -0,0 +1,102 @@
...
+        Decomposes `input_tensor` into a sequence of order-3 tensors (factors) [1]_.
+
+    Parameters
+    ----------
+    input_tensor : tensorly.tensor
+    rank : {int, int list}
+            maximum allowable TR rank of the factors
+            if int, then this is the same for all the factors
+            if int list, then rank[k] is the rank of the kth factor
+    mode : int, default is 0
```
JeanKossaifi, line 17: When would we want to not start by 0? Does it change much the final decomposition?

tensorly/tr_tensor.py
```diff
@@ -0,0 +1,219 @@
...
+        rank_prev, _, rank_next = factor.shape
+        factor = tl.reshape(factor, (rank_prev, -1))
+        full_tensor = tl.dot(full_tensor, factor)
+        full_tensor = tl.reshape(full_tensor, (-1, rank_next))
+
+    full_tensor = tl.reshape(full_tensor, (factors[-1].shape[2], -1, factors[-1].shape[0]))
+    full_tensor = tl.moveaxis(full_tensor, 0, -1)
+    full_tensor = tl.reshape(full_tensor, (-1, factors[-1].shape[0] * factors[-1].shape[2]))
+    factor = tl.moveaxis(factors[-1], -1, 1)
+    factor = tl.reshape(factor, (-1, full_shape[-1]))
```
JeanKossaifi, line 41: That seems like a lot of reshapes and moveaxis, though I haven't had the time to look if we can avoid some.

JeanKossaifi (commented):
Thanks @merajhashemi, we've been meaning to add tensor ring for a while!

Looks good overall, I would just make a few changes in the decomposition function and add some unit-tests!

JeanKossaifi:
@merajhashemi are you still working on this - that's a great feature to have!

JeanKossaifi:
@merajhashemi looks good to me - I had just left some comments in the code, if you can have a quick look and make any change where appropriate then I think we're good to merge.

tensorly/decomposition/_tr.py
merajhashemi: The authors in the original paper state:
> TR-SVD might obtain different representations by choosing a different mode as the start point. This indicates that TR-ranks is not necessary to be the global optimum in TR-SVD.

Generally, the ranks on the sides are smaller than the ones in middle. Changing the mode allows to shift the sides and to explore different ranks.

tensorly/tr_tensor.py
merajhashemi: I think only line 35 is redundant for the last factor in the for loop, as it is reshaped again in line 37.

JeanKossaifi:
Awesome, thanks @merajhashemi, looks great, merging!

PR #231, opened 2021-03-03
First iteration of entropy metric, test functions, backend updates

taylorpatti:
Basic implementation (matrix and CP only) of vonNeumann entropy and corresponding test functions. Minor update of main 3 backends to accommodate new functions.


taylorpatti:
Looks like an issue with how I added log2 to some of the backends (Numpy, PyTorch, and MXNet). If you know the issue off the top of your head, please let me know. If not, I can troubleshoot it a little later.

tensorly/metrics/entropy.py
```diff
@@ -0,0 +1,34 @@
...
+    tensor : (matrix or CP tensor)
+        Data structure
+    offset : (float)
+        Minimum value at which to not consider eigenvalue zero to avoid singularities.
+
+    Returns
+    -------
+    float
+    """
+    if tensor.__class__.__name__ == 'CPTensor':
```
merajhashemi, line 22:
```python3
if isinstance(tensor, CPTensor):
```

tensorly/metrics/entropy.py
```diff
@@ -0,0 +1,34 @@
...
+        weights = tensor[0]
+        if len(weights) == 1:
+            return 0
+        if np.allclose(weights, np.ones(len(weights))):
+            weights = T.tensor(cp_normalize(tensor)[0])
+        return T.sum(-weights*T.log2(weights))
```
merajhashemi, line 28:
```python3
return -float(T.sum(weights*T.log2(weights)))
```
Some backends return 0-dim tensors

tensorly/metrics/entropy.py
```diff
@@ -0,0 +1,34 @@
...
+
+    tensor = T.tensor(tensor)
+    eig_vals = T.eigh(tensor)[0]
+    eig_vals = eig_vals[eig_vals > offset]
+    return -T.sum(T.log2(eig_vals)*eig_vals)
```
merajhashemi, line 33:
```python3
return - float(T.sum(T.log2(eig_vals)*eig_vals))
```

tensorly/metrics/tests/test_entropy.py
```diff
@@ -0,0 +1,63 @@
...
+       [0.02577351, 0.10426328, 0.06440067, 0.02632028, 0.05930842,
+        0.05479861, 0.04390419, 0.10996953],
+       [0.09599539, 0.12271515, 0.09704334, 0.14030151, 0.03368433,
+        0.04390419, 0.15988065, 0.12634305],
+       [0.07482428, 0.2212482 , 0.14317333, 0.08937212, 0.11477836,
+        0.10996953, 0.12634305, 0.23241005]])
+
+def test_vonNeumann_entropy_pure_state():
+    tl_vne = np.array(vonNeumann_entropy(mat_pure))
+    assert np.allclose(tl_vne, np.array([0]), rtol=1e-03, atol=1e-03)
```
merajhashemi, line 44: We can have random tests like in your previous commit.
```python3
from ...random import random_tensor


def test_vonNeumann_entropy_pure_state():
    state = random_tensor((16, 1))
    state = state / tl.norm(state)
    mat = tl.dot(state, tl.transpose(state))
    tl_vne = vonNeumann_entropy(mat)
    np.testing.assert_almost_equal(tl_vne, 0.)

```

tensorly/metrics/tests/test_entropy.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_vonNeumann_entropy_pure_state_CP():
+    mat = parafac(tl.tensor(mat_pure), rank=1, normalize_factors=True)
+    tl_vne = np.array(vonNeumann_entropy(mat))
+    assert np.allclose(tl_vne, np.array([0]), rtol=1e-03, atol=1e-03)
+
+def test_vonNeumann_entropy_mixed_state():
+    tl_vne = np.array(vonNeumann_entropy(mat_mixed))
+    assert np.allclose(tl_vne, np.array([0.7072100566602336]), rtol=1e-03, atol=1e-03)
```
merajhashemi, line 53: We can use a maximally mixed state as the test case here.
```python3
def test_vonNeumann_entropy_mixed_state():
    # maximally mixed state
    d = 16
    basis, _ = tl.qr(random_tensor((d, d)))  # get orthonormal basis
    mat = (1 / d) * tl.dot(basis, tl.transpose(basis))
    tl_vne = vonNeumann_entropy(mat)
    np.testing.assert_almost_equal(tl_vne, tl.log2(d))
```

tensorly/metrics/entropy.py
merajhashemi: Isn’t the entropy defined for a density matrix though? When will it be a higher-order tensor?

tensorly/backend/__init__.py
```diff
@@ -224,6 +224,7 @@ def inner(*args, **kwargs):
 finfo = dispatch(Backend.finfo)
 index = Backend.index
 index_update = dispatch(Backend.index_update)
+log2 = dispatch(Backend.log2)
```
merajhashemi, line 227: We need to add `log2` to Cupy, JAX, and TensorFlow too.

tensorly/metrics/entropy.py
taylorpatti: We're going to sometimes organize it as a higher order tensor, at which point it will be more efficient to do the calculation in tensor form.

merajhashemi:
```python3
    if isinstance(tensor, CPTensor):
        weights = cp_normalize(tensor).weights
        vn_entropy = -T.sum(weights * T.log2(weights))

    else:
        eig_vals = T.eigh(tensor)[0]
        eig_vals = eig_vals[eig_vals > offset]
        vn_entropy = -T.sum(T.log2(eig_vals) * eig_vals)

    return float(vn_entropy)
```

tensorly/metrics/tests/test_entropy.py
taylorpatti: Jean requested that I hardcore an example so as to have a pre-calculated answer and not import Qutip.

tensorly/metrics/tests/test_entropy.py
taylorpatti: Is there an advantage to changing this?

tensorly/backend/__init__.py
taylorpatti: Yep, indeed. Working on that for sure.

tensorly/metrics/entropy.py
taylorpatti: Good to know, thanks. Under what circumstances?

tensorly/metrics/entropy.py
taylorpatti: How about:

    if isinstance(tensor, CPTensor):
        eig_vals = cp_normalize(tensor).weights

    else:
        eig_vals = T.eigh(tensor)[0]
        eig_vals = eig_vals[eig_vals > offset]

    vn_entropy = -T.sum(T.log2(eig_vals) * eig_vals)

    return float(vn_entropy)

tensorly/metrics/entropy.py
merajhashemi: For Pytorch, JAX, TensorFlow it will return a 0-dim tensor.

tensorly/metrics/tests/test_entropy.py
merajhashemi: We know the entropy is zero for pure states so there's no need for `Qutip` with this example.

tensorly/metrics/tests/test_entropy.py
merajhashemi: It just looks nicer than hardcoding the tensors!

tensorly/metrics/tests/test_entropy.py
taylorpatti: Right, which is why before it was showing it both for QuTip and with zero itself - as an illustration. Jean requested I remove QuTip on all accounts though, which is why I hardcoded both.

tensorly/metrics/entropy.py
merajhashemi:
```python3
eig_vals = eig_vals[eig_vals > offset]
```
How about we keep this line outside the else statement?

taylorpatti: Sounds good. I'm also consolidating the last two lines.

tensorly/metrics/entropy.py
```diff
@@ -0,0 +1,33 @@
...
+    tensor : (matrix or CP tensor)
+        Data structure
+    offset : (float)
+        Minimum value at which to not consider eigenvalue zero to avoid singularities.
+
+    Returns
+    -------
+    float
+    """
+    if isinstance(tensor, CPTensor):
```
JeanKossaifi, line 23: I'm wondering whether it would be worth having an explicit cp_von_neumann_entropy?

tensorly/metrics/entropy.py
```diff
@@ -0,0 +1,33 @@
+from .. import backend as T
+from ..cp_tensor import CPTensor
+from ..cp_tensor import cp_normalize
+import numpy as np
+
+# Author: Taylor Lee Patti <taylorpatti@g.harvard.edu>
+
+
+def vonNeumann_entropy(tensor, offset=1e-12):
```
JeanKossaifi, line 9: I would tend to use von_neumann_entropy for a function and VonNeumannEntropy for a class.

tensorly/metrics/entropy.py
```diff
@@ -0,0 +1,33 @@
...
+    -------
+    float
+    """
+    if isinstance(tensor, CPTensor):
+        eig_vals = cp_normalize(tensor).weights
+
+    else:
+        eig_vals = T.eigh(tensor)[0]
+
+    eig_vals = eig_vals[eig_vals > offset]
```
JeanKossaifi, line 29: This offset will only be valid for float64, we can use the dtype's machine precision instead: `eps = tl.eps(tensor.dtype)`.

tensorly/metrics/entropy.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+    return float(-T.sum(T.log2(eig_vals) * eig_vals))
```
JeanKossaifi, line 32: Not sure we really want to cast to float, this will break the computational graph if we're e.g. tracking gradients.

tensorly/metrics/tests/test_entropy.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
+import numpy.random as npr
+import numpy.linalg as npl
+import tensorly as tl
+from ..entropy import vonNeumann_entropy
+from ...decomposition import parafac
+
+state = npr.rand(8, 1).astype(np.float32)
```
JeanKossaifi, line 8: I would define these inside the unit-test functions. If we no longer cast to float it would also be better to use tl.randn and tl.tensor so the arrays are in created with the right backend.

tensorly/metrics/tests/test_entropy.py
```diff
@@ -0,0 +1,40 @@
...
+state = state/npl.norm(state)
+mat_pure = tl.tensor(np.matmul(state, np.transpose(state)))
```
JeanKossaifi, line 10:
```suggestion
mat_pure = tl.dot(state, tl.transpose(state))
```

tensorly/metrics/tests/test_entropy.py
JeanKossaifi, line 8:
```suggestion
state = tl.tensor(npr.rand(8, 1), dtype=tl.float32)
```

tensorly/metrics/tests/test_entropy.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
+import numpy.random as npr
+import numpy.linalg as npl
+import tensorly as tl
+from ..entropy import vonNeumann_entropy
+from ...decomposition import parafac
+
+state = npr.rand(8, 1).astype(np.float32)
+state = state/npl.norm(state)
```
JeanKossaifi, line 9:
```suggestion
state = state/tl.norm(state)
```

tensorly/metrics/tests/test_entropy.py
JeanKossaifi: We can also have both: a random test + deterministic one.
If all are tl.tensors, we can test using `tensorly.testing`, e.g. `tensorly.testing.assert_array_almost_equal`.

JeanKossaifi (changes requested):
Thanks @taylorpatti for the PR and @merajhashemi for the comments!

Looks good, I'd just make it fully backend agnostic - in general it's best to not convert to numpy whenever possible as this is the most flexible for end users.

tensorly/metrics/entropy.py
taylorpatti: Yeah, I disagreed with this as well.

tensorly/metrics/entropy.py
taylorpatti: So you would advise against autodetect for users? Or do you mean just define the main function in terms of the general, cp, and soon tt version? The later would lead to more repetitious code but

JeanKossaifi: Right -- the issue is that now the API for the class is not fixed and mostly experimental (something we need to work on). So far I have a function cp_do_something for most operations that users can call explicitly, though I'm not sure it's the best way either. 

In general now, all cp_tensors are in the form (weights, factors) so the cp_von_neumann_entropy could be:
```
def cp_neumann_entropy(cp_tensor):
    weights, factors = cp_normalize(cp_tensor)
    ...
```
which also will allow tuples, not just CPTensors

As a minor side note, would it make sense to name it just `neumann_entropy`?

taylorpatti: Okay, I get you, for consistency. I'll separate them then.

How about cp_vonneumann_entropy? I think the "von" makes the function's purpose clear - to my background at least.

JeanKossaifi: I agree, let's go with `vonneumann_entropy`

tensorly/metrics/entropy.py
taylorpatti: It seems that CPTensor might not have a dtype attribute? Should I add one?

merajhashemi: The native backend tensors have a dtype attribute. we can do something like:
```python3
eps = T.eps(eig_vals.dtpe)
```

tensorly/metrics/entropy.py
merajhashemi: I suggested this to be consistent with the docstring. We can say the function returns a scalar tensor instead of float:?

tensorly/metrics/entropy.py
```diff
@@ -0,0 +1,89 @@
...
+    """Returns the von Neumann entropy of the tensor.
+
+    Parameters
+    ----------
+    tensor : (matrix)
+        Data structure
+
+    Returns
+    -------
+    float
```
JeanKossaifi, line 19:
```suggestion
    von_neumann_entropy : order-0 tensor
```

JeanKossaifi:
Awesome, thanks a lot @taylorpatti ! Merging.

PR #234, opened 2021-03-05
More simplifications using common numpy interface

### LINKED ISSUE #232: Update the pytorch interface to use the numpy interface
With version 1.8, pytorch now has a numpy-like interface. Specifically, qr and svd definitely need to be moved over.

aarmey:
Fixes #232 

This uses the new torch.linalg interface.

### DIFF
--- a/tensorly/backend/core.py
+++ b/tensorly/backend/core.py
@@ -710,8 +710,7 @@ class Backend(object):
         """
         raise NotImplementedError
 
-    @staticmethod
-    def kron(a, b):
+    def kron(self, a, b):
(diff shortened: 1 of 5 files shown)


aarmey:
@JeanKossaifi this should be ready to review. The only issue is I'm having a problem with `moveaxis` and the sparse backend. As far as I can tell it's provided everywhere, so I'm not sure why it's hitting the not-implemented exception...

aarmey:
Might have fixed the sparse backend problem.

JeanKossaifi:
Awesome, merging, thanks @aarmey!

PR #240, opened 2021-03-09
Fix partial_svd issue when n_eigenvecs is None

### LINKED ISSUE #233: Bug when n_eigenvecs is None in truncated_svd
#### Describe the bug
Hello, I have an issue with truncated_svd function. After the last commit, `partial_svd` function in `backend/core.py` calls `truncated_svd` function in line 836 `if (n_eigenvecs is None)`. However, there is a type error in line 901 because '<' is not supported between instances of 'int ' and NoneType. 

Maybe, we can take into account None for n_eigenvecs in line 901 as;

```python
if (n_eigenvecs is not None) and (max_dim < n_eigenvecs):
```

merajhashemi:
This PR resolves #233.

### DIFF
--- a/tensorly/backend/core.py
+++ b/tensorly/backend/core.py
@@ -831,7 +831,10 @@ class Backend(object):
         dim_1, dim_2 = self.shape(matrix)
         min_dim, max_dim = min(dim_1, dim_2), max(dim_1, dim_2)
 
-        if (n_eigenvecs is None) or (min_dim <= n_eigenvecs):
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
Thanks @merajhashemi, this is fixed in main:

https://github.com/tensorly/tensorly/blob/05ecbefa75cbaa1ab0ed1f85443af37760a427c1/tensorly/backend/core.py#L838-L840

PR #247, opened 2021-03-22
Fix parafac for complex tensors

### LINKED ISSUE #246: Complex Support
Here are 3 issues with complex support:

```python
import math
import tensorly as tl
from tensorly import testing
from tensorly.decomposition import parafac, tensor_train
```

1.) Issue with "norm" from backend

```python
def test_norm_complex():
    """The norm at a minimum does not work for complex tensors with order = 2, as
    it should be tensor*conj(tensor), not tensor**2. I'm not sure about the other
    orders, they too should be checked."""
    tensor = tl.tensor([1, 1j, 0], dtype=torch.complex64)
    norm = tl.norm(tensor)
    testing.assert_array_almost_equal(norm, math.sqrt(2), decimal=6)
```

2.) Issue with CP tensor decomposition/recomposition with parafac and cp_to_tensor - fails in general by giving very wrong answer

```python
def test_cp_complex():
    """Test that parafac decomposition and cp_to_tensor work for complex tensors.
    Produces incorrect answers. Code needs debugging for complex tensors."""
    shape = (2, 4, 3, 2)
    tensor = tl.randn(shape) + tl.randn(shape)*1j
    decomp = parafac(tensor, rank=100)
    recomp = tl.cp_to_tensor(decomp)
    testing.assert_array_almost_equal(tensor, recomp)
```

3.) Issue with TT tensor decomposition/recomposition with tensor_train and tt_to_tensor - fails for TensorFlow and MxNet, but as it still works for 3 backend

```python
def test_tt_complex():
    """Test that tensor_train decomposition and tt_to_tensor work for complex tensors.
    Works for some backends, but not for TensorFlow and MxNet."""
    shape = (2, 4, 3, 2)
    tensor = tl.randn(shape) + tl.randn(shape)*1j
    decomp = tensor_train(tensor, rank=(1, 8, 24, 8, 1))
    recomp = tl.tt_to_tensor(decomp)
    testing.assert_array_almost_equal(tensor, recomp)
```

merajhashemi:
This pr fixes #246.

### DIFF
--- a/tensorly/cp_tensor.py
+++ b/tensorly/cp_tensor.py
@@ -392,7 +392,7 @@ def cp_to_tensor(cp_tensor, mask=None):
 
     if mask is None:
         full_tensor = T.dot(factors[0]*weights,
-                             T.conj(T.transpose(khatri_rao(factors, skip_matrix=0))))
+                            T.transpose(khatri_rao(factors, skip_matrix=0)))
     else:
         full_tensor = T.sum(khatri_rao([factors[0]*weights]+factors[1:], mask=mask), axis=1)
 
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
Thanks @merajhashemi !

PR #254, opened 2021-04-06
Non-negative tucker with HALS

caglayantuna:
# HALS for Non-negative Tucker
This pull request adds a non-negative tucker with HALS algorithm to `_tucker.py` file. In our previous PR, we added `hals_nnls` function and here we use it for tucker decomposition. We added `non_negative_tucker_hals` and `initialize_tucker` function to `_tucker.py` file. Additionally, we added two nonnegative least square algorithms to proximal.py in order to update core for nonnegative tucker decomposition.

# Motivation
Using HALS for non-negative Tucker decomposition gives more accurate results according to our experiments and the literature. We compared our algorithms with the current nonnegative Tucker in Tensorly and we observed pretty good improvement. We are planning to submit an example based PR soon which will include our experiments.

# Source Modification 
* **decomposition/_tucker.py:**

First, we added `initialize_tucker` function in order to have initializing function such as `initialize_cp` for Tucker. In this function, if non_negative is True, it returns non-negative core and factors. Otherwise, it can be used for unconstrained `Tucker` decomposition as well.

We added `non_negative_tucker_hals` function which uses `hals_nnls` algorithm to calculate factors for tucker decomposition.

To compare our function, we added return_error option to `non_negative_tucker`. We also include this version in this PR.

* **tenalg/_proximal.py:**

To update core during the iteration in `non_negative_tucker_hals`, we added two fast gradient algorithms which are active set and Fast Iterative Shrinkage Thresholding Algorithm (FISTA). These functions return updated nonnegative core for `non_negative_tucker_hals`. Besides, these functions can be used for different nonnegative least square problems in the future.

* **test_tucker.py:**

Lastly, we have added a test function *"test_non_negative_tucker_hals()"* for *"non_negative_tucker_hals"* function. Simply, we have copied *"test_non_negative_tucker"* and changed related function. We have tested our function with this test function as well.

# References

1-  Bro, R., & De Jong, S. (1997). A fast non‐negativity‐constrained least squares algorithm. Journal of Chemometrics: A Journal of the Chemometrics Society, 11(5), 393-401. [Link](http://xrm.phys.northwestern.edu/research/pdf_papers/1997/bro_chemometrics_1997.pdf)

2- Beck, A., & Teboulle, M. (2009). A fast iterative shrinkage-thresholding algorithm for linear inverse problems. SIAM journal on imaging sciences, 2(1), 183-202.  [Link](https://epubs.siam.org/doi/abs/10.1137/080716542)

### DIFF
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -4,7 +4,7 @@ from ..base import unfold
 from ..tenalg import multi_mode_dot, mode_dot	
 from ..tucker_tensor import tucker_to_tensor, TuckerTensor, validate_tucker_rank
 import tensorly.tenalg as tlg
-from ..tenalg.proximal import hals_nnls
+from ..tenalg.proximal import hals_nnls, active_set_nnls, fista_nnls
 from math import sqrt
 import warnings
 
@@ -67,7 +67,7 @@ def initialize_tucker(tensor, rank, modes, random_state, init='svd', svd='numpy_
         nn_core = tl.abs(core) 
     return nn_core, nn_factors
 
-def update_core_hals(tensor_cross, core, factors, pseudo_inverse, sparsity_coefficients, n_iter_max=100):
+def update_core_hals(tensor, core, factors, pseudo_inverse, sparsity_coefficients, algorithm='fpg', n_iter_max=100):
     """
     Updating core for non negative tucker hals function
     
@@ -85,37 +85,29 @@ def update_core_hals(tensor_cross, core, factors, pseudo_inverse, sparsity_coeff
     core    : ndarray
               updated core tensor
     """
-    all_MtX = tl.tenalg.mode_dot(tensor_cross, tl.transpose(factors[-1]), len(factors)-1)
     pseudo_inverse[-1] = tl.dot(tl.transpose(factors[-1]),factors[-1])#all_MtM
-    
     # Projected gradient
-    #Partial svd?
     gradient_step = 1
     for MtM in pseudo_inverse:
-       #gradient_step *= 1/(tl.partial_svd(MtM)[1][0])
-       gradient_step *= 1/(scipy.sparse.linalg.svds(MtM, k=1)[1][0])
+       gradient_step *= 1/(tl.partial_svd(MtM)[1][0])
 
-    if sparsity_coefficients[-1] is None:
-        sparse = 0
-    else:
-        sparse = sparsity_coefficients[-1]
-    
-    norm_0 = 0
-    update = 0
-    # Maybe: try fast gradient instead of gradient
-    for iteration in range(n_iter_max):
+    #gradient_step = round(gradient_step, 6) # Heurisitc, to avoid consecutive imprecision
+
+
         
-        # Nesterov Fast Gradient
-        core_upd= core-0.9*update
-        gradient = - all_MtX + tl.tenalg.multi_mode_dot(core_upd, pseudo_inverse, transpose=False) + sparse
-        update= 0.9*update + tl.where(gradient_step*gradient < core_upd, gradient_step*gradient, core_upd)
-        core = core - update
-        norm = tl.norm(update)
-        
-        if iteration == 1:
-             norm_0 = norm
-        if norm < 0.01 * norm_0:
-             break
+    # Fast Iterative Shrinkage-Thresholding Algorithm (FISTA)
+    if algorithm == 'fista':
+          core = fista(tensor, factors, x=core, n_iter_max=n_iter_max, gradient_step=gradient_step,                                                    sparsity_coefficient=sparsity_coefficients)
+
+    # Active set algorithm
+    if algorithm == 'as':
+          core = active_set(tensor, factors, x=core, n_iter_max=n_iter_max)  
+
+    if algorithm == 'nnls':
+        tensorvec = tl.base.tensor_to_vec(tensor)
+        factor_kr = tl.tenalg.kronecker(factors)
+        vectorcore, norm = scipy.optimize.nnls(factor_kr, tensorvec)
+        core = tl.reshape(vectorcore, tl.shape(core))
     return core
     
 def partial_tucker(tensor, modes, rank=None, n_iter_max=100, init='svd', tol=10e-5,
@@ -312,7 +304,7 @@ def tucker(tensor, rank, fixed_factors=None, n_iter_max=100, init='svd',
         return TuckerTensor((core, factors))
 
 def non_negative_tucker(tensor, rank, n_iter_max=10, init='svd', tol=10e-5,
-                        random_state=None, verbose=False):
+                        random_state=None, verbose=False, return_errors=False):
     """Non-negative Tucker decomposition
 
         Iterative multiplicative update, see [2]_
@@ -399,7 +391,11 @@ def non_negative_tucker(tensor, rank, n_iter_max=10, init='svd', tol=10e-5,
                 print('converged in {} iterations.'.format(iteration))
             break
 
-    return TuckerTensor((nn_core, nn_factors))
+    tensor = TuckerTensor((nn_core, nn_factors))
+    if return_errors:
+        return tensor, rec_errors
+    else:
+        return tensor
     
 def non_negative_tucker_hals(tensor, rank, n_iter_max=100, init="svd", svd='numpy_svd', tol=1e-8,
                              sparsity_coefficients=None, fixed_modes=None, random_state=None,
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -7,6 +7,7 @@ import tensorly.tenalg as tlg
 from ..tenalg.proximal import hals_nnls, active_set_nnls, fista
 from math import sqrt
 import warnings
+from tensorly.decomposition._nn_cp import make_svd_non_negative
 
 # Author: Jean Kossaifi <jean.kossaifi+tensors@gmail.com>
 
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -452,7 +452,7 @@ def non_negative_tucker_hals(tensor, rank, n_iter_max=100, init="svd", svd='nump
             UtM = tl.transpose(MtU)
 
             # Call the hals resolution with nnls, optimizing the current mode
-            nn_factor, _, _, _ = hals_nnls(UtM, UtU, tl.transpose(factors[mode]),
+            nn_factor, _, _, _ = hals_nnls(UtM, UtU, tl.transpose(nn_factors[mode]),
                                            n_iter_max=100, sparsity_coefficient=sparsity_coefficients[mode],
                                            exact=exact)
             nn_factors[mode] = tl.transpose(nn_factor)
@@ -470,7 +470,6 @@ def non_negative_tucker_hals(tensor, rank, n_iter_max=100, init="svd", svd='nump
             pseudo_inverse[-1] = tl.dot(tl.transpose(nn_factors[-1]), nn_factors[-1])
             AtB = tl.base.tensor_to_vec(tl.tenalg.mode_dot(tensor_cross, tl.transpose(nn_factors[modes[-1]]), modes[-1]))
             AtA = tl.tenalg.kronecker(pseudo_inverse)
-            nn_core = active_set_nnls(AtA, AtB, x=nn_core, n_iter_max=n_iter_max)
             vectorcore = active_set_nnls(AtA, AtB, x=nn_core, n_iter_max=n_iter_max)
             nn_core = tl.reshape(vectorcore, tl.shape(nn_core))
         
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -74,6 +74,7 @@ def initialize_tucker(tensor, rank, modes, random_state, init='svd', svd='numpy_
  
     return core, factors
     
+
 def partial_tucker(tensor, modes, rank=None, n_iter_max=100, init='svd', tol=10e-5,
                    svd='numpy_svd', random_state=None, verbose=False, mask=None):
     """Partial tucker decomposition via Higher Order Orthogonal Iteration (HOI)
@@ -419,26 +420,26 @@ def non_negative_tucker_hals(tensor, rank, n_iter_max=100, init="svd", svd='nump
     -----
     Tucker decomposes a tensor into a core tensor and list of factors:
     .. math::
-        \begin{equation}
+        \\begin{equation}
             tensor = [| core; factors[0], ... ,factors[-1] |]
-        \end{equation}
+        \\end{equation}
 
     We solve the following problem for each factor:
     .. math::
-        \begin{equation}
-            \min_{tensor >= 0} ||tensor_[i] - factors[i]\times core_[i] \times (\prod_{i\neq j}(factors[j]))^T||^2
-        \end{equation}
+        \\begin{equation}
+            \\min_{tensor >= 0} ||tensor_[i] - factors[i]\\times core_[i] \\times (\\prod_{i\\neq j}(factors[j]))^T||^2
+        \\end{equation}
 
     If we define two variables such as:
     .. math::
-            U = core_[i] \times (\prod_{i\neq j}(factors[j]))^T \\
+            U = core_[i] \\times (\\prod_{i\\neq j}(factors[j]))^T \\
             M = tensor_[i]
 
     Gradient of the problem becomes:
         .. math::
-        \begin{equation}
-            \delta = -U^TM + factors[i] \times U^TU
-        \end{equation}
+        \\begin{equation}
+            \\delta = -U^TM + factors[i] \\times U^TU
+        \\end{equation}
 
     References
     ----------
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -4,7 +4,7 @@ from ..base import unfold
 from ..tenalg import multi_mode_dot, mode_dot	
 from ..tucker_tensor import tucker_to_tensor, TuckerTensor, validate_tucker_rank
 import tensorly.tenalg as tlg
-from ..tenalg.proximal import hals_nnls, active_set_nnls, fista_nnls
+from ..tenalg.proximal import hals_nnls, active_set_nnls, fista
 from math import sqrt
 import warnings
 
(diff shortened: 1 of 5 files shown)


tensorly/decomposition/_tucker.py
```diff
@@ -294,7 +349,156 @@ def non_negative_tucker(tensor, rank, n_iter_max=10, init='svd', tol=10e-5,
...
+                if i != mode:
+                    pseudo_inverse[i] = tl.dot(tl.conj(tl.transpose(factor)), factor)
+            # Second, the multiway product with core G
+            core_cross = multi_mode_dot(nn_core, pseudo_inverse, skip=mode)
+            UtU = tl.dot(unfold(core_cross, mode), tl.transpose(unfold(nn_core, mode)))
+
+            # UtM
+            tensor_cross = multi_mode_dot(tensor, nn_factors, skip=mode, transpose=True)
+            MtU = tl.dot(unfold(tensor_cross, mode), tl.transpose(unfold(nn_core, mode)))
+            UtM = tl.transpose(MtU)
```
JeanKossaifi, line 454: If we use MtU, UtM, UtU, it would be nice to have a short Notes section in the docstring to briefly introduce the notation. We can also add a short mathematical bit in the user guide.

tensorly/decomposition/_tucker.py
```diff
@@ -294,7 +349,156 @@ def non_negative_tucker(tensor, rank, n_iter_max=10, init='svd', tol=10e-5,
...
+        reconstruction errors or not
+        Default: False
+    return_errors : boolean
+        Indicates whether the algorithm should return all reconstruction errors
+        and computation time of each iteration or not
+        Default: False
+    exact : If it is True, the HALS nnls subroutines give results with high precision but it needs high computational cost. 
+        If it is False, the algorithm gives an approximate solution.
+        Default: False
+    algorithm : {'fista', 'as'}
```
JeanKossaifi, line 401: I would explicitly use `'active_set'` instead of just `'as'`

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+
+    Computes and approximate solution for Ax=b linear system.
+
+    Parameters
+    ----------
+    AtB: ndarray
+       Pre-computed product of the transposed of A and B.
+    pseudo_inverse: ndarray
+       Pre-computed product of the transposed of A and A.
+    x: initialized array
```
JeanKossaifi, line 266: shall we call that `init`?

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+def fista(AtB, pseudo_inverse, x=None, n_iter_max=100, non_negative=True, gradient_step=None,
+          sparsity_coefficient=None):
+    """
+    Fast Iterative Shrinkage Thresholding Algorithm (FISTA)
+
+    Computes and approximate solution for Ax=b linear system.
+
+    Parameters
+    ----------
+    AtB: ndarray
```
JeanKossaifi, line 262: Should that be Atb or ATb?

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+          SIAM journal on imaging sciences, 2(1), 183-202.
+    """
+
+    if sparsity_coefficient[-1] is None:
+        sparse = 0
+    else:
+        sparse = sparsity_coefficient[-1]
+
+    if gradient_step is None:
+        gradient_step = 0.001
```
JeanKossaifi, line 293: This should just be a default parameter.

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+
+        xnew = x_upd - delta_x
+        momentum = (1 + tl.sqrt(1 + 4 * momentum_old ** 2)) / 2
+        x_upd = xnew + ((momentum_old - 1) / momentum) * (xnew - x)
+        momentum_old = momentum
+        x = tl.copy(xnew)
+        norm = tl.norm(delta_x)
+        if iteration == 1:
+            norm_0 = norm
+        if norm < 0.01 * norm_0:
```
JeanKossaifi, line 319: should we add a parameter `tol=0.01` ?

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+    x : Updated ndarray
+
+    Reference
+    ----------
+    [1] : Beck, A., & Teboulle, M. (2009). A fast iterative
+          shrinkage-thresholding algorithm for linear inverse problems.
+          SIAM journal on imaging sciences, 2(1), 183-202.
+    """
+
+    if sparsity_coefficient[-1] is None:
```
JeanKossaifi, line 287: why is sparsity_coefficient assumed to be a list here? 
Should we just use `sparsity_coef` to replace both sparsity_coefficient and sparse and have `sparsity_coef = 0` in the signature?

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+        Maximum number of iteration
+        Default: 100
+    non_negative : bool, default is False
+                   if True, result will be non-negative
+    gradient_step : float
+    sparsity_coefficient : float or None
+
+    Returns
+    -------
+    x : Updated ndarray
```
JeanKossaifi, line 278:
```suggestion
    x : approximate solution such that Ax = b
```

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+    if gradient_step is None:
+        gradient_step = 0.001
+
+    if x is None:
+        x = tl.zeros([tl.shape(pseudo_inverse)[0], tl.shape(AtB)[1]])
+
+    # Parameters
+    momentum_old = tl.tensor(1.0)
+    norm_0 = 0.0
+    x_upd = tl.copy(x)
```
JeanKossaifi, line 301: We may as well spell it out, e.g. x_update

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+
+    for iteration in range(n_iter_max):
+        gradient = - AtB + tl.tenalg.multi_mode_dot(x_upd, pseudo_inverse, transpose=False) + sparse
+
+        if non_negative is True:
+            delta_x = tl.where(gradient_step * gradient < x, gradient_step * gradient, x_upd)
+        else:
+            delta_x = gradient_step * gradient
+
+        xnew = x_upd - delta_x
```
JeanKossaifi, line 311:
```suggestion
        x_new = x_update - gradient_step*x_gradient
```

nitpick: x_new to be consistent with rest?
I would also remove the gradient step from above and make it here for clarity.

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+    # Parameters
+    momentum_old = tl.tensor(1.0)
+    norm_0 = 0.0
+    x_upd = tl.copy(x)
+
+    for iteration in range(n_iter_max):
+        gradient = - AtB + tl.tenalg.multi_mode_dot(x_upd, pseudo_inverse, transpose=False) + sparse
+
+        if non_negative is True:
+            delta_x = tl.where(gradient_step * gradient < x, gradient_step * gradient, x_upd)
```
JeanKossaifi, line 307: Why not update gradient instead of introducing delta_x? We no longer need gradient after this. 
Generally it's good to introduce as few variables as needed to limit memory use and improve readability.

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+    if x is None:
+        x = tl.zeros([tl.shape(pseudo_inverse)[0], tl.shape(AtB)[1]])
+
+    # Parameters
+    momentum_old = tl.tensor(1.0)
+    norm_0 = 0.0
+    x_upd = tl.copy(x)
+
+    for iteration in range(n_iter_max):
+        gradient = - AtB + tl.tenalg.multi_mode_dot(x_upd, pseudo_inverse, transpose=False) + sparse
```
JeanKossaifi, line 304: x_gradient?

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+    pseudo_inverse: ndarray
+       Pre-computed product of the transposed of A and A.
+    x: initialized array
+       Default: None
+    n_iter_max : int
+        Maximum number of iteration
+        Default: 100
+    non_negative : bool, default is False
+                   if True, result will be non-negative
+    gradient_step : float
```
JeanKossaifi, line 273: we could call this learning_rate or lr for consistency with opt. frameworks?

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+    if x is None:
+        x_vec = tl.zeros(tl.shape(AtA)[1])
+    else:
+        x_vec = tl.base.tensor_to_vec(x)
+
+    first = AtB
+    second = tl.dot(AtA, x_vec)
+    w = first - second
+    P = x_vec > 0
+    R = x_vec <= 0
```
JeanKossaifi, line 362: Could we find an explicit name for P and R (e.g. positive_set ...)? I would remove first and second and directly use their value, and rename w to something more explicit as well.

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+    s = tl.zeros(tl.shape(x_vec), **tl.context(x_vec))
+
+    for iteration in range(n_iter_max):
+        # Early stopping criterion
+        if True in R and tl.max(w[R]) <= tol:
```
JeanKossaifi, line 367: I think we can add an `any` function to the backend and use `if tl.any(R)`?

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+                        s = tl.index_update(s, tl.index[int(i)], 0)
+
+                if True not in P or tl.min(s[P]) > 0:
+                    break
+        # set x to s
+        x_vec = tl.clip(s, 0, tl.max(s))
+
+        # gradient update
+        second = tl.dot(AtA, x_vec)
+        w = first - second
```
JeanKossaifi, line 449: same as above, no need to introduce second as it's not used anywhere else, just 
`AtB - tl.dot(AtA, x_vec)`

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+                    c = tl.solve(a, b)
+                    indice_list = []
+                    for i in range(tl.shape(s)[0]):
+                        if P[i]:
+                            indice_list.append(i)
+                            x_vec = tl.index_update(x_vec, tl.index[int(i)], c[len(indice_list) - 1])
+                        else:
+                            x_vec = tl.index_update(x_vec, tl.index[int(i)], 0)
+                    x_vec = tl.clip(x_vec, 0, tl.max(x_vec))
+                except:
```
JeanKossaifi, line 382: What errors are we catching? We should add a specific exception here and not catch everything (e.g. user may be sending `KeyboardInterrupt`.

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+                    x_vec = tl.copy(x_vec)
+            break
+
+        if True in R:
+            indice = tl.argmax(w)
+            P = tl.index_update(P, tl.index[indice], True)
+            R = tl.index_update(R, tl.index[indice], False)
+        # To avoid singularity error when initial x exists
```
JeanKossaifi, line 390: Is that a single occurrence, can we test for that?

JeanKossaifi (changes requested):
Thanks @caglayantuna, another great PR!

It would be great to have a pass focusing on making the code easy to read with more explicit (and where possible, less) variables, particularly for the active set solver. 
Would be good to have a small unit-test for the solvers as well.

tensorly/tenalg/proximal.py
caglayantuna: When initial x exists, we may have this error. According to the my tests, it occurs when all the P (changed, now passive set) is True. So, it happens once because when P has only true, R will have only false then iteration stops in the end.

tensorly/tenalg/proximal.py
caglayantuna: It is numpy.linalg.LinAlgError. I dont know which exception covers that error.

tensorly/tenalg/proximal.py
caglayantuna: There is no second or first anymore

tensorly/tenalg/proximal.py
caglayantuna: yes, it could be better

tensorly/tenalg/proximal.py
caglayantuna: Now, their names are changed to passive_set and active_set.

tensorly/tenalg/proximal.py
caglayantuna: Done

tensorly/tenalg/proximal.py
caglayantuna: Done

tensorly/tenalg/proximal.py
caglayantuna: Now, it is in the signature as a float

tensorly/tenalg/proximal.py
caglayantuna: Ok, it is x_update now

tensorly/decomposition/_tucker.py
caglayantuna: It is 'active_set' now.

tensorly/tenalg/proximal.py
caglayantuna: The idea was to handle non_negativity easily. But as you said, it is possible to do that without it. There is no delta_x now.

tensorly/tenalg/proximal.py
caglayantuna: Now it is "lr"

tensorly/tenalg/proximal.py
cohenjer: It's basically when the linear system is ill-conditionned or has infinitely many solutions. Not sure how to catch this either, some idea?

tensorly/decomposition/_tucker.py
```diff
@@ -1,18 +1,73 @@
...
+        # The initial core approximation is needed here for the masking step
+        core = multi_mode_dot(tensor, factors, modes=modes, transpose=True)
+    elif init == 'random':
+        rng = tl.check_random_state(random_state)
+        core = tl.tensor(rng.random_sample(rank) + 0.01, **tl.context(tensor))  # Check this
+        factors = [tl.tensor(rng.random_sample(s), **tl.context(tensor)) for s in zip(tl.shape(tensor), rank)]
+    else:
+        (core, factors) = init
+ 
+    if non_negative is True:
```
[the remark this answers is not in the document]
caglayantuna: Now, make_svd_non_negative is used to initialize. Should we add this function to init?

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,207 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+        if iteration == 1:
+            norm_0 = norm
+        if norm < 0.01 * norm_0:
+            break
+    return x
+
+
+def active_set_nnls(AtA, AtB, x=None, n_iter_max=100, tol=10e-8):
+    """
+     Active set algorithm for non-negative least square solution.
```
JeanKossaifi, line 324: Can we also describe briefly the problem solved - otherwise the AtA and AtB, combined with x being None by default may be opaque for users (though admittedly they are not that likely to use this function directly).

tensorly/decomposition/_tucker.py
```diff
@@ -294,7 +349,156 @@ def non_negative_tucker(tensor, rank, n_iter_max=10, init='svd', tol=10e-5,
...
+
+            # Call the hals resolution with nnls, optimizing the current mode
+            nn_factor, _, _, _ = hals_nnls(UtM, UtU, tl.transpose(nn_factors[mode]),
+                                           n_iter_max=100, sparsity_coefficient=sparsity_coefficients[mode],
+                                           exact=exact)
+            nn_factors[mode] = tl.transpose(nn_factor)
+        # updating core
+        if algorithm == 'fista':
+            pseudo_inverse[-1] = tl.dot(tl.transpose(nn_factors[-1]), nn_factors[-1])  # all_MtM
+            AtB = multi_mode_dot(tensor, nn_factors, transpose=True)
```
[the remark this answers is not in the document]
caglayantuna: hey, names are clear in the current version. I am using only UTM and UtU now. For instance, AtB is core_estimation

tensorly/decomposition/_tucker.py
caglayantuna: I added mathematical notes to non_negative_tucker_hals function. I think these variables are mote clear now.

tensorly/tenalg/proximal.py
caglayantuna: Now, this variable is UtM. Since M is matrix here, I kept as capital.

caglayantuna:
Hey @JeanKossaifi, thanks for your review. I updated the PR according to your comments. You could find my answers to your comments. 

I have added unit tests for each solver however I have a problem about tensor_to_vec function in test_proximal file. I couldn't understand why it returns error in there although I use the same function elsewhere without any error. Also, I think these solver tests are not perfect because they are not passing the test always. If you have any suggestion, it would be perfect.

JeanKossaifi:
Thanks @caglayantuna, this looks good!
Re. the test failure, you should use directly `tl.tensor_to_vec` (`base` is not automatically imported, if you want to use tensorly.base you have to import it first). 

I've also added `any` to the backend so we can use it now!

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,196 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+
+    for iteration in range(n_iter_max):
+
+        if iteration > 0 or tl.all(x_vec == 0):
+            indice = tl.argmax(x_gradient)
+            passive_set = tl.index_update(passive_set, tl.index[indice], True)
+            active_set = tl.index_update(active_set, tl.index[indice], False)
+        # To avoid singularity error when initial x exists
+        try:
+            a = tl.tensor(tl.to_numpy(UtU)[passive_set, :][:, passive_set], **tl.context(UtU))
```
JeanKossaifi, line 385: I've been debating about this and trying to make it work - the main issue is TensorFlow: I am thinking to either not support tensorflow here or checking for the backend and having a separate case for it. What do you think @caglayantuna @cohenjer 

More generally I'm more and more inclined to phase out tensorflow support (we have jax anyway).

tensorly/tenalg/proximal.py
```diff
@@ -248,3 +248,196 @@ def hals_nnls(UtM, UtU, V=None, n_iter_max=500, tol=10e-8,
...
+                for i in range(tl.shape(s)[0]):
+                    if passive_set[i]:
+                        indice_list.append(i)
+                        s = tl.index_update(s, tl.index[int(i)], c[len(indice_list) - 1])
+                    else:
+                        s = tl.index_update(s, tl.index[int(i)], 0)
+
+                if True not in passive_set or tl.min(s[passive_set]) > 0:
+                    break
+        # set x to s
```
JeanKossaifi, line 434: Could we also find a more explicit name for `s` (and `c` and `a`?) 
The other names we now use make the code much easier to read (passive/active_set, x_gradient, etc)!

caglayantuna: I changed c to passive_solution, s to support_vec and removed a by merging that calculation with the next line.

tensorly/tenalg/proximal.py
caglayantuna: Yes I agree, it is difficult to handle tensorflow for some situations. I changed the codes as you suggested. As a suggestion, seperate case could be directing active set to fista with an informative warning if backend is tensorflow.

caglayantuna:
Hey @JeanKossaifi , thanks for "any" function, it is better for Tensorly. I realized the problem with tensor_to_vec function. In test_proximal file, tensorly is imported from backend as T. I think either we should add tensor_to_vec to backend init file or we should import tensorly as usual. 

Additionally, I added exact definition to docstring of hals because it was missing. Besides, I have changed the parameters of hals when exact is true in order to pass test.

tensorly/tenalg/proximal.py
cohenjer: It does seem like the Tensorflow backend is causing us a lot of problems. Not sure if people are actually using it? Maybe a quick survey on this backend usage would help checking if it is worth the effort to support it at the moment?

JeanKossaifi: Yes, I'll create a survey. @caglayantuna that sounds like a good solution at least for now - we can just make it work for backends other than tensorflow and at the beginning of the nn_tucker, if method is 'fista' then check for the backend and return an error if is tensorflow.

tensorly/tenalg/proximal.py
JeanKossaifi: So we could test with tl.any?

JeanKossaifi:
Looks great overall - just wondering whether we can find a way to remove the try/except. 
We also need to add this to the doc. Would be great to have a small example to e.g. compare regular decomposition vs NN multiplicative updates vs NN via HALS!

caglayantuna:
Thanks for your comment. We were discussing with Jeremy today and we think that there is no any other solution than try except for active set solution. Because, there could be a singularity error depending on the initial x especially when higher ranks are selected. So, I can't define a specific reason to avoid that. Related to example, I have almost finished preparing notebooks for both cp and tucker nn. [Link for notebooks](https://github.com/caglayantuna/tensorly-notebooks/tree/non-negative/02_tensor_decomposition). I will submit these examples after this tucker HALS PR merged.

JeanKossaifi:
Awesome, thanks @caglayantuna and @cohenjer, merging!

PR #255, opened 2021-04-08
fixed modes and compatibility issues

caglayantuna:
As we talked during hals PR here https://github.com/tensorly/tensorly/pull/224#discussion_r565358313, I changed fixed modes to None. In addition, I did some edits according to pep8 style in _cp.py and _nn_cp.py file.


JeanKossaifi:

PR #262, opened 2021-04-22
Non negativity in only some modes

MarieRoald:
Often, it is useful to impose non-negativity in only some of the modes of our tensor decomposition models. For example, if we center our data across a mode, then we cannot impose non-negativity in that mode, but we can still constrain the uncentered modes.

This is also useful for PARAFAC2 models, where constraining all modes are difficult, and require some other update scheme than the traditional ALS algorithm. 

Luckily, with the newly implemented CP-HALS algorithm (thank you @caglayantuna and @cohenjer!), we can implement CP with non-negativity in only some modes with only minor modifications. Also, since PARAFAC2 uses CP behind-the-scenes, we can also implement non-negativity in only some mode for PARAFAC2.

This pull request adds the `nn_modes` argument to `non_negative_parafac_hals` and `parafac2`, which can be a list of constrained modes, `None` or the string `'all'` for all modes (defaults to 'all' for `non_negative_parafac_hals` and `None` for `parafac2`).

### DIFF
--- a/tensorly/decomposition/tests/test_cp.py
+++ b/tensorly/decomposition/tests/test_cp.py
@@ -255,6 +255,9 @@ def test_non_negative_parafac_hals_one_unconstrained():
     assert_(best_congruence(B, nn_estimate[1][1]) > 0.99, "Factor recovery not high enough")
     assert_(best_congruence(C, nn_estimate[1][2]) > 0.99, "Factor recovery not high enough")
 
+    assert_(T.all(nn_estimate[1][0] > -1e-10))
+    assert_(T.all(nn_estimate[1][2] > -1e-10))
+
 
 @pytest.mark.xfail(tl.get_backend() == 'tensorflow', reason='Fails on tensorflow')
 def test_sample_khatri_rao():
--- a/tensorly/decomposition/tests/test_cp.py
+++ b/tensorly/decomposition/tests/test_cp.py
@@ -1,3 +1,4 @@
+import itertools
 import numpy as np
 import pytest
 
@@ -11,6 +12,29 @@ from ...random import random_cp
 from ...tenalg import khatri_rao
 from ... import backend as T
 from ...testing import assert_array_equal, assert_
+
+
+def tucker_congruence(A, B):
+    As = A/T.norm(A, axis=0)
+    Bs = B/T.norm(B, axis=0)
+
+    return T.dot(T.transpose(As), Bs)
+
+
+def best_congruence(A, B):
+    _, r = T.shape(A)
+    corr_matrix = T.abs(tucker_congruence(A, B))
+
+    best_corr = 0
+    for permutation in itertools.permutations(range(r)):
+        corr = 1
+        for i, j in zip(range(r), permutation):
+            corr *= corr_matrix[i, j]
+        
+        if corr > best_corr:
+            best_corr = corr
+    
+    return best_corr
 
 
 @pytest.mark.parametrize("linesearch", [True, False])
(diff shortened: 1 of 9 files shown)


cohenjer:
Great addition @MarieRoald ! I checked the code, as far as I'm concerned it's good to go!

tensorly/decomposition/tests/test_cp.py
```diff
@@ -13,6 +14,29 @@
 from ...testing import assert_array_equal, assert_
 
 
+def tucker_congruence(A, B):
```
JeanKossaifi, line 17: Should we just call it congruence_coefficient and add a short docstring ? 
I find it very useful, do you think it may be worth making it available more generally?
Or should it be used just under the hood, e.g to "align" CP decompositions?

tensorly/decomposition/tests/test_cp.py
```diff
@@ -13,6 +14,29 @@
...
+    As = A/T.norm(A, axis=0)
+    Bs = B/T.norm(B, axis=0)
+
+    return T.dot(T.transpose(As), Bs)
+
+
+def best_congruence(A, B):
```
JeanKossaifi, line 24: If we do decide to make it more widely accessible, I guess we could also return the optimal permutation?

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -90,6 +90,109 @@ def test_parafac2(normalize_factors, init):
...
+    
+    # Fit with only one iteration to check non-negativity
+    weights, factors, projections = random_parafac2(
+        shapes=[(15 + rng.randint(5), 30) for _ in range(25)],
+        rank=rank,
+        random_state=rng
+    )
+    # The default random parafac2 tensor has non-negative A and C
+    # wet therefore multiply them randomly with -1, 0 or 1 to get both positive and negative components
+    factors = [factors[0] * rng.randint(-1, 2, factors[0].shape), factors[1], factors[2] * rng.randint(-1, 2, factors[2].shape)]
```
JeanKossaifi, line 154: the rng.randint should probably be wrapped in a tl.tensor

JeanKossaifi (commented):
Indeed, a great addition, thanks @MarieRoald!
I find the function to find the best congruence coefficient quite useful and it may be worth making it more widely accessible?

MarieRoald:
Thank you @cohenjer and @JeanKossaifi 

@JeanKossaifi: Yes, I think it is a good idea to make the congruence coefficient function more readily available and return the permutation. However, in that case, we should probably use the optimal average congruence instead of the optimal product of the congruences. This allows us use the `linear_sum_assignment` function in `scipy.optimize` to avoid the exponential runtime (in terms of number of components) of iterating over all possible column permutations.

My main question is where in TensorLy we should have this? Should we add a new module in `tensorly.metrics`? For example `tensorly.metrics.factors`?

I added that to this branch, together with a unit test that compares it to the slow version (i.e. iterating over all permutations). Let me know if I should place it somewhere else.

Also, I changed the unit test for both non-negative CP with one unconstrained mode and PARAFAC2 to use this congruence function.

JeanKossaifi:
Thanks @MarieRoald - I agree, let's go for metrics.factors and if we find a better place later we can always move it! 

This is a great PR - if you have time it would be great to have an example for the congruence_factor - e.g. aligning two CP? 
I also love having the slow and fast version - it's great to explain the coef. in a pedagogical manner, might be a good thing to add to the example or somewhere in the user guide too! 
In the mean-time, merging this!

Issue #263, opened 2021-04-25
The stopping condition of PARAFAC2 uses an absolute loss decrease condition, not a relative loss decrease condition


MarieRoald:
In TensorLy, the PARAFAC2-ALS algorithm stops based on an absolute loss decrease condition. That is

```python
if rec_error[-2] - rec_error[-1] < tol:
    break
```

where `rec_error` is the Frobenius norm of the reconstruction error. However, the PARAFAC2-ALS algorithm is usually described with a relative loss decrease condition based on the SSE. That is

```python
if rec_error[-2]**2 - rec_error[-1]**2 < tol * rec_error[-2]**2:
    break
```

While this difference is not large, I believe it is better to follow the [reference MATLAB implementation](http://www.models.life.ku.dk/algorithms) by Rasmus Bro to prevent confusion. 

Unfortunately, this change will alter the behaviour for existing code. There are three options that I see:

- Make the change, with a note in the documentation
- Make the change, with a note in the documentation and a disableable warning
- Make the change, with an option to enable the old behaviour.
- Don't make any change

I don't know which is best, but I have implemented the first option and written documentation for it (in [this branch](https://github.com/MarieRoald/tensorly/tree/parafac2_stopping_condition)). If we decide to go with the first or second option, then I think we should specify which version the behaviour changed (like scikit-learn does, see for example the `validate` argument of the [`FunctionTransformer` documentation](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html)). Since I didn't know the relevant version, I left it as with a TODO in the docstring.

What do you think is best action here?

JeanKossaifi:
Thanks @MarieRoald. In general I think we should go with the best possible default so happy to go with 1 (using the reference by Rasmus Bro + note in the documentation). Is there any study of the behaviour of various convergence criteria? 

We probably need to give some more thoughts to the convergence criteria in general - @cohenjer might have more experience with these.

cohenjer:
Relative decrease is indeed generally better than absolute decrease. Absolute decrease is extremely problem-dependent while the relative decrease is a little more robust and a default value of say 1e-8 makes more sense.

Since the P2 in Tensorly is still relatively new, its Ok I think to go with 1 like Jean said.

JeanKossaifi:
Fixed in #267, thanks @MarieRoald!

PR #267, opened 2021-05-06
Minor fixes for PARAFAC2 (#263 and more)

MarieRoald:
This pull request fixes some small issues with the PARAFAC2 code

- Changed to a relative loss decrease stopping condition
- Added an absolute loss decrease stopping condition which defaults to 1000 times the machine precision (necessary for convergence on noise-free PARAFAC2 tensors) (discussion here: #263 )
- Changed the default maximum number of iterations to 2000
- Added nn_modes to the docstring (thanks @IsabellLehmann :smiley:)
- Updated the `Parafac2` class to have the same arguments and docstring as the `parafac2` 
- Added tests for stopping conditions of PARAFAC2

The added absolute tolerance and new maximum number of iterations were also based on the MATLAB code by Rasmus Bro (available here: http://www.models.life.ku.dk/algorithms).

@JeanKossaifi, before we can merge this, we should update the version number in the documentation (see the todos on line 185, 203, 378 and 396 of the _parafac2.py file). What would be the right version number?

Also, when I built the sphinx documentation, I noticed that there were no special formatting for the `versionchanged` HTML class. Maybe we should update the CSS?

### DIFF
--- a/tensorly/decomposition/_parafac2.py
+++ b/tensorly/decomposition/_parafac2.py
@@ -207,6 +207,11 @@ def parafac2(tensor_slices, rank, n_iter_max=2000, init='random', svd='numpy_svd
         That is, when the relative sum of squared error is less than the specified tolerance.
         The absolute tolerance is necessary for stopping the algorithm when used on noise-free
         data that follows the PARAFAC2 constraint.
+    nn_modes: None, 'all' or array of integers
+        (Default: None) Used to specify which modes to impose non-negativity constraints on.
+        We cannot impose non-negativity constraints on the the B-mode (mode 1) with the ALS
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
Awesome, thanks @MarieRoald!

I've bumped the current version to `'0.6.1'`. This is a good point, we can add it to the scss and extend either the generic `admonition` or a specific one, e.g. `note`:

https://github.com/tensorly/tensorly-sphinx-theme/blob/d7d7a6a1c97535481296015b1778912e74760053/src/sass/tensorly_style.scss#L187

MarieRoald:
I added the version number, and I also changed the absolute tolerance to 1e-13, since 1000*eps was too large for the single-precision backends.

PR #268, opened 2021-05-06
Test utility for class wrappers and some small bugfixes

yngvem:
I noticed that many of the small class wrappers for the decomposition functions had outdated signatures, so I made a test utility to automatically compare the decomposition function and the wrapped class. It essentially works like this

1. Inspect the decomposition function to find all arguments with default values (`_get_defaultkwargs`)
1. Create a function that checks all the given keyword arguments against a predefined dictionary (generated via a factory function that takes this dictionary as input in the `_get_decomposition_checker`)
1. Use the `monkeypatch` fixture in PyTest to monkeypatch the check function created in step 2 to replace the specified decomposition function
1. Create a decomposition object where each argument of the decomposition function that had a default value are specified with a keyword-argument, setting it equal to a test-string
1. Fit this model. However, since the decomposition function is monkeypatched, we instead check that all arguments are set correctly

I tested this and found inconsistent signatures for many of the decomposition classes, and fixed these inconsistencies in all cases except for `Parafac2`, since I know @marieroald has fixed that in #267.

### DIFF
--- a/tensorly/testing.py
+++ b/tensorly/testing.py
@@ -1,3 +1,6 @@
+import sys
+from inspect import getfullargspec
+
 import numpy as np
 
 from tensorly import backend as T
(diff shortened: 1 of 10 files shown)


JeanKossaifi:
Thanks @yngvem, this is really neat! It will be super useful to make sure the doc is up to date!

yngvem:
After merging the latest updates to main all tests pass 😁

tensorly/testing.py
```diff
@@ -25,5 +28,66 @@ def _tensor_to_numpy(x):
...
+        kwonlydefaults = {}
+
+    start_defaults_idx = len(arguments) - len(defaults)
+    arguments = arguments[start_defaults_idx:]
+    default_args = {argument: default for argument, default in zip(arguments, defaults)}
+
+    return {**default_args, **kwonlydefaults,}
+
+
+def _get_decomposition_checker(supposed_kwargs, output_length):
```
JeanKossaifi, line 51: Could we add a short docstring?

JeanKossaifi (commented):
Thanks @yngvem, this is great!

I think it would be worth documenting this in the [contribution guide](https://github.com/tensorly/tensorly/blob/main/doc/development_guide/contributing.rst). We can add a short section "Testing" and explain how this works (adding monkeypatching to the test functions can be particularly confusing) so would be great to guide contributors who may add new decompositions!

JeanKossaifi:
@yngvem would be great to finalize this PR! Happy to help with some of it if you're short on time! :)

yngvem:
> 
> 
> @yngvem would be great to finalize this PR! Happy to help with some of it if you're short on time! :)

Thanks for reminding me, I've been quite busy, but I managed to set off some time to look at it today again :)

JeanKossaifi:
Wonderful, thanks a lot @yngvem! :D 
Merging.

PR #270, opened 2021-05-16
Rewrote vonneumann_entropy to handle multidimensional tensors.

taylorpatti:


tensorly/backend/tensorflow_backend.py
```diff
@@ -198,8 +198,12 @@ def log2(self,x):
     (tf.einsum, 'einsum'),
     (tf.tensordot, 'tensordot'),
     (tfm.sin, 'sin'),
+<<<<<<< HEAD
+    (tfm.cos, 'cos')
+=======
```
JeanKossaifi, line 203: Looks like there was a merge conflict that wasn't resolved - here the lines below should be kept, adding cos and anu.

tensorly/metrics/entropy.py
```diff
@@ -18,6 +18,8 @@ def vonneumann_entropy(tensor):
     -------
     von_neumann_entropy : order-0 tensor
     """
+    square_dim = int(tl.sqrt(tl.prod(tl.tensor(tensor.shape))))
```
JeanKossaifi, line 21: It may be worth documenting this in the docstring and specifying the possible sizes of tensor (either directly a matrix or a tensor satisfying prod(t.shape) = N**2).

JeanKossaifi:
Thanks @taylorpatti  - the TensorFlow issue is due to merge conflict unresolved in the backend file and we can add a little more detail in the docstring about the input sizes (and perhaps a short Notes section explaining how it is computed with a reference?). Otherwise looks good to me!

PR #271, opened 2021-05-18
Bugfix:HALS convergence checking fails when non-negativity is not imposed on final mode

MarieRoald:
@IsabellLehmann informed me of a bug in the convergence checking of NN CP HALS. If non-negativity is not imposed on the final factor matrix, then the convergence checking will fail, and if non-negativity is imposed, then the loss computation will be off by one iteration. The culprit is line 423 in _nn_cp.py, which says

    iprod=tl.sum(tl.sum(mttkrp*factor, axis=0) *weights)

This works only accidentally when non-negativity is imposed on all modes, since the final factor matrix of the previous iterate is stored in a variable named factor. If non-negativity is not imposed on the final mode, we get an error because the line

    factor=tl.solve(tl.transpose(pseudo_inverse), tl.transpose(mttkrp))

which sets `factor` to the transposed factor matrix. Fixing it is simple, we only change 

    iprod=tl.sum(tl.sum(mttkrp*factor, axis=0) *weights)

to be

    iprod=tl.sum(tl.sum(mttkrp*factors[-1], axis=0) *weights)

I added this fix and a simple test.

### DIFF
--- a/tensorly/decomposition/_nn_cp.py
+++ b/tensorly/decomposition/_nn_cp.py
@@ -420,7 +420,7 @@ def non_negative_parafac_hals(tensor, rank, n_iter_max=100, init="svd", svd='num
                 factors[mode] = tl.transpose(factor)
         if tol:
             factors_norm = cp_norm((weights, factors))
-            iprod = tl.sum(tl.sum(mttkrp * factor, axis=0) * weights)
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
Nice fix, thanks @MarieRoald and @IsabellLehmann, merging!

PR #275, opened 2021-05-27
Issue5

matthewmerris:
Resolved failing l-bfgs-b optimization.

Added scaling by tensor size to the intermediate gradient tensor calculation as per algo1 in paper


matthewmerris:
Disregard

PR #276, opened 2021-05-31
current version uses rank rather than ranks

bkmgit:


JeanKossaifi:
Looks good, thanks @bkmgit !

PR #278, opened 2021-06-04
adding tl.trace and tl.cumsum functions

caglayantuna:
For another PR that I am planning to submit soon, I need to use trace and cumsum functions for some algorithms. Since these functions could be useful for other algorithms as well, I wanted to add them to Tensorly. According to my experiments, I haven't seen any issue with the backends.


JeanKossaifi:

PR #279, opened 2021-06-07
Bugfix: Initialising non-negative CP with precomputed decomposition erroneously modified the initialisation

MarieRoald:
Here is another bugfix to a bug encountered by @IsabellLehmann. When we initialised non-negative CP with a pre-computed decomposition, the pre-computed decomposition was modified before fitting. Specifically, this line led to problems

    kt.factors = [tl.abs(f) for f in kt[1]]

Since it took the absolute value of all factor matrices. If we input a pre-defined decomposition, then it shouldn't be modified by the initialisation function. This led to problematic behaviour for PARAFAC2, when some factor matrices had negative elements.


JeanKossaifi:
Great, thanks @IsabellLehmann for reporting and @MarieRoald for the nice fix!

Issue #272, opened 2021-05-18
Parafac doesn't work for sparse tensors


meettaraviya:
#### Describe the bug

After creating a sparse tensor using the `sparse` library as per the example in the [documentation](http://tensorly.org/stable/user_guide/sparse_backend.html), I cannot pass it to the `tensorly.contrib.sparse.decomposition.parafac` without getting the following error:

```
RuntimeError: Cannot convert a sparse array to dense automatically. To manually densify, use the todense method.
```

I have given a minimal code snippet. I know that the documentation suggests using `tensorly.decomposition.parafac`,  but uncommenting either of the commented lines doesn't help either.

#### Code to Reproduce

```
import numpy as np
import sparse
from tensorly.contrib.sparse import tensor
import tensorly as tl
from tensorly.contrib.sparse.decomposition import parafac
# from tensorly.decomposition import parafac

X = sparse.COO(np.arange(24).reshape((3, 4, 2)).astype(float))
X = sparse.COO(X.coords, X.data, shape=(100, 100, 100))
X = tensor(X)
# X = tl.tensor(X)

factors = parafac(X, rank=1)
```

#### Actual result

I get the following error:

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-11-28a8abf458f2> in <module>
      8 X = sparse.COO(np.arange(24).reshape((3, 4, 2)).astype(float))
      9 X = sparse.COO(X.coords, X.data, shape=(100, 100, 100))
---> 10 X = tl.tensor(X)
     11 
     12 factors = parafac(X, rank=1)

~/opt/miniconda3/lib/python3.8/site-packages/tensorly/backend/__init__.py in inner(*args, **kwargs)
    159 
    160     def inner(*args, **kwargs):
--> 161         return _get_backend_method(name)(*args, **kwargs)
    162 
    163     # We don't use `functools.wraps` here because some of the dispatched

~/opt/miniconda3/lib/python3.8/site-packages/tensorly/backend/numpy_backend.py in tensor(data, dtype)
     12     @staticmethod
     13     def tensor(data, dtype=None):
---> 14         return np.array(data, dtype=dtype)
     15 
     16     @staticmethod

~/opt/miniconda3/lib/python3.8/site-packages/sparse/_sparse_array.py in __array__(self, *args, **kwargs)
    227 
    228         if not AUTO_DENSIFY:
--> 229             raise RuntimeError(
[...]

PR #280, opened 2021-06-07
Sparse backend usage section example

caglayantuna:
This pull request is related to issue #272. People may use this tensor to apply sparse parafac, however this version returns an error when sparse parafac is used on `X`. In addition, `unfold` function in sparse backend doesn't seem to work as it is expected. Therefore, I have used data class and tl.sort to print sequential values.


JeanKossaifi:
We shouldn't use sorting as it changes the ordering of the elements. `unfold` at least used to work as expected, not sure what happened.

JeanKossaifi:
Thanks @caglayantuna, merging! Fixes #272

Issue #264, opened 2021-04-29
potential bug in non-negative PARAFAC


rusillini:
#### Describe the bug

In tensorly 0.5.1 installed from the Anaconda channel, non-negative PARAFAC with normalization returns NaNs as a result when run on GPU using PyTorch 1.8.1 as the backend. Non-negative PARAFAC without normalization flag works fine on GPU.

#### Steps or Code to Reproduce

```python
import numpy as np
import tensorly as tl
from tensorly.decomposition import non_negative_parafac

tl.set_backend('pytorch')

p1_g = np.array([1,0,2,0], dtype=float)
p1_c = np.array([1,1,0,0], dtype=float)
p2_g = np.array([0,1,2,0], dtype=float)
p2_c = np.array([0,1,1,0], dtype=float)
p3_g = np.array([1,0,0,2], dtype=float)
p3_c = np.array([0,0,1,1], dtype=float)

t = np.outer(p1_c,p1_g)
t = np.append(t,np.outer(p2_c,p2_g),axis=0)
t = np.append(t,np.outer(p3_c,p3_g),axis=0)
t = t.reshape(3,4,4)

t = tl.tensor(t, device='cuda:0')
factors = non_negative_parafac(t, rank=3, normalize_factors=True)
print(factors[1])
factors = non_negative_parafac(t, rank=3)
print(factors[1]) 
```

#### Expected behavior

PARAFAC should give meaningful results with and without the normalization flag when run on GPU with PyTorch as the backend.

#### Actual result

[tensor([[nan, nan, nan],
        [nan, nan, nan],
        [nan, nan, nan]], device='cuda:0'), tensor([[nan, nan, nan],
        [nan, nan, nan],
        [nan, nan, nan],
        [nan, nan, nan]], device='cuda:0'), tensor([[nan, nan, nan],
        [nan, nan, nan],
        [nan, nan, nan],
        [nan, nan, nan]], device='cuda:0')]
[tensor([[1.9754e-02, 0.0000e+00, 2.0000e+00],
        [2.9770e+00, 0.0000e+00, 4.4415e-04],
        [0.0000e+00, 2.1693e+00, 0.0000e+00]], device='cuda:0'), tensor([[3.7141e-05, 2.0000e-12, 1.1050e+00],
        [7.6445e-01, 2.0000e-12, 1.0991e+00],
        [7.6456e-01, 1.0000e+00, 0.0000e+00],
[...]

PR #281, opened 2021-06-17
Normalization

caglayantuna:
This pull request is related to issue #264. We suggest some updates in _nn_cp and _cp.py files in order to have consistent normalization options for non_negative_parafac, non_negative_parafac_hals and parafac functions. In these 3 functions, we added weights to mttkrp and pseudo_inverse (accum for non_negative_parafac) computation; 

```python
mttkrp = unfolding_dot_khatri_rao(tensor, (weights, factors), mode)
pseudo_inverse = tl.reshape(weights, (-1, 1)) * pseudo_inverse * tl.reshape(weights, (1, -1))
``` 

Since we have used weights to compute mttkrp, we removed the weights from the iprod computation;

```python
iprod = tl.sum(tl.sum(mttkrp * factors[-1], axis=0))
```

Finally, we suggest to use cp_normalize function after the error computation for all functions. 

According to our experiments, we don't observe any error as in issue #264 with these modifications.

### DIFF
--- a/tensorly/decomposition/_cp.py
+++ b/tensorly/decomposition/_cp.py
@@ -338,6 +338,8 @@ def parafac(tensor, rank, n_iter_max=100, init='svd', svd='numpy_svd',
             factor = tl.transpose(tl.solve(tl.transpose(pseudo_inverse),
                                   tl.transpose(mttkrp)))
             factors[mode] = factor
+            if normalize_factors and mode != modes_list[-1]:
+                   weights, factors = cp_normalize((weights, factors))
 
         # Will we be performing a line search iteration
         if linesearch and iteration % 2 == 0 and iteration > 5:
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
I'm not sure this is necessarily better - it will be slightly heavier computationally. It's also not necessary as the weights would just be naturally absorbed in the last factor and we discussed, I don't have a strong intuition on whether one is better than the other. In Kolda's seminal paper(s) (and in the tensor-toolbox), I believe they do something similar to what we already have. 

I think a discussion is needed to simplify and make both code and API more uniform, I'll post that in the issue. 

It would be helpful to benchmark both approaches and see how it affects convergence, performance and numerical stability (e.g. are the factors in a better range, etc)?

cohenjer:
@JeanKossaifi Sure we should try even further to study normalization, but for now the nonnegative Parafac normalization is bugged (#264) and with the PR, we add a class method for normalization, which we can always tinker with later on. I think it would be better to merge the current PR, and maybe open an issue for further discussions?

JeanKossaifi:
Thanks @cohenjer, I agree with all your points.

JeanKossaifi:
Thanks for the great work @caglayantuna and @cohenjer. 
Is everyone happy with merging this?

caglayantuna:
Thanks @JeanKossaifi. From my side, it is ok.

JeanKossaifi:
Awesome, merging!

PR #282, opened 2021-06-24
Added random_state to CP and NN_CP when init='random'

earmingol:
This pull request added the option of passing random_state to CP and NN_CP decomposition. Although the option was implemented, it was working only for cases when init='svd' and not when init='random'.


JeanKossaifi:
Great, thanks @earmingol, merging!

PR #283, opened 2021-06-24
Normalization for Tucker factors

caglayantuna:
In this PR, we suggest to add `normalize_factors` option to Tucker decomposition functions (tucker, non_negative_tucker and non_negative_tucker_hals). To do this, first we normalize factors by their norms as in the parafac. Then, we weight the unfolded core with the norm of each factor separately.

### DIFF
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -76,7 +76,8 @@ def initialize_tucker(tensor, rank, modes, random_state, init='svd', svd='numpy_
     
 
 def partial_tucker(tensor, modes, rank=None, n_iter_max=100, init='svd', tol=10e-5,
-                   svd='numpy_svd', random_state=None, verbose=False, mask=None):
+                   svd='numpy_svd', random_state=None, verbose=False, mask=None,
+                   normalize_factors=False):
     """Partial tucker decomposition via Higher Order Orthogonal Iteration (HOI)
 
         Decomposes `tensor` into a Tucker decomposition exclusively along the provided modes.
@@ -107,6 +108,7 @@ def partial_tucker(tensor, modes, rank=None, n_iter_max=100, init='svd', tol=10e
         the values are missing and 1 everywhere else. Note:  if tensor is
         sparse, then mask should also be sparse with a fill value of 1 (or
         True).
+    normalize_factors : if True, aggregates the core which will contain the norms of the factors.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
Thanks, good to have, for consistency with CP if anything! In regular Tucker case since we use SVD the factors will already be unitary.

cohenjer:
We will update this PR as soon as we reach some consensus for the CP normalization strategy!

JeanKossaifi:
Looks good overall but we need to update the class accordingly. Why normalize in partial_tucker since the factors are obtained via truncated SVD?

caglayantuna:
I updated the classes accordingly. 

Actually, our target was to be consistent with the current parafac normalization method which normalizes all the factors in the inner loop as we do for Tucker factors in this PR. However, parafac normalization approach may change when we decide which method to use among the methods [here](https://github.com/tensorly/lab/pull/2). Then, we can change this Tucker normalization PR accordingly or we can keep it as it is.

cohenjer:
I agree with @JeanKossaifi we don't need to normalize after svd; normalization is only meant for nonnegative Tucker. In the end, we normalize so that each factor has unit column norm, so that normalization in nnTucker is consistent with the outputs of Tucker and CP+normalization !

caglayantuna:
Ok, I removed normalization from Tucker since it is unnecessary. On the other hand, I keep normalization for both non-negative Tucker decompositions which are ready to merge.

JeanKossaifi:
Shouldn't we adjust the core after normalizing the factors? 
Perhaps we can add a function `normalize_factors` since we're using this snippet frequently.

caglayantuna:
I added a `tucker_normalize` function which updates the core after normalizing all the factors

cohenjer:
@JeanKossaifi What did you mean by ``after'' ? In the previous commit, the core was already updated after each factor normalization. @caglayantuna the new version is not correct I think? Let's revert to the previous one.

We can however replace unfold with transpose? Which one is faster? Also we are going to add a test function.

JeanKossaifi:
@cohenjer you're right I missed that line.

I don't think we need to unfold: we can either: i) use a n-mode product with diag(weights) or ii) reshape the weights to be `(1,)*i + 1 + (1,)*(tl.ndim(tensor) - i - 1)` and take an element-wise multiplication with the core (that should be the fastest)

JeanKossaifi:
Looks good to me, merging. Thanks @caglayantuna and @cohenjer!

PR #293, opened 2021-07-07
CMTF-ALS

IsabellLehmann:
This version works for the numpy backend. The other backends fail because I used np.linalg.lstsq. If we have a tensorly-based implementation of that function, we can change it. Then it should hopefully work for all backends.


aarmey:
All the backends are working, aside from tensorflow that has a shape mismatch error when concatenating the matrix. I'm not sure what is happening there.

JeanKossaifi:
Ah that's annoying - perhaps we can use `stack` instead?

IsabellLehmann:
> Ah that's annoying - perhaps we can use `stack` instead?

I don't think so, as stack concatenates the arrays along a new axis while we want to concatenate along an existing axis (see e.g. https://stackoverflow.com/a/43673479/8307067). In the debug mode tried replacing
`tl.concatenate((unfolded, matrix), axis=1)`
by
`tl.stack((unfolded, matrix), axis=1)`,
but that raises a "Value error: all input arrays must have the same shape"

aarmey:
It's probably something before the concat, because the two arrays are indeed mismatched shapes:

> tensorflow.python.framework.errors_impl.InvalidArgumentError: ConcatOp : Ranks of all input tensors should match: shape[0] = [96,3] vs. shape[1] = [7] [Op:ConcatV2] name: concat

merajhashemi:
I think the issue in TF is in the `solve_least_squares` function. In TensorFlow unlike other other backends `tf.linag.lstsq` only returns the least squares solution. So `tl.lstsq(A, B)[0]` is the first row of the solution instead of the whole solution. I've made a PR (#305) to add `lstsq` to the backend to unify this for all backends.  



https://github.com/tensorly/tensorly/blob/a0aa423e716036c8cb63d6fc57312a4bb3c1c65a/tensorly/tenalg/core_tenalg/least_squares.py#L4-L18

aarmey:
Good catch @merajhashemi!

IsabellLehmann:
> (1) If you perform SVD initialization on the coupled mode (after concatenation), then start with ALS on the uncoupled modes, there's no need to align the factors. This might be more suitable for Tensorly but would require a custom CMTF initialization function.

The align_factors function was used to overcome the permutation and scaling ambiguity (when comparing true with estimated factor matrices). Is that not longer necessary with the specific SCV initialization or do you mean another alignment?

aarmey:
> > (1) If you perform SVD initialization on the coupled mode (after concatenation), then start with ALS on the uncoupled modes, there's no need to align the factors. This might be more suitable for Tensorly but would require a custom CMTF initialization function.
> 
> The align_factors function was used to overcome the permutation and scaling ambiguity (when comparing true with estimated factor matrices). Is that not longer necessary with the specific SVD initialization or do you mean another alignment?

Ah! I misunderstood what it was used for then. I thought it was used to ensure correspondence between the SVD initialized factors within the matrix and tensor.

aarmey:
> Makes sense, thanks!
> I think either is good: we can merge the PR and add 1) later or add it directly to this PR.
> 
> 1. also sounds great, are/will the datasets be public? If so it would be great to start having some (tensor) datasets in TensorLy (as well as various modes and coupling!).

@JeanKossaifi we have two datasets that could make good examples of CP—one is a [3-mode tensor of SARS-CoV-2 serology](https://syno.seas.ucla.edu/drive/d/s/l3ghhaGzkMs8cjoONZsbo5kgtEt1TZD3/6KhcgAZoi8DksO7sVOzo4XlJNgBWGhKh-BLCADTjC4Qg), another is a 4-mode tensor of profiling a [cytokine's responses in different immune cells](https://www.biorxiv.org/content/10.1101/2021.07.03.451002v1). Send me a message on Slack if there's somewhere you'd like me to put these with a demonstration notebook. (I don't want to completely derail this PR.)

For our use of CMTF, one dataset is public but has many missing values and thus probably does not make a great example. A second does not contain missing values and results in clear patterns but is made up of RNA expression profiling and is consequently quite large (~5-10MB compressed). If that's not actually all that big then it's publicly available and making a demonstration notebook wouldn't be difficult.

JeanKossaifi:
Great! So for this PR, shall we just remove lstsq ( #305 was merged and fixes the remaining issue) and merge? 
We can make further changes as we go along?

aarmey:
If all the builds pass this should be ready.

JeanKossaifi:
Awesome, thanks @IsabellLehmann and @aarmey! 
Merging.

PR #294, opened 2021-07-07
Gradient of the difference norm between a CP and dense tensor

aarmey:
A few notes:

- This does not take the gradient with respect to the weights.
- This only solves for the l2 norm difference. Adding other norms, if useful, probably wouldn't be too much work.
- In testing, I figured that comparing to numerical differencing would be slow, so I check that the negative gradient is a direction of descent with some small perturbation.
- There might be faster ways to calculate this... this is just what I came across in a literature search.

### DIFF
--- a/tensorly/cp_tensor.py
+++ b/tensorly/cp_tensor.py
@@ -359,13 +359,13 @@ def cp_lstsq_grad(cp_tensor, tensor, return_loss=False, mask=None):
 
     .. math::
 
-        \nabla 0.5 ||\mathcal{X} - [\mathbf{w}; \mathbf{A}, \mathbf{B}, \mathbf{C}]||^2
+        \nabla 0.5 ||\\mathcal{X} - [\\mathbf{w}; \\mathbf{A}, \\mathbf{B}, \\mathbf{C}]||^2
 
-    where :math:`[\mathbf{w}; \mathbf{A}, \mathbf{B}, \mathbf{C}]` is the CP decomposition with weights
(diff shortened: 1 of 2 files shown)


tensorly/cp_tensor.py
```diff
@@ -355,6 +354,36 @@ def cp_flip_sign(cp_tensor, mode=0, func=None):
     return CPTensor((weights, factors))
 
 
+def cp_lstsq_grad(cp_tensor, tensor, mask=None):
+    """Returns the gradient of the least squares difference between a CP and dense tensor.
+
```
yngvem, line 361: Maybe add the exact function we compute the gradient for?

```rst
This function computes (for a third-order tensor)

.. math::

    \nabla 0.5 ||\mathcal{X} - [\mathbf{w}; \mathbf{A}, \mathbf{B}, \mathbf{C}]||^2

where :math:`[\mathbf{w}; \mathbf{A}, \mathbf{B}, \mathbf{C}]` is the CP decomposition with weights :math:`\mathbf{w}` and factor matrices :math:`\mathbf{A}`, :math:`\mathbf{B}` and :math:`\mathbf{C}`.
```

yngvem (commented):
Nice addition! I just have one note — with this implementation, we can compute the loss essentially for free (`0.5*tl.sum(error**2)`), and you are often interested in getting both the loss and the gradient simultaneously (e.g. in a line-search). Maybe we should add an optional keyword-argument `return_loss` that we can enable to compute the loss too?

JeanKossaifi:
Looks good to me, thanks @aarmey @yngvem! Merging.

PR #295, opened 2021-07-13
NN-Tucker hals sparsity coefficient issue

caglayantuna:
We realized an issue when using only one sparsity coefficient value (without a list ) with non_negative_tucker_hals function. This PR solves this issue and adds core_sparsity_coefficient as an input in order to provide separate factor and core sparsity values.

### DIFF
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -393,7 +393,7 @@ def non_negative_tucker_hals(tensor, rank, n_iter_max=100, init="svd", svd='nump
         If set to None, the algorithm is computed without sparsity
         Default: None
     core_sparsity_coefficient : array of float. This coefficient imposes sparsity on core
-        when it is udpated with fista.
+        when it is updated with fista.
         Default: None
     fixed_modes : array of integers (between 0 and the number of modes)
         Has to be set not to update a factor, 0 and 1 for U and V respectively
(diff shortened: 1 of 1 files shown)


tensorly/decomposition/_tucker.py
```diff
@@ -460,7 +464,7 @@ def non_negative_tucker_hals(tensor, rank, n_iter_max=100, init="svd", svd='nump
     """
     rank = validate_tucker_rank(tl.shape(tensor), rank=rank)
     n_modes = tl.ndim(tensor)
-    if sparsity_coefficients is None or len(sparsity_coefficients) != n_modes:
+    if sparsity_coefficients is None or isinstance(sparsity_coefficients, float):
```
JeanKossaifi, line 467:
```suggestion
    if sparsity_coefficients is None or not isinstance(sparsity_coefficients, Iterable):
```

This requires to import Iterable:
```
from collections.abc import Iterable
```

JeanKossaifi (commented):
Looks good to me! 
I just suggested checking if the parameter is iterable rather than float (so as to also include edge cases, e.g. int).

tensorly/decomposition/_tucker.py
caglayantuna: Thanks, I changed this line according to your suggestion.

JeanKossaifi:
Thanks, merging!

PR #298, opened 2021-07-23
Make parafac() robust to complex tensors

maximeguillaud:
This makes PARAFAC correctly handle tensors with complex factors.

### DIFF
--- a/tensorly/cp_tensor.py
+++ b/tensorly/cp_tensor.py
@@ -598,7 +598,7 @@ def unfolding_dot_khatri_rao(tensor, cp_tensor, mode):
     _, rank = _validate_cp_tensor(cp_tensor)
     weights, factors = cp_tensor
     for r in range(rank):
-        component = multi_mode_dot(tensor, [tl.conj(f[:, r]) for f in factors], skip=mode)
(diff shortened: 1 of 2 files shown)


tensorly/cp_tensor.py
```diff
@@ -598,7 +598,7 @@ def unfolding_dot_khatri_rao(tensor, cp_tensor, mode):
     _, rank = _validate_cp_tensor(cp_tensor)
     weights, factors = cp_tensor
     for r in range(rank):
-        component = multi_mode_dot(tensor, [f[:, r] for f in factors], skip=mode)
+        component = multi_mode_dot(tensor, [f[:, r].conj() for f in factors], skip=mode)
```
yngvem, line 601: This should probably be `tl.conj(f[:, r])`

yngvem:
Nice addition :)

My only comment is that we should maybe add a test with a complex tensor before merging.

JeanKossaifi:
Thanks @maximeguillaud! Looks ready to me too after making the changes mentioned by @yngvem

JeanKossaifi:
Looks good to me aside from the test - if anyone has time to look at this would be great, it's just missing a test. Otherwise I'll look at it in a couple days, would be good to merge if it fixes a bug in complex support. @caglayantuna @yngvem

PR #305, opened 2021-08-20
Add lstsq to backend

merajhashemi:
This PR adds `lstsq`, a function that computes the minimum norm least squares solution, to the backend.


JeanKossaifi:
Thanks @merajhashemi! I don't think we need to just return the solution, e.g. the residuals etc are also useful.

merajhashemi:
@JeanKossaifi TensorFlow and PyTorch(<1.9) do not return the rank and singular values, and also TensorFlow doesn't return the residuals. I'm not sure if it's worth adding an extra call to `svd` and `matrix_rank`.

JeanKossaifi:
Hmm - perhaps we can add at least the residuals and just compute that in the TensorFlow case? We're thinking of dropping tf backend anyway, and JAX is more powerful and flexible.

JeanKossaifi:
Looks great @merajhashemi, merging!

PR #312, opened 2021-08-25
Fix svd for pytorch < 1.8

### LINKED ISSUE #311: [BUG] right singular vectors transposed in `tl.svd` (PyTorch < 1.8)
```python3
import tensorly as tl
tl.set_backend('pytorch')

A = tl.randn((5, 6))
u,s,v = tl.svd(A, full_matrices=False)
print(v.shape)
```
outputs `[6, 5]`, while the correct shape should be `[5, 6]`.

merajhashemi:
This PR fixes #311.


JeanKossaifi:
Awesome, thanks!

PR #313, opened 2021-08-25
Fix dot and matmul in PyTorch and TF

### LINKED ISSUE #307: [BUG] dot and matmul in TensorFlow and PyTorch
@JeanKossaifi Thanks for adding matmul. There are a couple of small bugs I noticed.
`matmul` is not added to the PyTorch backend:

```python3
with tl.backend_context('pytorch'):
    a = tl.randn((4,))
    b = tl.randn((4,))
    print(tl.matmul(a,b).shape)
```
raises a `NotImplementedError:`.

---
In TensorFlow `tf.matmul` does not support one dimensional tensors:
```python3
with tl.backend_context('tensorflow'):
    a = tl.randn((5,4))
    b = tl.randn((4,))
    print(tl.matmul(a,b).shape)
```
raises `InvalidArgumentError: In[0] and In[1] has different ndims: [5,4] vs. [4] [Op:MatMul]`.

So I think we should do something like this:
```python3
def tf_matmul(a, b):
    if tf.rank(a)==1 or tf.rank(b)==1:
        return tf.tensordot(a, b, 1)
    return tf.linalg.matmul(a, b)
```

---
While I was looking more closely I found two cases where `tl.dot` is broken in TensorFlow and PyTorch.
So if we plan to keep `tl.dot` (I don't think we need to) we should fix them. 

```python3
with tl.backend_context('pytorch'):  # or 'tensorflow'
    a = tl.randn((6,7,4,3))
    b = tl.randn((6,9,3,2))
    print(tl.dot(a,b).shape)
```
raises a `RuntimeError:` while in NumPy and other backends for ndim > 2 tensors the last mode of `a` and the second to last mode of `b` are contracted. 

---
Also unlike numpy zero-dimensional tensors are not supported:
```python3
with tl.backend_context('pytorch'):  # or 'tensorflow'
    a = tl.randn((5,4))
    b = tl.randn(())
    print(tl.dot(a,b).shape)
```

---
I think a small fix like this will do it:
```python3
def pytorch_dot(a, b):
    if a.ndim > 2 and b.ndim > 2:
        return torch.tensordot(a, b, dims=([-1], [-2]))
    if not a.ndim or not b.ndim:
        return a*b
    return torch.matmul(a, b)
```
```python3
def tf_dot(a, b):
    if tf.rank(a) > 2 and tf.rank(b) > 2:
        return tf.tensordot(a, b, axes=([-1], [-2]))
    if not tf.rank(a) or not tf.rank(b):
        return a*b
    return tf_matmul(a, b)  # tf.linalg.matmul doesnt support 1d tensors
```

merajhashemi:
This PR resolves #307.

### DIFF
--- a/tensorly/backend/pytorch_backend.py
+++ b/tensorly/backend/pytorch_backend.py
@@ -99,6 +99,14 @@ class PyTorchBackend(Backend):
         return torch.norm(tensor, **kwds)
 
     @staticmethod
+    def dot(a, b):
+        if a.ndim > 2 and b.ndim > 2:
+            return torch.tensordot(a, b, dims=([-1], [-2]))
+        if not a.ndim or not b.ndim:
+            return a * b
+        return torch.matmul(a, b)
+
+    @staticmethod
     def mean(tensor, axis=None):
         if axis is None:
             return torch.mean(tensor)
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
Great, thanks, merging!

PR #320, opened 2021-08-28
Fix  singular vectors to be orthonormal

### LINKED ISSUE #319: [BUG] singular vectors in `tl.partial_svd` are not orthonormal
When `n_eigenvecs` is greater than the input matrix rank, the returned singular values are not orthonormal. However, the result is still a valid low-rank decomposition (). 

```python3
import tensorly as tl
import numpy as np
np.random.seed(0)

a = tl.dot(tl.randn((4, 2)), tl.randn((2, 4)))
u, s, v = tl.partial_svd(a, n_eigenvecs=3)
print(tl.dot(u.T, u))
```
```
[[ 1. -0.  0.]
 [-0.  1.  0.]
 [ 0.  0.  0.]]
```
 ☹️.

---
To fix this bug we could use the Gram–Schmidt process and extend the singular vectors basis corresponding to the zero singular values.

merajhashemi:
This PR fixes #319.

### DIFF
--- a/tensorly/backend/core.py
+++ b/tensorly/backend/core.py
@@ -1079,6 +1079,7 @@ class Backend(object):
                 S = np.sqrt(np.clip(S, 0, None))
                 S = np.clip(S, np.finfo(S.dtype).eps, None)  # To avoid divide by zero warning on next line
                 V = np.dot(matrix.T.conj(), U * np.where(np.abs(S) <= np.finfo(S.dtype).eps, 0, 1/S)[None, :])
+                U, S, V = U[:, ::-1], S[::-1], V[:, ::-1]
                 V, R = np.linalg.qr(V)
                 V = V * (2*(np.diag(R) >= 0) - 1)  # we can't use np.sign because np.sign(0) == 0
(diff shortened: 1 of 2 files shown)


merajhashemi:
@JeanKossaifi I think this is ready 🙂

JeanKossaifi:
Looks good but do isn't it a surprising behaviour, and shouldn't we follow the NumPy API here?
When dealing with a low rank matrix (e.g.  `max_dim > n_eigenvecs > matrix_rank`), I would *expect* some of the eigenvectors to be 0 . What do others think @merajhashemi @cohenjer @aarmey @yngvem @MarieRoald?

merajhashemi:
This is consistent with NumPy's svd as the new test cases runs for `tl.truncated_svd` as well. The left (right) singular vectors corresponding to the zero singular vectors form the basis for the null space of  (). They shouldn't be zero vectors.

JeanKossaifi:
Yes, you're right, I read too quickly. Thanks @merajhashemi, merging!

Issue #297, opened 2021-07-22
tl. diag function supports main diagonal only with Tensorflow backend 


caglayantuna:
#### Describe the bug
Tensorly [diag](http://tensorly.org/stable/modules/generated/tensorly.diag.html#tensorly.diag) function has an input `k` to select diagonal.  However, this option doesn't work properly with Tensorflow backend.

#### Steps or Code to Reproduce
Selected vector is always being on the main diagonal. For instance:

```python
tl.diag([1,1,1], 1)

array([[1, 0, 0],
       [0, 1, 0],
       [0, 0, 1]], dtype=int32)>
```
[Same function](https://www.tensorflow.org/api_docs/python/tf/linalg/diag) with tensorflow gives an expected output:

```python
tf.linalg.diag([1,1,1], k=1)

array([[0, 1, 0, 0],
       [0, 0, 1, 0],
       [0, 0, 0, 1],
       [0, 0, 0, 0]], dtype=int32)>
```

Most probably, reason is the place of `k` in the Tensorflow function. But I couldn't find how to fix it in Tensorly.

merajhashemi:
I think the problem is the use of `tf.linalg.tensor_diag` instead of `tf.linalg.diag` in tensorly. The former only supports the main diagonal. So I guess an easy fix would be to change this in:
https://github.com/tensorly/tensorly/blob/e669855fe755a38ba775b700b9bcfb601880bd25/tensorly/backend/tensorflow_backend.py#L175

caglayantuna:
Tkanks @merajhashemi . I submitted a PR #321 to solve this issue.

PR #321, opened 2021-08-31
fix diag function for tensorflow and pytorch backends

caglayantuna:
This PR solves issue #297 and adds `k` input to torch backend since it was missing.


JeanKossaifi:
Great, thanks @caglayantuna! Merging.

PR #322, opened 2021-09-01
Add test for matmul

merajhashemi:
There was a small bug in `tl.matmul` for TensorFlow when the first argument was 1-d.


JeanKossaifi:
Grea, thanks @merajhashemi , merging!

PR #324, opened 2021-09-17
Copy cptensor

caglayantuna:
This PR adds copy function for CPTensor class in order to avoid modification when `init` is a CPTensor for tensorly decomposition methods. Users can use this as in the following example;

```python          
weights_init, factors_init = initialize_cp(tensor, rank=rank)
cp_init = CPTensor((weights_init, factors_init))
weights, factors = parafac(tensor, rank=rank, init=cp_init.copy())
```

### DIFF
--- a/tensorly/cp_tensor.py
+++ b/tensorly/cp_tensor.py
@@ -71,6 +71,9 @@ class CPTensor(FactorizedTensor):
     def to_unfolded(self, mode):
         return cp_to_unfolded(self, mode)
 
+    def cp_copy(self):
+        return CPTensor((T.copy(self.weights), [T.copy(self.factors[i]) for i in range(len(self.factors))]))
+
     def mode_dot(self, matrix_or_vector, mode, keep_dim=False, copy=True):
         """n-mode product of a CP tensor and a matrix or vector at the specified mode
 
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
You're right, we might as well provide it.

I'm not sure deepcopy will work for all frameworks. Perhaps we should just return a new Factorized Tensors, created with a copy of each factors.

cohenjer:
@caglayantuna Can you check deepcopy with all backends (if you did not already) ? 

It will be an easy PR then :)

caglayantuna:
I checked current version with all the backends and I haven't seen any problem. Besides, I have added a test to `test_cp_tensor` function which passes backend tests. To avoid any naming conflict, I changed the function name to `cp_copy`. 

For this feature, my first suggestion was the following function:

```
def copy(self):
    return CPTensor((T.copy(self.weights), [T.copy(self.factors[i]) for i in range(len(self.factors))]))
```

I think, this way is similar to what you mean about copying each factor of a Factorized Tensor. If this version is better, I can change the function easily.

JeanKossaifi:
Yes, I ran into some issues with deep copy, so I think your initial suggestion might be the best to avoid issues with e.g. nn.Parameters in PyTorch @caglayantuna

caglayantuna:
I updated this PR according to our discussion

JeanKossaifi:
Looks good, thanks @caglayantuna, merging!

Issue #292, opened 2021-07-06
Allow last mode to be fixed in decompositions


cohenjer:
#### Problem
In response to #72 and for a few months now, Tensorly has allowed to fix some modes in decompositions, e.g. using the following syntax for parafac:
```python
out = parafac(tensor, rank, fixed_modes=[0,1])
```
In the above example, the first two modes are fixed, meaning they will not be updated during the iterations of the optimization algorithm.

However, for technical reasons, the last mode may not be fixed, i.e. if the data is a three-way tensor,
```python
out = parafac(tensor, rank, fixed_modes=[0,2])
```
will raise a warning and mode `2` will not be fixed.

#### Solution

Note the following two facts:
- the last mode should never be fixed inside the optimization algorithm since the last mode update rule allows for fast error estimation. 
- if all modes are fixed then the decomposition should just return the initialization

Therefore, a quick patch would be:
- check if all modes are fixed. If true, then don't run the optimization algorithm.
- Otherwise, if the last mode is fixed, then permute a non-fixed mode with the last one at the beginning of the algorithm, and permute it back in place at the end of the algorithm.

#### Who can PR for this issue
Anyone :) This is not a difficult fix, so feel free to tackle this issue to get started in Tensorly!

ParvaH:
Hi, 

Can I try to add code for first part of the solution suggested - _check if all modes are fixed. If true, then don't run the optimization algorithm_?

-Parvathy

PR #325, opened 2021-09-20
Adding check for all modes fixed case and if true then to just return the initialization

ParvaH:
This pull request adds check for case where all modes are fixed and if true, then just returns the initialization without running the optimization algorithm. This is part of solution to issue #292 (Allow last mode to be fixed in decompositions)


cohenjer:
Looks good to me :) Thanks @ParvaH !

ParvaH:
> 
> 
> Thanks! Actually, your code should be moved before we remove the last mode from the list of fixed modes, a few lines above ^^

Thanks for the feedback.  Just placed the code before the part where last mode is removed.

JeanKossaifi:
Looks good to me, are you happy with it as well @cohenjer ?

cohenjer:
Sure we can merge it. It does not fix all the problem of last mode being unfixable, but it's a nice trick anyway to avoid useless computations.

PR #326, opened 2021-09-28
add cmtf_als to init file so that it can be imported directly via tl.decomposition

IsabellLehmann:


JeanKossaifi:
Thanks @IsabellLehmann, merging!

PR #334, opened 2021-10-27
non-negative hals examples

caglayantuna:
This PR adds two examples to explain HALS method for non-negative cp and tucker decompositions. Besides, these examples compare this new HALS method with the multiplicative update method which was already in Tensorly for non-negative decompositions.


JeanKossaifi:
Awesome, thanks @caglayantuna! Merging

Issue #338, opened 2021-11-10
Potential bug in hals_nnls


ruochiz:
Hi,

Thanks for creating this amazing package.

In the tensorly/tensorly/tenalg/proximal.py function hals_nnls.

The variable `rec_error0` is not updated as it suppose to be. Based on the current version, it's always 0 except the iteration 1 (not after iteration 1).

This bug comes from that the original code being written in the while loop instead of for loop. And in the current version within the for loop the `rec_error0` is reset to 0 everytime. I suppose a quick fix would be to just put the `rec_error0=0` outside of the for loop. This bug would cause the hals_nnls always try to reach the n_iter_max instead of early stop.

caglayantuna:
Hi, 

You are right, it is not correct to have `rec_error0` in the loop. I opened a PR related to this issue. Thanks.

cohenjer:
Hi,
Thanks @ruochiz, indeed it was wrong :) It will be corrected very soon! Please don't hesitate if you have more feedback on this new feature.

PR #339, opened 2021-11-15
solves hals rec_error0 issue and does some pep improvements

caglayantuna:
This PR solves the issue #338. I removed `rec_error0` from the loop and changed `if iteration==1` to `if iteration==0` to calculate initial reconstruction error. Also, I have done some pep improvements in the same function to make it readable.


JeanKossaifi:
Thanks @caglayantuna! Looks good, merging.

PR #340, opened 2021-11-15
fix tensorflow dtype issue

caglayantuna:
This PR adds [tensorflow cast](https://www.tensorflow.org/api_docs/python/tf/cast) function to `tensorly.tensor` in `tensorflow_backend` to allow changing dtype when tensor is a tensorflow tensor. Otherwise, dtype remains same even if a user wants to change it.

### DIFF
--- a/tensorly/backend/tensorflow_backend.py
+++ b/tensorly/backend/tensorflow_backend.py
@@ -20,8 +20,8 @@ class TensorflowBackend(Backend, backend_name='tensorflow'):
 
     @staticmethod
     def tensor(data, dtype=np.float32, device=None, device_id=None):
-        if isinstance(data, tf.Tensor):
-            return data
+        if isinstance(data, tf.Tensor) or isinstance(data, tf.Variable):
+            return tf.cast(data, dtype=dtype)
 
         out = tf.Variable(data, dtype=dtype)
         return out.gpu(device_id) if device == 'gpu' else out


JeanKossaifi:
Thanks @caglayantuna. Looks good, just one comments perhaps: should we check the type of data and explicitly cast only if needed? Though I'm not sure how tensorflow handles the cast : if they already do this check inside `cast` then the current code if fine.

caglayantuna:
[Inside cast](https://github.com/tensorflow/tensorflow/blob/v2.7.0/tensorflow/python/ops/math_ops.py#L938-L1007) , there are some checks for data type. It would be better for us to check data to avoid any error inside cast. If the input data is a tensor from another backend (except mxnet), both `tf.Variable` and `tf.cast` work fine. Mxnet returns error with both `tf.cast`  and `tf.Variable` functions. But, I think it is a rare situation for the users. Only difference is that while `tf.cast` returns `EagerTensor` , `tf.Variable` returns `ResourceVariable` . Actually, I don't know about their difference.

Also, I realized another issue. When the input data is a tf.Tensor, we don't return output with device_id. Is it important? As a suggestion, I could remove first return and data type check to use both `tf.cast` and `tf.Variable` (without dtype because it returns an error if dtype is different) functions respectively. Thus, we could have `ResourceVariable` output for all backends except mxnet.

JeanKossaifi:
Thanks @caglayantuna, looks good, merging.

PR #344, opened 2021-11-26
Add random_state for non negative PARAFAC HALS

MarieRoald:
Currently, `non_negative_parafac_hals` and `NN_PARAFAC_HALS` do not support a random seed. If random initialisation is used, then the random state is set to None. This pull requests fixes that, and adds a `random_state` argument.


JeanKossaifi:
Thanks @MarieRoald, great to have this!
I fixed the merge conflict with the previous PR but there seems to be a repeated argument for one of the calls.

JeanKossaifi:
Thanks @MarieRoald, merging!

PR #347, opened 2021-12-07
Constrained parafac example and improved docstrings

caglayantuna:
This PR adds  `plot_guide_for_constrained_cp` file to the examples in order to explain how to use constrained parafac decomposition feature. Besides, we improve the docstring of constraints as we discussed before. We also realized a missing `tl.abs` in normalize constraint and fix it.


JeanKossaifi:
Looks great, thanks @caglayantuna and @cohenjer!

PR #348, opened 2021-12-10
IL-2 stimulation dataset

borcuttjahns:


borcuttjahns:
Draft for adding NN PARAFAC Example Notebook. @aarmey - Pending your review.

aarmey:
@JeanKossaifi a first question—would you prefer datasets go in `examples/datasets/` or `tensorly/datasets`?

JeanKossaifi:
Thanks @borcuttjahns and @aarmey, looking forward to having this!

I was talking with @caglayantuna and @cohenjer as they're also working on adding some more. We were thinking to follow the [scikit-learn format](https://scikit-learn.org/stable/datasets.html), e.g. returning a dict (they return a `Bunch`, a dict exposing its keys as attribute so both `data['key']` and `data.key` are equivalent).

So based on this, how about keeping the data in tensorly/datasets. We can create a subfolder `tensorly/datasets/data` for the raw data and have the rest in `.py` scripts in `tensorly/datasets`? Then we can add in `examples` an example script like you have to demonstrate usage? 

Do we need a license file for the data or can we just share the `.npy`?

tensorly/datasets/imports.py
```diff
@@ -0,0 +1,16 @@
+"""
+Load example datasets.
+"""
+
+import numpy as np
+from sklearn.utils import Bunch
```
aarmey, line 6: @borcuttjahns TensorLy does not have sklearn as a dependency. You can instead [copy the definition of a Bunch](https://github.com/scikit-learn/scikit-learn/blob/0d378913b/sklearn/utils/__init__.py#L83). It sounds like it will be used here.

caglayantuna:
Hello, I am also working on the adding some benchmarks to Tensorly by using some existing real dataset. I would like to show you some parts of my current code below; 
```
class Dataset:

    def load_indian_pines(self):
        url = 'http://www.ehu.eus/ccwintco/uploads/6/67/Indian_pines_corrected.mat'
        r = requests.get(url, allow_redirects=True)
        image = scipy.io.loadmat(BytesIO(r.content))['indian_pines_corrected']
        metadata = {"tensor": tl.tensor(image, "float"),
                    "Reference": "Jouni, M., Dalla Mura, M., & Comon, P. (2019, July). Classification of \
                                 hyperspectral images as tensors using nonnegative CP decomposition. In \
                                 International Symposium on Mathematical Morphology and Its Applications \
                                 to Signal and Image Processing (pp. 189-201). Springer, Cham",
                    "Link": url,
                    "Application": "Classification",
                    "Constraint": "Non-negative",
                    "Modes": ["Spatial dimension", "Spatial dimension", "Hyperspectral bands"]}
        return metadata
        

    def load_chicago_crime(self):
        url = 'https://s3.us-east-2.amazonaws.com/frostt/frostt_data/chicago-crime/comm/chicago-crime-comm.tns.gz'
        r = requests.get(url, allow_redirects=True)
        zip = gzip.open(BytesIO(r.content))
        array = [[int(x) for x in line.split()] for line in zip]
        tensor = np.zeros([6186, 24, 77, 32])
        for i in range(len(array)):
            tensor[array[i][0]-1, array[i][1]-1, array[i][2]-1, array[i][3]-1] = array[i][4]
        metadata = {"tensor": tensor,
                    "Reference": "Smith, S., Huang, K., Sidiropoulos, N. D., & Karypis, G. (2018, May). "
                                 "Streaming tensor factorization for infinite data sources. In "
                                 "Proceedings of the 2018 SIAM International Conference on Data Mining "
                                 "(pp. 81-89). Society for Industrial and Applied Mathematics",
                    "Link": url,
                    "Application": "Streaming",
                    "Constraint": "Sparse",
                    "Modes": ["Day", "Hour", "Community", "Crime Type"]}
        return metadata
```        

Here I am using request library to import data from the given url directly. Thus, I don't need to store the data in the disk. It returns a dictionary which includes the data as tensor and some information to avoid any license issue (don't know if it's enough).  It is just a preliminary work, we can discuss together and decide which way to go.

aarmey:
@caglayantuna this is great, and we should definitely include the additional metadata you provide about the task, reference, etc.

Regarding how to pull in our data, it's not available in the right form elsewhere and so I think we have to store a file. However, it's originally derived from academic work with sharing/reuse requirements, and most proximately it comes from MIT-licensed work on which I am corresponding author. For these reasons I think a reference should be sufficient.

aarmey:
@JeanKossaifi this should be ready to merge after your review. We've decided to make a separate PR with the example code. I added a simple test.

JeanKossaifi:
Thanks, it looks good to me!
We could potentially add a little description in a doctoring and perhaps a `DESCR` string to the class itself?

aarmey:
@JeanKossaifi thanks. Brian has added these.

JeanKossaifi:
Thanks @aarmey and @borcuttjahns, looks great & ready to merge!
@aarmey: can you just confirm that we don't need any license or that we have permission to share/host?

aarmey:
@JeanKossaifi yes, we own the license and have licensed it under the MIT license so you have permission to distribute and/or modify it without restriction.

However, in reading through it I'm realizing we need to distribute the license with it. Where would you suggest we place the license?

JeanKossaifi:
How about we add a field LICENSE in the object? We can also add a note
about the MIT license and a link to it in the docstring?

aarmey:
@JeanKossaifi the license is now included.

JeanKossaifi:
Awesome, thank you - merging!

PR #349, opened 2021-12-23
Generalized cp

caglayantuna:
# Generalized CP Decomposition
This pull request adds a generalized and stochastic parafac decompositions [1,2] using Fista and ADAM optimizations respectively.

# Motivation
GCP allows different kind of losses such as:

1. Rayleigh
2. Bernoulli (odds and log)
3. Gamma
4. Poisson (count and log)
5. Gaussian

Depending on the data distribution, any loss can be selected by the user.

# Source Modification
* **./decomposition/_generalized_parafac.py:**

We added a new file which includes `initialize_generalized_parafac` ,  `stochastic_gradient`, `generalized_parafac`, `stochastic_generalized_parafac` functions and relevant classes.

* **./decomposition/tests/test_generalized_parafac:**
Test functions are added for each decomposition. 

* **./metrics/losses.py:**
This file includes `loss_operator` and `gradient_operator`. 

* **./metrics/test/test_losses.py:**
Tests for `loss_operator` and `gradient_operator`. 

* **./decomposition/_cp.py:**
We updated `sample_khatri_rao` function to have indices list as an input when it is necessary. 

In addition, `log` and  `exp` functions are added to all backends.

# References

[1]-  Hong, D., Kolda, T. G., & Duersch, J. A. (2020).
Generalized canonical polyadic tensor decomposition. SIAM Review, 62(1), 133-163. [Link](https://arxiv.org/abs/1808.07452)

[2]- Kolda, T. G., & Hong, D. (2020). Stochastic gradients for large-scale tensor decomposition. SIAM Journal on Mathematics of Data Science, 2(4), 1066-1095. [Link](https://arxiv.org/abs/1906.01687)

### DIFF
--- a/tensorly/metrics/losses.py
+++ b/tensorly/metrics/losses.py
@@ -1,0 +1,97 @@
+import tensorly as tl
+import math
+
+
+def loss_operator(tensor, estimated_tensor, loss, mask=None):
+    """
+    Operator to use loss functions from [1] in order to compute loss for
+    generalized parafac decomposition.
+
+    Parameters
+    ----------
+    tensor : ndarray
+    estimated_tensor : ndarray
+    loss : {'gaussian', 'gamma', 'rayleigh', 'poisson_count', 'poisson_log', 'bernoulli_odds', 'bernoulli_log'}
+    mask : ndarray
+        array of booleans with the same shape as ``tensor`` should be 0 where
+        the values are missing and 1 everywhere else.
+
+    Returns
+    -------
+    error : ndarray
+         Size based normalized loss for each entry
+    References
+    ----------
+    .. [1] Hong, D., Kolda, T. G., & Duersch, J. A. (2020).
+           Generalized canonical polyadic tensor decomposition. SIAM Review, 62(1), 133-163.
+    """
+    if mask is not None:
+        estimated_tensor = estimated_tensor * mask
+
+    epsilon = 1e-8
+    if loss == 'gaussian':
+        error = (tensor - estimated_tensor) ** 2
+    elif loss == 'bernoulli_odds':
+        error = tl.log(estimated_tensor + 1) - (tensor * tl.log(estimated_tensor + epsilon))
+    elif loss == 'bernoulli_logit':
+        error = tl.log(tl.exp(estimated_tensor) + 1) - (tensor * estimated_tensor)
+    elif loss == 'rayleigh':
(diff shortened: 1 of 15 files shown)


caglayantuna:
Hello @JeanKossaifi , in case you missed this PR, I wanted to remind you.

tensorly/metrics/losses.py
```diff
@@ -0,0 +1,97 @@
...
+
+    Returns
+    -------
+    error : ndarray
+         Size based normalized loss for each entry
+    References
+    ----------
+    .. [1] Hong, D., Kolda, T. G., & Duersch, J. A. (2020).
+           Generalized canonical polyadic tensor decomposition. SIAM Review, 62(1), 133-163.
+    """
```
JeanKossaifi, line 27: I wonder whether we are going to actually need this somewhere else? (If not, it could probably be, along with the gradient, directly in the function)

JeanKossaifi (commented):
Took me a while to get round to it as it's quite a large PR :) 

It will be great to have this added. Given how involved it is, it would be really great to have more details in the docstring, and perhaps also some text in the user-guide directly. For instance, some more detail about the loss that's optimized etc. 

Regarding the actual stochastic part, given that we are not many maintainers I am weary of adding too much, as there are libraries that focus solely on that (e.g. PyTorch). In that sense I think it may make more sense to focus here on one simple version and add more variety directly in TensorLy-Torch.

cohenjer:
So we had a chat with @caglayantuna, and we totally understand your position. Here is what we can do:
- Move the stochastic gradient part of the PR to tensorly torch, and use this occasion to remove our custom adam implementation and rather use pytorch optimizer.
- Keep the updated sample Khatri Rao function (it is already in tensorly, we merely improve on it slightly) in this PR; this will be used by the stochastic algorithm in tensorly torch and could be used in fact by many others
- keep a GCP PR with only the deterministic algorithm. It is less efficient for some losses, but this makes the PR smaller and the code easier to maintain arguably.

In any case and on a more global scale, we do plan to slow down on the big algorithmic PRs (this is probably the last one). We want to spend more time working on the documentation (user guide) and on benchmarking.

cohenjer:
Well, let's close this PR anyway, we can keep up the discussions in PR dedicated to the deterministic version, which will use scipy optimize (preleminary results are looking good, better than what we had when starting this PR). I guess adding a stochastic version to tensorly torch can come later if still useful.

PR #355, opened 2022-01-08
Fix bug with `tl.clip` for the PyTorch and TensorFlow backends

MarieRoald:
For both PyTorch and TensorFlow, the values for a_max and a_min are computed by TensorLy whenever their values are None. However, the way these values are computed leads to erroneous behaviour. With PyTorch, this occurs when `a_min > max(tensor)` and `a_max=None`. I noticed this for PyTorch when I tried to clip a negative tensor to be non-negative. Here is an example:

```python
tl.set_backend("pytorch")
x = tl.tensor([-1, -1, -1])
tl.clip(x, 0, None)
```
```
tensor([-1., -1., -1.])
```

For TensorFlow, the error occurs when `a_min=None` and `a_max<min(tensor)`. Here is an example:

```python
tl.set_backend("tensorflow")
x = tl.tensor([1, 1, 1])
tl.clip(x, None, 0)
```
```
<tf.Tensor: shape=(3,), dtype=float32, numpy=array([1., 1., 1.], dtype=float32)>
```

### DIFF
--- a/tensorly/backend/pytorch_backend.py
+++ b/tensorly/backend/pytorch_backend.py
@@ -61,14 +61,10 @@ class PyTorchBackend(Backend, backend_name='pytorch'):
 
     @staticmethod
     def clip(tensor, a_min=None, a_max=None, inplace=False):
-        if a_max is None:
-            a_max = torch.max(tensor)
-        if a_min is None:
-            a_min = torch.min(tensor)
         if inplace:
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
Thanks @MarieRoald, I always love your PRs: clear problem, clear fix and matching unit-tests! 🌠 
Merging.

PR #357, opened 2022-01-08
Make `tl.shape` return tuple for PyTorch backend

MarieRoald:
Currently, `tl.shape` returns a tuple for most backends. However, with the PyTorch backend, it returns a `torch.Size` object, which can make it more difficult to make backend agnostic code. This pull request makes it return a tuple instead and adds a test that fails if the object returned from `tl.shape` is not a `tuple`.

Alternatively, if we don't want to cast the `torch.Size` to a tuple automatically, then we could add an optional `as_tuple` argument to all backends. However, I made it cast to a tuple since that is what's done in the TensorFlow backend.

### DIFF
--- a/tensorly/backend/pytorch_backend.py
+++ b/tensorly/backend/pytorch_backend.py
@@ -46,7 +46,7 @@ class PyTorchBackend(Backend, backend_name='pytorch'):
 
     @staticmethod
     def shape(tensor):
-        return tensor.shape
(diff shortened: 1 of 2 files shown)


cohenjer (approved):
Seems very reasonable to me !

JeanKossaifi:
Thanks @MarieRoald, looks good to me too, merging.

Issue #352, opened 2022-01-07
Similarity metric between CP tensors invariant to scale, sign, permutation


aarmey:
It'd be nice to have a built-in way to compare CP tensors that doesn't have sign, scale, and permutation issues. Perhaps I could implement the [CorrIndex](https://hal.archives-ouvertes.fr/hal-03230210/document)? Our lab has dealt with this problem when comparing PARAFAC runs, and I recently met another systems biologist TensorLy user having the same problem.

cohenjer:
I think this is related to #316 , because if we can align two CP then comparison nay be done on the factors (which will be permuted and scaled so more readily comparable).
I guess CorrIndex (which I was unfamiliar with) is another solution but a little less intuitive ? Definitively valuable to have it anyways but I guess this will not solve #316.

aarmey:
Agreed. It wasn't obvious to me that it would be possible to add

aarmey:
@cohenjer, agreed. It wasn't obvious to me that it would be possible to do #316 without SciPy though? Is there information on how often a greedy approach would find a suboptimal solution?

On a more basic level, perhaps we could use a simple operation to permute the factors first. I know I've implemented this multiple times now, and it becomes annoying when you have multiple backends.

JeanKossaifi:
Would be great to have CorrIndex! For #316 I would thinking just a simple utility fun, e.g. correlation based. 

It might be a good time to discuss dropping the most cumbersome backends for the next big version and converging to the array API for all backends. We could use the opportunity to clean up / unify the code base and API.

cohenjer:
@JeanKossaifi I would be VERY happy to discuss dropping annoying backends.

@aarmey This could be a method of the CP tensor class, pretty easy to implement and also valuable imo.

Regarding the suboptimality of a correlation/greedy based method, I think that:
- When the two CP tensors compared are similar and well conditionned (no colinearity in factors or such), correlation is good enough.
- When the two CP tensors are very different or degenerate, the greedy approach may be very different from the optimal, linear programming approach (which requires e.g. the Hungarian algorithm indeed implemented in SciPy).
 
Nevertheless, it is not very sensible to try and find the best permutation to compare tensors when they are extremely different; we might artificially reduce distance between CP tensors. So correlation/greedy method might be good enough (although I typically use the optimal permutation in my research works).

cohenjer:
[...]

PR #358, opened 2022-01-09
Add CP indexing function

aarmey:
To help with #352, this adds a very simple function to allow indexing the components within a CP tensor. This makes it a little simpler to permute or remove a component.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ test:
 
 test-all:
 	TENSORLY_BACKEND='numpy' pytest -v tensorly
-	# TENSORLY_BACKEND='cupy' pytest -v tensorly
+	TENSORLY_BACKEND='cupy' pytest -v tensorly
 	TENSORLY_BACKEND='pytorch' pytest -v tensorly
 	TENSORLY_BACKEND='mxnet' pytest -v tensorly
 	TENSORLY_BACKEND='jax' pytest -v tensorly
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
Hi @aarmey - thanks, this is a neat feature. 
Actually, I was hoping to slowly merge TensorLy-Torch's new tensor class within TensorLy itself and add general indexing. The advantage is that indexing a factorized indexing returns a factorized tensor when possible and a dense tensor otherwise:

https://github.com/tensorly/torch/blob/24bdae7b5b87548c7db37c6e9a53b03259ec2002/tltorch/factorized_tensors/factorized_tensors.py#L89


What do you think?

aarmey:
Tests are still failing... looking into it.

aarmey:
As far as I can tell, there is no front-end way to index an array consistently across backends. The current solution works for all but tensorflow. Unless we plan to change tensorflow to the experimental numpy interface soon, I'll add a backend function to help with indexing.

JeanKossaifi:
Do you think we need a separate function to do this? If it's useful as it's done repetitively, perhaps we could call it something like index_cp_rank to differentiate it from indexing a tensor in factorized CP form? 

I'm happy to just skip the tensorflow test right now with pytest while we decide what to do with that backend.

aarmey:
I agree about the renaming.

I'd be interested to learn a bit more about the backend plans when we talk tomorrow. I do use this frequently but would rather not merge it if it is broken for some backends. I can always stash this away until later.

PR #359, opened 2022-01-11
Import COVID-19 systems serology example dataset

cyrillustan:
This is an example dataset of COVID-19 systems serology for demonstration of applying CP to biological research.


cyrillustan:
@aarmey You can check if this is done correctly.

aarmey:
@cyrillustan this looks good to me. @JeanKossaifi, just like the other dataset, we figured that we would merge the data itself before the example analysis. I can confirm that the license is correct here, and that there is no problem with distribution.

JeanKossaifi:
Thanks, this is really great to have!
Just one note : shouldn't we create one file per dataset and have them explicitly named? imports seems strange. We can also add the resulting function to datasets/__init__ so users can directly import from `tensorly.datasets`.

caglayantuna:
I am also planning to add some real data to Tensorly. Instead of having a lot of files, what do you think about having a function which returns data by using the name of the data in the end of imports.py (name can change) and put this function in datasets/init? For instance, this function could be;
```
def get_tensor(name):
    """
    Returns a tensor according to the given data name.
    """
    if name == "indian_pines":
        return indian_pines()
    elif name == "IL2data":
        return IL2data()
            .
            .
            .
```

JeanKossaifi:
Thanks! Let us know when this is ready for review.

cyrillustan:
@aarmey if you don't see any issue, I think this is ready

aarmey:
@cyrillustan seems ready to me.

JeanKossaifi:
Looks great!

Based on previous conversation, I guess we could remove task, name the function `load_covid19`  (or `load_covid19_serology` or something to make it less generic?) and maybe add a `DESC` with a short description of the dataset for users?

JeanKossaifi:
Amazing, thanks @cyrillustan @aarmey, exciting to have these new datasets and examples coming!

PR #361, opened 2022-01-14
Add Covid example notebook

cyrillustan:
I try to format it as much as the existing example script. Don't know how the doc website works, but I think everything may be built automatically with `sphinx-gallery`? Let me know if I need to add something to make this work.

### DIFF
--- a/examples/decomposition/apply_covid.py
+++ b/examples/decomposition/apply_covid.py
@@ -83,10 +83,10 @@ tfac.factors[2][:, 0] *= -1
 # To evaluate the classifier, we plot a ROC-AUC curve
 
 def COVIDpredict(tfac):
-    subjj = np.isin(data.samples, ["Severe", "Deceased"])
+    subjj = np.isin(data.ticks[0], ["Severe", "Deceased"])
 
     X = tfac.factors[0][subjj, :]
-    y = pd.factorize(data.samples[subjj])[0]
+    y = pd.factorize(data.ticks[0][subjj])[0]
     aucs = []
 
     kf = KFold(n_splits=10, shuffle=True)
@@ -157,9 +157,9 @@ def comp_plot(factors, xlabel, ylabel, plotLabel, ax, d=False):
 components = [str(ii + 1) for ii in range(tfac.rank)]
 
 fig3, axes = plt.subplots(1, 3, figsize=(16,6))
-comp_plot(tfac.factors[0], components, list(data.samples), "Samples", axes[0], True)
-comp_plot(tfac.factors[1], components, data.antigens, "Antigens", axes[1])
-comp_plot(tfac.factors[2], components, data.receptors, "Receptors", axes[2])
+comp_plot(tfac.factors[0], components, list(data.ticks[0]), "Samples", axes[0], True)
+comp_plot(tfac.factors[1], components, data.ticks[1], "Antigens", axes[1])
+comp_plot(tfac.factors[2], components, data.ticks[2], "Receptors", axes[2])
 
 ##############################################################################
 # From the results, we can see that serum COVID-19 immunity separates into two distinct signals,
--- a/examples/decomposition/apply_covid.py
+++ b/examples/decomposition/apply_covid.py
@@ -22,7 +22,16 @@ Apply CP decomposition to COVID-19 Serology Dataset
 
 import numpy as np
 import pandas as pd
+import tensorly as tl
+from tensorly.decomposition import parafac
 from tensorly.datasets.imports import COVID19_data
+from matplotlib import pyplot as plt
+from sklearn.linear_model import LogisticRegression
+from sklearn.metrics import roc_curve, roc_auc_score
+from sklearn.model_selection import KFold
+from itertools import groupby
+import seaborn as sns
+from scipy.stats import sem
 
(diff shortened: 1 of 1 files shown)


cyrillustan:
@aarmey here is the notebook

examples/decomposition/apply_covid.py
```diff
@@ -0,0 +1,180 @@
...
+
+comps = np.arange(1, 7)
+CMTFfacs = [parafac(data.tensor, cc, tol=1e-10, n_iter_max=1000,
+                    linesearch=True, orthogonalise=2) for cc in comps]
+
+##############################################################################
+# To evaluate how well CP decomposition explains the variance in the dataset, we plot the percent
+# variance reconstructed (R2X) for a range of ranks.
+
+def calcR2X(tFac, tIn=None):
```
JeanKossaifi, line 45: Would it be worth calling it reconstructed_variance and adding a very short docstring?
Is it used enough to add to the metrics @aarmey @cohenjer ?
In it's current form it would not be ideal due to depending on NumPy but we could probably adapt. Anyway, this is an issue separate from this PR :)

examples/decomposition/apply_covid.py
```diff
@@ -0,0 +1,180 @@
...
+    tMask = np.isfinite(tIn)
+    vTop = np.sum(np.square(tl.cp_to_tensor(tFac) * tMask - np.nan_to_num(tIn)))
+    vBottom = np.sum(np.square(np.nan_to_num(tIn)))
+    return 1.0 - vTop / vBottom
+
+import matplotlib
+from matplotlib import pyplot as plt
```
JeanKossaifi, line 52: Might look neater to group all the imports at the top of the file

examples/decomposition/apply_covid.py
```diff
@@ -0,0 +1,180 @@
...
+
+##############################################################################
+# Build a Logistic Regression classifier to predict severe versus deceased COVID-19 patients
+# -----------------------
+# An important function of CP decomposition here is to reduce data while keeping significant
+# information.
+
+from sklearn.linear_model import LogisticRegression
+from sklearn.metrics import roc_curve, roc_auc_score
+from sklearn.model_selection import KFold
```
JeanKossaifi, line 74: That's going to crash the current doc as we don't install scikit-learn. We can add it if this is crucial to have.

examples/decomposition/apply_covid.py
```diff
@@ -0,0 +1,180 @@
...
+tfac = CMTFfacs[1]
+
+# Ensure that factors are negative on at most one direction.
+tfac.factors[1][:, 0] *= -1
+tfac.factors[2][:, 0] *= -1
+
+##############################################################################
+# To evaluate the classifier, we plot a ROC-AUC curve
+
+def COVIDpredict(tfac):
```
JeanKossaifi, line 85: Note a huge deal but the function is currently a little obscure to read, mostly due to none obvious variable names.

JeanKossaifi (commented):
Great notebook! We can make the code more readable but this is a wonderful example.

examples/decomposition/apply_covid.py
aarmey: @JeanKossaifi I've been meaning to bring up whether we should include this. :-) We use it a *ton* so it would be great to add. We'll do so in a separate PR.

examples/decomposition/apply_covid.py
cyrillustan: This is mostly for making predictions and plotting the ROC curve. It would be an interesting plot to show, especially for how tensor method can be used in ML context, but this is not essential to the function.

cyrillustan: Also, I don't know if seaborn is a dependency, but we use them to plot too so hopefully it's not a problem.

examples/decomposition/apply_covid.py
cyrillustan: I made a few changes to this. Let me know which variable names are still obscure.

examples/decomposition/apply_covid.py
aarmey: @cyrillustan I think we can just plot the factors and then point out there is a difference between severe / deceased.

Regarding seaborn, if it's not too difficult to just use matplotlib, that would limit the number of necessary packages.

examples/decomposition/apply_covid.py
aarmey: If we do my suggestion above then we can just get rid of this function.

examples/decomposition/apply_covid.py
cyrillustan: Okay, removed the ROC curve and changed all plots in matplotlib. Just a reminder that even matplotlib is not a dependency currently.

JeanKossaifi: Thanks! Matplotlib is a dependency for the doc.

JeanKossaifi:
Looks great! I've merged the Covid dataset so I guess we don't need to add it here anymore?

cyrillustan:
This one branches out from this other one. So yes, I think we are good once the master merges to this branch (I think I’ve already done that)

JeanKossaifi:
Perfect, merging!

Issue #354, opened 2022-01-07
Simplex Projection does not work


cohenjer:
#### Describe the bug
I ran the `simplex_prox` new function located in proximal.py from the AOADMM PR, and it turns out it does not work.

Moreover the function can be improved, for instance by allowing 1-d tensor input and removing some computation overhead.

#### Steps or Code to Reproduce

For instance:

```python
from tensorly.tenalg.proximal import simplex_prox

t = tl.tensor([[0.4,1.5,1],[0.5,2,3],[0.6,0.3,2.9]])
print(simplex_prox(t,parameter=1))
```
produces
```python
[[0.35 0.5  0.  ]
 [0.45 1.   1.  ]
 [0.55 0.   0.9 ]]
```
 which columns do not sum to 1.

Instead the correct output should be
```python
[[0.23333333 0.25       0.        ]
 [0.33333333 0.75       0.55      ]
 [0.43333333 0.         0.45      ]]
```
By the way, the current unit test passes by chance; I suggest we replace it with another one such as above.

#### Bug source

The bug mainly comes from the fact that one of the inner computations is missing a division. I have cooked a corrected version of the function, which I will run by @caglayantuna to fine tune. We will make the correction PR asap.

JeanKossaifi:
Fixed in #363

PR #363, opened 2022-01-18
simplex projection issue

caglayantuna:
This PR fixes the issue #354.

### DIFF
--- a/tensorly/tenalg/proximal.py
+++ b/tensorly/tenalg/proximal.py
@@ -604,10 +604,10 @@ def simplex_prox(tensor, parameter):
     # Broadcasting is used to divide rows by 1,2,3...
     cumsum_min_param_by_k = (tl.cumsum(tensor_sort, axis=0) - parameter) / tl.cumsum(tl.ones([row, 1]), axis=0)
     # Added -1 to correspond to a Python index
-    to_change = tl.sum(tl.where(tensor_sort > cumsum_min_param_by_k, 1.0, 0.0), axis=0) - 1
+    to_change = tl.sum(tl.where(tensor_sort > cumsum_min_param_by_k, 1, 0), axis=0) - 1
     difference = tl.zeros(col)
     for i in range(col):
-        difference = tl.index_update(difference, tl.index[i], cumsum_min_param_by_k[int(to_change[i]), i])
+        difference = tl.index_update(difference, tl.index[i], cumsum_min_param_by_k[to_change[i], i])
     if col > 1:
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
Thanks, looks good to me, happy to merge. I guess @cohenjer already reviewed?

cohenjer:
@JeanKossaifi Let me skim through it one more time tomorrow just in case, but everything should be OK now.

cohenjer:
Looks good to me !

tensorly/tenalg/proximal.py
```diff
@@ -593,17 +593,25 @@ def simplex_prox(tensor, parameter):
...
+    cumsum_min_param_by_k = (tl.cumsum(tensor_sort, axis=0) - parameter) / tl.cumsum(tl.ones([row, 1]), axis=0)
+    # Added -1 to correspond to a Python index
+    to_change = tl.sum(tl.where(tensor_sort > cumsum_min_param_by_k, 1.0, 0.0), axis=0) - 1
     difference = tl.zeros(col)
     for i in range(col):
-        if to_change[i] > 0:
-            difference = tl.index_update(difference, tl.index[i], tl.cumsum(tensor_sort, axis=0)[int(to_change[i] - 1), i])
-    difference = (difference - parameter) / to_change
-    return tl.clip(tensor - difference, a_min=0)
+        difference = tl.index_update(difference, tl.index[i], cumsum_min_param_by_k[int(to_change[i]), i])
```
JeanKossaifi, line 610: the explicit casting to int of `to_change[i]` is causing issues in numpy -- does it not work if we don't explicitly cast?

caglayantuna: I removed casting and used `1` and `0` instead of `1.0` and `0.0` when creating `to_change`. However mxnet returns an error when testing not simplex but constrained parafac with simplex interestingly. I am trying to find a solution but if you have any idea, it would be nice to hear.

JeanKossaifi: What's the issue with MXNet?

caglayantuna: Now, there is no issue with any backend.

cohenjer:
There seems to be no more issues with the backends, and now the math is correct :)

cohenjer:
I think we can merge, and close #354 as well.

JeanKossaifi:
Perfect, thanks @caglayantuna and @cohenjer, merging!
Closes #354.

PR #364, opened 2022-01-19
CorrIndex implementation for comparing decomposition outputs

hmbaghdassarian:
@aarmey  I have an implementation of CorrIndex here that can hopefully provide a good starting point

### DIFF
--- a/tensorly/metrics/similarity.py
+++ b/tensorly/metrics/similarity.py
@@ -47,7 +47,7 @@ def correlation_index(factors_1: list, factors_2: list, tol: float = 5e-16) -> f
     c_prod_mtx = tl.abs(tl.matmul(tl.conj(X_1.T), X_2))
 
     # correlation index scoring
-    n_elements = tl.shape(c_prod_mtx)[1] * tl.shape(c_prod_mtx)[0] #TODO: N notation in equation 8 unclear, make sure this is correct
+    n_elements = tl.shape(c_prod_mtx)[1] + tl.shape(c_prod_mtx)[0]
     score = (1/(n_elements)) * (tl.sum(tl.abs(tl.max(c_prod_mtx,1) - 1)) + tl.sum(tl.abs(tl.max(c_prod_mtx, 0) - 1)))
     if score < tol:
         score = 0
--- a/tensorly/metrics/similarity.py
+++ b/tensorly/metrics/similarity.py
@@ -1,0 +1,81 @@
+from typing import List
+
+import numpy as np
+from sklearn.preprocessing import normalize
+
+# Authors: Hratch Baghdassarian <hmbaghdassarian@gmail.com>, Erick Armingol <earmingol14@gmail.com>
+# similarity metrics for tensor decompsoitions
+
+def stack_loading_matrices(factors: List[np.ndarray]) -> np.ndarray:
+    """Vertically stack the loading/factor matrices.
+
+    Parameters
+    ----------
+    factors : List[np.ndarray]
+        The loading/factor matrices [A^1 ... A^R] for a decomposed low-rank tensor from its factors.
+
+    Returns
+    -------
+    np.ndarray
+        A vertically stacked matrix of shape [sum(tensor.shape)xR)] 
+    """
+    
+    if not np.all([A.shape[1] for A in factors]):
+        raise ValueError('factors should be a list of loading matrices')
+    return np.vstack(factors)
+
+def calc_corrindex(C: np.ndarray) -> float:
+    """Calculate the CorrIndex.
+
+    Parameters
+    ----------
+    C : np.ndarray
+        absolute value of the product of two MxN matrices that have had their columns L2 normalized, 
+        taking the complex conjugate transpose of the first matrix -- C = |X1^H * X2|
+
+    Returns
+    -------
+    float
+        CorrIndex metric [0,1]; lower score means higher between matrices
+    """
+    
+    MN = C.shape[1]*C.shape[0]
+    return (1/(MN))*(np.sum(np.abs(np.amax(C, 1) - 1)) + np.sum(np.abs(np.amax(C, 0) - 1)))
+
+def corrindex(factors_1: List[np.ndarray], factors_2: List[np.ndarray], tol: float = 5e-16) -> float:
+    """CorrIndex implementation to assess tensor decomposition outputs. 
+    From Sobhani et al 2022 (https://doi.org/10.1016/j.sigpro.2022.108457). 
+    Metric is scaling and column-permutation invariant.  
+
+    Parameters
+    ----------
+    factors_1 : List[np.ndarray]
+        The loading/factor matrices [A^1 ... A^R] for a low-rank tensor from its factors, output from first decomposition
+    factors_2 : List[np.ndarray]
+        The loading/factor matrices [A^1 ... A^R] for a low-rank tensor from its factors, output from second decomposition
+    tol : float, optional
+        precision threshold below to call the CorrIndex score 0, by default 1e-16
+        
+    Returns
+    -------
+    score : float
+         CorrIndex metric [0,1]; lower score means higher similarity between matrices
+    """   
+    # get stacked loading matrices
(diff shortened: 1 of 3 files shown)


tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,81 @@
...
+        absolute value of the product of two MxN matrices that have had their columns L2 normalized, 
+        taking the complex conjugate transpose of the first matrix -- C = |X1^H * X2|
+
+    Returns
+    -------
+    float
+        CorrIndex metric [0,1]; lower score means higher between matrices
+    """
+    
+    MN = C.shape[1]*C.shape[0]
```
earmingol, line 42: Given that C is generated from C = np.abs(np.matmul(np.matrix(X_1).H, X_2)) (in Line 76) and both X_1, X_2 must be of shape M x N,  C should be of shape N x N. So C.shape[1] should be equal to C.shape[0]. Maybe include an assert condition before this.

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,81 @@
...
+    return (1/(MN))*(np.sum(np.abs(np.amax(C, 1) - 1)) + np.sum(np.abs(np.amax(C, 0) - 1)))
```
aarmey, line 43: To make this work for other backends, you'll want to use the Tensorly operations. You can add `import tensorly as tl` at the top, then most of these can just be changed from `np.X` to `tl.X`. I think `amax` might have to become `max`.

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,81 @@
+from typing import List
+
+import numpy as np
+from sklearn.preprocessing import normalize
```
aarmey, line 4: Tensorly doesn't have sklearn as a dependency, so we can't use this. Luckily there's a `tl.norm` operation to get the norm for normalization.

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,81 @@
...
+    
+    if X_1.shape != X_2.shape: 
+        raise ValueError('Factor matrices should be of the same shapes')
+    
+    # normalize columns to L2 norm - even if ran decomposition with normalize_factors = True  
+#   X/np.apply_along_axis(np.linalg.norm, 0, X) 
+    X_1 = normalize(X_1, norm = 'l2', axis = 0)
+    X_2 = normalize(X_2, norm = 'l2', axis = 0)
+    
+    C = np.abs(np.matmul(np.matrix(X_1).H, X_2))
```
aarmey, line 76: The element-wise complex conjugate is implemented as `tl.conj`, so this should be possible to do with that and a transpose.

aarmey (commented):
A couple initial suggestions.

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Returns
+    -------
+    np.ndarray
+        A vertically stacked matrix of shape [sum(tensor.shape)xR)] 
+    """
+
+    if len({A.shape[1] for A in factors}) != 1:
+        raise ValueError('factors should be a list of loading matrices')
+    return tl.concatenate(factors)
```
JeanKossaifi, line 26: I don't think we need a function for that?

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,83 @@
...
+
+def calc_corrindex(C: np.ndarray) -> float:
```
JeanKossaifi, line 28: What about just `correlation_index`?

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,83 @@
...
+    """Calculate the CorrIndex.
+
+    Parameters
+    ----------
+    C : np.ndarray
```
JeanKossaifi, line 33: Could we use a more explicit name for C?

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Returns
+    -------
+    float
+        CorrIndex metric [0,1]; lower score means higher between matrices
+    """
+    if C.shape[1] != C.shape[0]:
+        raise ValueError('C must be a square matrix')
+
+    MN = C.shape[1]*C.shape[0] #TODO: N notation in equation 8 unclear, make sure this is correct
```
JeanKossaifi, line 45: what about size or total size? Also better to use tl.shape(C)[1] just to make it backend agnostic.

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,83 @@
...
+    factors_2 : List[np.ndarray]
+        The loading/factor matrices [A^1 ... A^R] for a low-rank tensor from its factors, output from second decomposition
+    tol : float, optional
+        precision threshold below to call the CorrIndex score 0, by default 1e-16
+
+    Returns
+    -------
+    score : float
+         CorrIndex metric [0,1]; lower score means higher similarity between matrices
+    """
```
JeanKossaifi, line 66: I think this function and the one above can be merged and just called correlation_index, what do you think?

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,83 @@
...
+        precision threshold below to call the CorrIndex score 0, by default 1e-16
+
+    Returns
+    -------
+    score : float
+         CorrIndex metric [0,1]; lower score means higher similarity between matrices
+    """   
+    # get stacked loading matrices
+    X_1 = stack_loading_matrices(factors_1)
+    X_2 = stack_loading_matrices(factors_2)
```
JeanKossaifi, line 69: We can just do the concatenation here, no need for a separate function?

tensorly/metrics/similarity.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if X_1.shape != X_2.shape:
+        raise ValueError('Factor matrices should be of the same shapes')
+
+    # normalize columns to L2 norm - even if ran decomposition with normalize_factors = True  
+    X_1 = X_1/tl.norm(X_1, axis = 0)
```
JeanKossaifi, line 75: Shouldn't we check whether the norm is non-zero?
Also, we don't need spaces around `=` as it indicates a keyword argument (pep8)

JeanKossaifi (changes requested):
Thanks @hmbaghdassarian and @aarmey!
I've left a few comments, mostly for clarity.

hmbaghdassarian:
@JeanKossaifi thanks for the feedback, I tried implementing your suggestions. Let me know if I missed anything or if you have more changes you want

tensorly/metrics/similarity.py
```diff
@@ -1,83 +1,55 @@
...
+    X_1 = X_1/col_norm_1
+    X_2 = X_2/col_norm_2
 
-    C = tl.abs(tl.matmul(tl.conj(X_1.T), X_2))
+    # generate the correlation index input
+    c_prod_mtx = tl.abs(tl.matmul(tl.conj(X_1.T), X_2))
 
-    score = calc_corrindex(C)
+    # correlation index scoring
+    n_elements = tl.shape(c_prod_mtx)[1] * tl.shape(c_prod_mtx)[0] #TODO: N notation in equation 8 unclear, make sure this is correct
```
earmingol, line 50: This should be
```n_elements = tl.shape(c_prod_mtx)[1] + tl.shape(c_prod_mtx)[0]```
To meet the "2N" in the equation of the paper

aarmey: Thanks! That's updated.

aarmey:
I've now made some changes so that this *should* work for all backends.

earmingol:
> 

Thanks @aarmey ! We have just one change to implement. It is an option for computing the index either on the stacked matrices or on a 1-to-1 manner on each pair of A matrices and then select the worse score.

JeanKossaifi:
Looks good to me - @aarmey @cohenjer for an additional pair of eyes :)
Beautiful docstring for `correlation_index` 👌

aarmey (approved):
Also looks great to me! Looking forward to using this.

PR #367, opened 2022-01-27
copy tucker tensor

caglayantuna:
This PR adds `tucker_copy` function to TuckerTensor as `cp_copy` function merged in #324.


cohenjer:
Looks good to me, thanks @caglayantuna ! We have the same syntax for CP tensors so this is consistant with other tensorly .copy() functions.

PR #371, opened 2022-02-09
Test for reproducibility of CP

aarmey:
I spent a little while trying to hunt down why the CP tests very occasionally fail and came across this observation. As I understand it, with the random seed set the decomposition should be essentially reproducible. I'm not sure if this is the case for SVD initialization, but seems like it certainly should be the case for random initialization with a seed set.

Is my expectation reasonable? I'm not sure if there is something non-deterministic in one of the ALS steps I might be missing.

If this is reasonable to expect, I can start by adding the same check for the CP initialization function.

### DIFF
--- a/tensorly/decomposition/tests/test_cp.py
+++ b/tensorly/decomposition/tests/test_cp.py
@@ -10,40 +10,37 @@ from ...cp_tensor import cp_to_tensor
 from ...random import random_cp
 from ...tenalg import khatri_rao
 from ... import backend as T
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
It's still marked as draft - let me know when it's ready!

aarmey:
@JeanKossaifi thanks for pointing that out. This should be ready.

JeanKossaifi:
This looks good to me, great to have this fixed, thanks @aarmey! Merging.

PR #380, opened 2022-02-17
Permute cp factors

caglayantuna:
# Permute cp factors
This pull request adds `cp_permute_factors` function to `cp_tensor.py` and relevant test to `test_cp_tensor`.

This function has 2 inputs which are ref_cp_tensor (one cp tensor) and tensors_to_permute (one cp tensor or list of cp tensor). It returns permuted tensor or tensors and permutation list for each permuted tensor.

# tl.gather
Since permuting factors require an operator to swap columns, I added `tl.gather` function to Tensorly to make this function work for tensorflow backend. While using [:, indices] is enough for all the other backends, index operation is not possible with tensorflow unfortunately. I have used `take` function for numpy, jax, mxnet and `indice_select` for pytorch backend to have same behaviour for all backends.

# Note
I don't see any error with Jax backend in my computer but it fails in github test.

### DIFF
--- a/tensorly/cp_tensor.py
+++ b/tensorly/cp_tensor.py
@@ -7,7 +7,7 @@ from .base import fold, tensor_to_vec
 from ._factorized_tensor import FactorizedTensor
 from .tenalg import khatri_rao, multi_mode_dot
 from .utils import DefineDeprecated
-
+from .metrics.factors import congruence_coefficient
 import numpy as np
 
 # Author: Jean Kossaifi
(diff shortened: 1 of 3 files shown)


caglayantuna:
@cohenjer noticed a [function](http://tensorly.org/stable/modules/generated/tensorly.metrics.factors.congruence_coefficient.html#tensorly.metrics.factors.congruence_coefficient) in tensorly which returns permutation with the same scipy function. So, I updated this PR and now using that function but I still need `gather` to swap the columns for tensorflow.

aarmey:
Yes, please, let's skip the tensorflow test for now. The overlap of folks using this with TF will be nil. Given the constant frustration with TF I think it's reasonable to require the experimental interface eventually.

@cyrillustan and @murphymadeleine21, you'll be interested in this.

caglayantuna:
I updated this PR and skipped tensorflow test. To import `congruence_coefficient` , I had to use tl.cp_normalize in `metrics.entropy` instead of importing it to avoid circular import error.

JeanKossaifi:
Awesome! This is super useful and we should advertise it to users - @caglayantuna, let's also add it to the API in the doc and perhaps add a few lines in the user-guide? I'll merge this and we can add through another PR?

PR #381, opened 2022-02-18
Adds log function to tensorly

caglayantuna:


JeanKossaifi:
Looks good, thanks @caglayantuna, merging.

j6k4m8:
I think this still needs to be added to [the top-level `backend` method import list](https://github.com/tensorly/tensorly/blob/main/tensorly/__init__.py#L34), unless I'm misunderstanding?

JeanKossaifi:
Yes, good point, thanks @j6k4m8!
Would be great to have a small PR for that and perhaps add some doc on how to add new functionalities to the backend.

j6k4m8:
You got it!! :)

PR #382, opened 2022-02-18
User defined indices list for sample kr

caglayantuna:
As we discussed in #349 , I updated sample_khatri_rao function to allow users to input their own indices list.

### DIFF
--- a/tensorly/decomposition/_cp.py
+++ b/tensorly/decomposition/_cp.py
@@ -447,6 +447,7 @@ def sample_khatri_rao(matrices, n_samples, skip_matrix=None, indices_list=None,
         if not None, index of a matrix to skip
 
     indices_list : list, default is None
+        a list of length `matrices` containing the indices with the range of relevant matrice row
         if None, random indices will be created
 
     random_state : None, int or numpy.random.RandomState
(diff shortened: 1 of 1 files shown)


tensorly/decomposition/_cp.py
```diff
@@ -446,6 +446,9 @@ def sample_khatri_rao(matrices, n_samples, skip_matrix=None,
     skip_matrix : None or int, optional, default is None
         if not None, index of a matrix to skip
 
+    indices_list : list, default is None
+        if None, random indices will be created
```
JeanKossaifi, line 450: Can we add a longer description, e.g. what is selected and range of the indices?

JeanKossaifi (commented):
Looks good - I'd maybe just add a bit more info in the docstring.

JeanKossaifi (commented):
I'd just detail more what is expected for that list of indices and what exactly it does but looks good to me otherwise!

tensorly/decomposition/_cp.py
caglayantuna: I added some information, I hope it is ok for you

tensorly/decomposition/_cp.py
```diff
@@ -446,6 +446,10 @@ def sample_khatri_rao(matrices, n_samples, skip_matrix=None,
     skip_matrix : None or int, optional, default is None
         if not None, index of a matrix to skip
 
+    indices_list : list, default is None
```
JeanKossaifi, line 449:
```suggestion
    indices_list : list, default is None
        Contains, for each matrix in `matrices`, a list of indices of rows to sample.
        if None, random indices will be created
```
Slight suggestion for clarity

JeanKossaifi:
Looks good thanks @caglayantuna  - just added a small suggestion.

JeanKossaifi:
Looks good to me, merging!

Issue #310, opened 2021-08-23
Fix Test 


JeanKossaifi:
test_tr_tensor : seems to have accuracy issues occasionally (reduce decimal to 5? )

merajhashemi:
## `test_parafac`
```
_______________________ test_parafac[svd-3-5-True-True] ________________________

linesearch = True, orthogonalise = True, true_rank = 3, rank = 5, init = 'svd'
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7fa74e97a760>

    @pytest.mark.parametrize("linesearch", [True, False])
    @pytest.mark.parametrize("orthogonalise", [True, False])
    @pytest.mark.parametrize("true_rank,rank", [(1, 1), (3, 5)])
    @pytest.mark.parametrize("init", ['svd', 'random'])
    def test_parafac(linesearch, orthogonalise, true_rank, rank, init, monkeypatch):
        """Test for the CANDECOMP-PARAFAC decomposition
        """
        rng = tl.check_random_state(1234)
        tol_norm_2 = 0.01
        tol_max_abs = 0.05
        tensor = random_cp((6, 8, 4), rank=true_rank, orthogonal=orthogonalise, full=True, random_state=rng)
        fac, errors = parafac(tensor, rank=rank, n_iter_max=200, init=init, tol=10e-5, random_state=rng, orthogonalise=orthogonalise, linesearch=linesearch, return_errors=True)
    
        # Check that the error monotonically decreases
        # TODO: This doesn't always pass with these other options
        if (orthogonalise is False) and (linesearch is False):
            assert_(np.all(np.diff(errors) <= 1e-3))
    
        rec = cp_to_tensor(fac)
        error = T.norm(rec - tensor, 2)
        error /= T.norm(tensor, 2)
>       assert_(error < tol_norm_2,
                f'norm 2 of reconstruction higher = {error} than tolerance={tol_norm_2}')
E       AssertionError: norm 2 of reconstruction higher = nan than tolerance=0.01
```

caglayantuna:
For `validate_tucker_rank`, should we multiply [n_fixed_params](https://github.com/tensorly/tensorly/blob/f7d326125aaa4067887a52d6e84dbd7750945561/tensorly/tucker_tensor.py#L372) with x. It solves sign issue but [assertion issue](https://github.com/tensorly/tensorly/issues/288) remains arbitrarily.

PR #385, opened 2022-02-24
fix normalized sparsity test

caglayantuna:
As it is reported in #310 , `test_constrained_parafac_normalized_sparsity` fails sometimes. I changed `assert_` to `assert_array_almost_equal` and removed tolerance bug. Now, it seems ok according to my tests.


JeanKossaifi:
Excellent, I love improvements to the doc and tests, thanks @caglayantuna!

Issue #48, opened 2018-04-23
Hidden dependency from nose package.


dariosena:
Reproduction:
```
mkvirtualenv tensor ## Python 3 default
make install
pip install pytest
TENSORLY_BACKEND='numpy' pytest -v tensorly
```

Result:
```
8 failed, 38 passed in 1.32 seconds
```

The 8 tests made use of the following import from backend numpy:
```
/numpy/testing/nose_tools/utils.py:71: ImportError
```
Solution:
```
pip install nose
```
Maybe put nose as a dependency (for tests)?

JeanKossaifi:
Fixed by c729db708aa5d9f6126114b5c6443ac410c111bf.

PR #388, opened 2022-03-06
Drop nosetests

yan12125:
It was added as numpy.testing required nose [1], which is no longer the
case for numpy >= 1.15 [2] from 2018 [3].

Testing:
* nosetests fails as some testing code uses pytest-specific features
  (ex: @pytest.mark.parametrize in several files)
* `cd doc && make html` still works fine

Motivation: nose is [broken for Python 3.10](https://github.com/nose-devs/nose/issues/1099). Although there is no real harm, I don't want to install a package known to be not useful.

[1] https://github.com/tensorly/tensorly/issues/48
[2] https://numpy.org/devdocs/release/1.15.0-notes.html
[3] https://pypi.org/project/numpy/#history

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -135,7 +135,7 @@ Running the tests
 
 Testing and documentation are an essential part of this package and all functions come with uni-tests and documentation.
 
-The tests are ran using the `pytest` package (though you can also use `nose`). 
+The tests are ran using the `pytest` package. 
 First install `pytest`::
 
     pip install pytest
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -24,8 +24,6 @@ requirements:
 test:
   requires:
     - pytest
-    # NumPy requires nosetests e.g. for assert_raises.....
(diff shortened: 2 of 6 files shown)


yan12125:
Test failure from `test_parafac[svd-3-5-True-True]` in tensorflow-based test suite:

> E       AssertionError: norm 2 of reconstruction higher = 0.010107140988111496 than tolerance=0.01

Seems another case of https://github.com/tensorly/tensorly/issues/310#issuecomment-908024948

Test failure in mxnet-based test suite

> E           MXNetError: Check failed: shp.ndim() <= 6 (8 vs. 6) : Transpose support at most 6 dimensions

Seems to have failed since https://github.com/tensorly/tensorly/commit/e0831ea67e4748e52c8f07148c489aa70325b966

JeanKossaifi:
Fabulous, thank you @yan12125, this is something that was long overdue! :)
Merging.

PR #389, opened 2022-03-15
default axis for tensorflow concatenate

caglayantuna:
Tensorly concatenate function has a default axis in [description](http://tensorly.org/stable/modules/generated/tensorly.concatenate.html) but tensorflow [concat](https://www.tensorflow.org/api_docs/python/tf/concat) function doesn't have a default input. This PR fixes this issue.


JeanKossaifi:
Great, thanks @caglayantuna !
Merging.

PR #390, opened 2022-03-16
Constrained cp class

caglayantuna:
I realized that contrained parafac has no class in Tensorly. I know that classes are experimental for now but it could be useful for the future developments. This PR adds `ConstrainedCP` class to `_constrained_cp` file with related changes in init and test.


JeanKossaifi:
Great, thanks @caglayantuna! Can you also add it to the doc (`api`) ?

caglayantuna:
Done.

JeanKossaifi:
Looks great, thank you @caglayantuna!

PR #396, opened 2022-03-31
Return errors for tucker

caglayantuna:
This PR adds `return_errors` option to tucker decomposition like other decomposition methods and improves the `non_negative_tucker` doc which has no return_errors although function itself does.


JeanKossaifi:
Great, thank you, merging!

PR #397, opened 2022-03-31
JAX backend - v > 0.3.0

JeanKossaifi:
Update the backend for latest version of JAX and new way of indexing.

### DIFF
--- a/tensorly/backend/jax_backend.py
+++ b/tensorly/backend/jax_backend.py
@@ -19,7 +19,6 @@ import copy
 from .core import Backend
 
 
-
 class JaxBackend(Backend, backend_name='jax'):
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
It seems that in the monotonicity constraints we use index_update on numpy arrays instead of tensors in the current backend - I think that's what causing the issue in the JAX backend post update @caglayantuna @cohenjer

caglayantuna:
Jax copy changes `DeviceArray` to `ndarray`. Could it be the reason? 
```
import tensorly as tl
tl.set_backend('jax')
a = tl.zeros([3, 2])
c = tl.copy(a)
print(isinstance(a, np.ndarray))
print(isinstance(c, np.ndarray))
```
False 
True

[Here](https://github.com/tensorly/tensorly/blob/main/tensorly/backend/jax_backend.py), instead of copy.copy, `jax.numpy.array` could be better.

JeanKossaifi:
Good catch @caglayantuna - indeed, they seem to have an issue for that: https://github.com/google/jax/issues/3473
Not sure about the best way to deal with this: should we just use `tl.tensor( array.copy(), **tl.context(arrray))`?

Issue #394, opened 2022-03-29
Trig and inverse-trig functions?


j6k4m8:
It looks like there's already partial support for elementwise trig operations ([sin, cos](https://github.com/tensorly/tensorly/blob/main/tensorly/backend/numpy_backend.py#L79), and others). 

I am wondering if it would be alright for me to PR an addition to include:

### Trig functions

* sin
* cos
* tan

### Reciprocal functions

* csc
* sec
* cot

### Hyperbolic functions

* sinh
* cosh
* tanh

### Inverse functions

* arcsin
* arccos
* arctan

JeanKossaifi:
Feel free to add functions you need. Ideally they should be supported by most backends.

j6k4m8:
Fixed in #398 :)

PR #398, opened 2022-04-01
Add trig functions and constants

### LINKED ISSUE #395: Numerical constants?
I wonder if the authors of this project have considered adding numerical constants (π, e) to the library.

#### Argument in favor

Right now, as an author of another developer-facing library, I can remain un-opinionated about my users' backend of choice when using functions:

```python
import tensorly as tl

tl.log(tl.tensor(...))
```

But as soon as I want to include numerical constants, I have to pull them from a specific library:

```python
import tensorly as tl
import numpy as np

tl.log(np.e ** tl.tensor(...))
#      ^^^^
```

This means that I can't operate on numerical constants without also having a backend library as a dependency for my library. Adding these constants to tensorly would mitigate this.

#### Argument against

Adding these constants might mean that certain operations will return unexpected results if there are tiny differences between backends' definitions of the constants. For example,

```python
tl.pi - np.pi
```

might be nonzero, if `tl.pi` uses a definition from another library.

-----

Curious to hear your thoughts on this!

j6k4m8:
This fixes #395 and #394!


cohenjer:
Looks good to me, thanks for adding these !

j6k4m8:
Indeed — it doesn't appear that the issues are with this new PR code, but I'm happy to wait to merge this as the other issues work out, if you prefer (and I will see if I can figure out an easy fix for any of them!)

JeanKossaifi:
> it doesn't appear that the issues are with this new PR code

Yes, I'm fixing the JAX issue in the PR you mentioned and MXNet issue should be fixed whenever they switch to 2.0, we should use the nightly build until then.

j6k4m8:
Awesome! In that case, I think this PR is ready for final code review and merging (i.e., all issues with the actual trig functions and constants are ironed out, I believe)

JeanKossaifi:
Looks good to me, thanks @j6k4m8, merging!

PR #404, opened 2022-05-04
Permute factors api

caglayantuna:
As we discussed during the previous [PR](https://github.com/tensorly/tensorly/pull/380), we add cp_permute_factors function to api and advertise this function with an example in gallery since we couldn't find a suitable place in the user guide. Also, I removed redundant rank computation from the function.


JeanKossaifi:
Looks good, thanks @caglayantuna! Merging.

PR #405, opened 2022-05-10
Update applications examples

borcuttjahns:
Remove Pandas dependency for IL-2 CP decomposition example.


borcuttjahns:
This should fix the COVID applications build as well as remove the pandas dependency for the IL-2 application example. Let me know if you would like any other changes - @JeanKossaifi @aarmey

aarmey:
This looks reasonable to me. Is seaborn already used in the docs? If not, then that would need to be removed, too.

JeanKossaifi:
Looks good to me, merging, thanks @borcuttjahns @aarmey!

PR #406, opened 2022-05-13
Tensor permutation fix for Jax

### LINKED ISSUE #402: Jax test for permute factors
#### Describe the bug
After recent updates, testing with jax backend fails when indexing with lists in [cp_permute_factors](https://github.com/tensorly/tensorly/blob/main/tensorly/cp_tensor.py). Error occurs when updating weights in line 735. Related issue can be found [here](https://github.com/google/jax/issues/4564).

```python
import tensorly as tl
tl.set_backend('jax')
indices = [1, 2, 0, 3]
array = tl.tensor([0, 1, 2, 3])
new_array = array[indices]
```
TypeError: Using a non-tuple sequence for multidimensional indexing is not allowed; use `arr[array(seq)]` instead of `arr[seq]`. 

#### Expected behavior
If input indices is an array, it works. However, I couldn't find how to fix it in a that way in Tensorly.

```python
import tensorly as tl
tl.set_backend('jax')
indices = [1, 2, 0, 3]
array = tl.tensor([0, 1, 2, 3])
new_array = array[tl.tensor(indices)]
print(new_array)
```
[1 2 0 3]

aarmey:
Should fix #402

### DIFF
--- a/tensorly/cp_tensor.py
+++ b/tensorly/cp_tensor.py
@@ -730,8 +730,9 @@ def cp_permute_factors(ref_cp_tensor, tensors_to_permute):
     ref_factors = T.concatenate(ref_cp_tensor.factors)
     for i in range(n_tensors):
         _, col = congruence_coefficient(ref_factors, T.concatenate(tensors_to_permute[i].factors))
+        col = T.tensor(col, dtype=T.int64)
         for f in range(n_factors):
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
Thanks @aarmey, that's great! Merging.

PR #407, opened 2022-05-14
Moves tf to the numpy interface

aarmey:
The tensorflow numpy interface now seems to work, so I'm interested in feedback on this. One consideration is this interface only came out a version or two ago, and so we might deprecate the old interface first before moving up the required version of tensorflow.

### DIFF
--- a/tensorly/backend/tensorflow_backend.py
+++ b/tensorly/backend/tensorflow_backend.py
@@ -1,10 +1,9 @@
 try:
     import tensorflow as tf
     import tensorflow.math as tfm
+    import tensorflow.experimental.numpy as tnp
 except ImportError as error:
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
It looks good to me. The big question is whether we keep TensorFlow or just remove it altogether. 
If the new interface works transparently perhaps it's worth keeping just in case some users still rely on it.

aarmey:
The new interface seems to work transparently for me. Perhaps we include it but reevaluate if there are future issues? (I know it has frequently been a headache.)

JeanKossaifi:
Great, I've merged the conflict. Let me know if I've missed anything, otherwise looks good to merge!

PR #408, opened 2022-05-31
2 new dataset

caglayantuna:
This PR adds 2 functions to `data_imports` file in order to load indian pines and kinetic fluorescence dataset from their url. Both functions use `urllib` library which should not be a dependency for tensorly. 

I added necessary information according to Bunch dictionary and related tests to `test_imports` file. Also, I changed the `IL2data` name to `load_IL2data` to be consistent.

### DIFF
--- a/tensorly/datasets/data_imports.py
+++ b/tensorly/datasets/data_imports.py
@@ -4,6 +4,10 @@ Load example datasets.
 
 from os.path import dirname
 import numpy as np
+import requests
+import scipy.io
+from zipfile import ZipFile
+from io import BytesIO
 
 
 class Bunch(dict):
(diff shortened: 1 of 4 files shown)


cohenjer:
Just to provide additional context, Indian pines is one of these widely used remotely acquired hyperspectral image, for which tensor decompositions have been used, e.g. recently [here](https://mdpi-res.com/d_attachment/remotesensing/remotesensing-13-02930/article_deploy/remotesensing-13-02930.pdf?version=1627299276).
The kinetic fluorescence dataset is one of Rasmus Bro's 4-way fluorescence dataset which are almost `toy` data for fitting nonnegative tensor factorisation models such as CP, and can easily be used for educational purposes: [link to data description](http://www.models.life.ku.dk/Kinetic_Fluor)

tensorly/datasets/data_imports.py
```diff
@@ -209,4 +213,82 @@ def load_covid19_serology():
...
+                   1650.73, 1660.64, 1670.56, 1680.47, 1690.38, 1700.28, 1710.19, 1720.1, 1730.0, 1739.9, 1749.81,
+                   1759.71, 1769.6, 1779.5, 1789.4, 1799.29, 1809.19, 1819.08, 1953.26, 1963.25, 1973.24, 1983.23,
+                   1993.22, 2003.2, 2013.18, 2023.16, 2033.13, 2043.1, 2053.07, 2063.04, 2073.0, 2082.97, 2092.92,
+                   2102.88, 2112.83, 2122.78, 2132.73, 2142.68, 2152.62, 2162.56, 2172.5, 2182.43, 2192.37, 2202.3,
+                   2260.22, 2270.15, 2232.07, 2241.99, 2251.9, 2261.82, 2271.73, 2281.64, 2291.54, 2301.45, 2311.35,
+                   2321.25, 2331.14, 2341.03, 2350.92, 2360.81, 2370.7, 2380.58, 2390.46, 2400.33, 2410.21, 2420.08,
+                   2429.95, 2439.81, 2449.68, 2459.54, 2469.4, 2479.25, 2489.11, 2498.96]
+
+    return Bunch(
+        tensor=np.array(image, "float"),
```
JeanKossaifi, line 261: Should we further wrap the numpy array in a tensor?

JeanKossaifi (commented):
Looks good to me!
Perhaps we should return a tensorly tensor so it's using the user's current backend.

I wonder whether it would make sense to call the new one fetch_dataset (following sklearn API) since it actually downloads from the web? We could potentially cache the dataset too?

tensorly/datasets/data_imports.py
caglayantuna: Done

caglayantuna:
Thanks for your comments Jean. I changed other datasets output to tl.tensor as well and also changed the names of new ones from `load_` to `fetch_`.

JeanKossaifi:
Thanks @caglayantuna! I think the datasets are already used in examples by @aarmey and others, we should probably update those if we change the names as well.

caglayantuna:
I have already updated `examples/applications/plot_IL2.py` accordingly. I think it is the only example had to be changed.

JeanKossaifi:

Issue #372, opened 2022-02-11
Modifying returns for decompositions without breaking compatibility with older version


cohenjer:
#### Problem description
Right now in tensorly, functions such as `parafac` return a `cp_tensor` when no error return is requested, and a tuple `cp_tensor, err` when error return is requested. While this behavior is not ideal (one must be vary when toogling on and off error returns since the output format change), it works fine currently. However, adding new returns such as time or something provided by a callback function (see #368) is a pain if we need to add new conditions checking which return is required. E.g. what if the user wants to return timings but not error, and then changes his mind ? The output of `parafac` would be something like `cp_tensor, time ` in the first call but `cp_tensor, err` in the second call.

#### Possible solutions

Looking at what other libraries with similar functionalities do (e.g, scipy, scikit-learn), both use object-oriented programming where the outputs of the algorithms are stored as attributes of some class instance (a returns class instance in scipy, and the model class itself in scikit-learn).
While we do have classes for decomposition algorithms in tensorly, using these classes to store the output of algorithms would mean a drastic change in the way we handle returns.

Another solution would be to return a tuple `cp_tensor, dictionary` with the dictionary containing all requested informations (error, time, anything computed by a callback function...). 

In fact I think we should support both options, to facilitate a transition to object-oriented programming.

Issue #368, opened 2022-02-02
Processing time for each iteration


caglayantuna:
We thought that it would be useful if Tensorly has an option to return processing time for each iteration. Decomposition functions could return that if `return_time_it` is True as reconstruction error.  `Time` library can be used as follows:

```python:
time_it = []
    for iteration in range(n_iter_max):
        time_init = time.time()
        for mode in modes:
                   .
                   .
                   .
        time_it.append(time.time() - time_init)
```

What do you think about this new feature? If you agree, I can open a PR.

cohenjer:
Just to add to the discussion, as we discussed with @JeanKossaifi, maybe the output timings should be provided in a dictionary, along with e.g. the reconstruction error. But I'm afraid it would break compatibility with previous version?

aarmey:
@cohenjer @caglayantuna I had another idea after we met about this. Perhaps rather than adding another optional return, we should add a callbacks interface (e.g., see [`scipy.optimize.minimize`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html))? This would provide one solution to return time, error, or anything else. For example, we use tqdm for progress bars in our hand-rolled methods, and a callback would make it possible with Tensorly methods as well. Folks might also want to do things like store parts of the decomposition at each iteration to study convergence. Essentially this allows you to run and save whatever you want on each iteration.

cohenjer:
@aarmey This is a great idea in fact. I have concerns about how to return the results, see #372 I just oppened. Once we figure out how to return things properly, returning time and/or adding callback functions will be very easy.

PR #417, opened 2022-06-29
Draft callback function interface

aarmey:
I'd like to start a discussion on a callback function in each method as this will be helpful to several efforts (e.g., #368, #372). Here, I just implemented the simplest possible setup:

`callback(factors, weights, rec_error)`

However, methods may be able to pass along different information, and the information available may vary with settings (e.g., whether line search is used). An alternative interface would be:

`callback(info)`

where `info` is a dict of all local variables. I currently think the latter option is better. Two questions:

1. Do you agree that a dict, passing along all the local variables, would be best? If not, what would be an ideal alternative?
2. If you think the information passed along should be more restricted or defined, how would you choose what to pass?

### DIFF
--- a/tensorly/decomposition/tests/test_cp.py
+++ b/tensorly/decomposition/tests/test_cp.py
@@ -25,7 +25,13 @@ def test_parafac(linesearch, orthogonalise, true_rank, rank, init, monkeypatch):
     tol_norm_2 = 0.01
     tol_max_abs = 0.05
     tensor = random_cp((6, 8, 4), rank=true_rank, orthogonal=orthogonalise, full=True, random_state=rng)
-    fac, errors = parafac(tensor, rank=rank, n_iter_max=200, init=init, tol=10e-5, random_state=rng, orthogonalise=orthogonalise, linesearch=linesearch, return_errors=True)
+
+    # Callback to record error
+    errors = list()
+    def callback(_, __, rec_error):
+        errors.append(rec_error)
+
+    fac = parafac(tensor, rank=rank, n_iter_max=200, init=init, tol=10e-5, random_state=rng, orthogonalise=orthogonalise, linesearch=linesearch, callback=callback)
 
     # Check that the error monotonically decreases
     # TODO: This doesn't always pass with these other options
--- a/tensorly/decomposition/_cp.py
+++ b/tensorly/decomposition/_cp.py
@@ -16,7 +16,8 @@ from ..cp_tensor import (cp_to_tensor, CPTensor,
 
 # License: BSD 3 clause
 
-def initialize_cp(tensor, rank, init='svd', svd='numpy_svd', random_state=None, normalize_factors=False):
+
+def initialize_cp(tensor, rank, init="svd", svd="numpy_svd", random_state=None, normalize_factors=False):
     r"""Initialize factors used in `parafac`.
 
     The type of initialization is set using `init`. If `init == 'random'` then
@@ -43,15 +44,14 @@ def initialize_cp(tensor, rank, init='svd', svd='numpy_svd', random_state=None, 
     """
     rng = tl.check_random_state(random_state)
 
-    if init == 'random':
+    if init == "random":
         kt = random_cp(tl.shape(tensor), rank, normalise_factors=False, random_state=rng, **tl.context(tensor))
 
-    elif init == 'svd':
+    elif init == "svd":
         try:
             svd_fun = tl.SVD_FUNS[svd]
         except KeyError:
-            message = 'Got svd={}. However, for the current backend ({}), the possible choices are {}'.format(
-                      svd, tl.get_backend(), tl.SVD_FUNS)
+            message = "Got svd={}. However, for the current backend ({}), the possible choices are {}".format(svd, tl.get_backend(), tl.SVD_FUNS)
             raise ValueError(message)
 
         factors = []
@@ -74,13 +74,15 @@ def initialize_cp(tensor, rank, init='svd', svd='numpy_svd', random_state=None, 
 
     elif isinstance(init, (tuple, list, CPTensor)):
         # TODO: Test this
-        try:         
+        try:
             if normalize_factors is True:
-                warnings.warn('It is not recommended to initialize a tensor with normalizing. Consider normalizing the tensor before using this function')
-            
+                warnings.warn(
+                    "It is not recommended to initialize a tensor with normalizing. Consider normalizing the tensor before using this function"
+                )
+
             kt = CPTensor(init)
             weights, factors = kt
-            
+
             if tl.all(weights == 1):
                 kt = CPTensor((None, factors))
             else:
@@ -103,14 +105,14 @@ def initialize_cp(tensor, rank, init='svd', svd='numpy_svd', random_state=None, 
 
 
 def sparsify_tensor(tensor, card):
-    """Zeros out all elements in the `tensor` except `card` elements with maximum absolute values. 
-    
+    """Zeros out all elements in the `tensor` except `card` elements with maximum absolute values.
+
     Parameters
     ----------
     tensor : ndarray
     card : int
         Desired number of non-zero elements in the `tensor`
-        
+
     Returns
     -------
     ndarray of shape tensor.shape
@@ -118,12 +120,12 @@ def sparsify_tensor(tensor, card):
     if card >= np.prod(tensor.shape):
         return tensor
     bound = tl.sort(tl.abs(tensor), axis=None)[-card]
-    
+
     return tl.where(tl.abs(tensor) < bound, tl.zeros(tensor.shape, **tl.context(tensor)), tensor)
 
 
 def error_calc(tensor, norm_tensor, weights, factors, sparsity, mask, mttkrp=None):
-    r""" Perform the error calculation. Different forms are used here depending upon 
+    r"""Perform the error calculation. Different forms are used here depending upon
     the available information. If `mttkrp=None` or masking is being performed, then the
     full tensor must be constructed. Otherwise, the mttkrp is used to reduce the calculation cost.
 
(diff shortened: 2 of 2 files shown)


aarmey:
(Sorry for the noise. I accidentally ran black.)

tensorly/decomposition/tests/test_cp.py
```diff
@@ -25,7 +25,13 @@ def test_parafac(linesearch, orthogonalise, true_rank, rank, init, monkeypatch):
...
     tol_max_abs = 0.05
     tensor = random_cp((6, 8, 4), rank=true_rank, orthogonal=orthogonalise, full=True, random_state=rng)
-    fac, errors = parafac(tensor, rank=rank, n_iter_max=200, init=init, tol=10e-5, random_state=rng, orthogonalise=orthogonalise, linesearch=linesearch, return_errors=True)
+
+    # Callback to record error
+    errors = list()
+    def callback(_, __, rec_error):
+        errors.append(rec_error)
+
+    fac = parafac(tensor, rank=rank, n_iter_max=200, init=init, tol=10e-5, random_state=rng, orthogonalise=orthogonalise, linesearch=linesearch, callback=callback)
```
cohenjer, line 34: Should we add some assert here, such as 
```python
fac, rec_errors = parafac(tensor, rank=rank, n_iter_max=200, init=init, tol=10e-5, random_state=rng, orthogonalise=orthogonalise, linesearch=linesearch, return_errors=True, callback=callback)
assert errors == rec_errors
```

cohenjer (commented):
Great PR @aarmey, since it's quite simple I have not seen anything problematic;
Wondering about a few things:
- It may be non-standard, but should we add a callback call before the first iteration? This way we could also observe things related to the initialization (first error, time when the algorithm starts...)
- I guess this was left out on purpose, but we need a clear documentation for the callback functionality.

aarmey:
@cohenjer good point about having one call before any iterations. I'll add that.

I will fill this out with documentation and an interface definition, then open an actual PR.

aarmey:
@cohenjer good point about having one call before any iterations. I'll add that.

I will fill this out with documentation and an interface definition, then open an actual PR.

cohenjer:
About the question on what argument should the callback function take, in fine indeed we could use a dictionary, but maybe we can go for a simple version of callback first, and then depending on the needs, refine it to have all local variables.

A problem with passing all local variables is maybe that it may make the callback functionality a little harder to use for most users; in particular we would need to document all the local variables accessible with it. But as a researcher I definitively would use it to monitor line-search for instance, or any other algorithmic details.

aarmey:
> About the question on what argument should the callback function take, in fine indeed we could use a dictionary, but maybe we can go for a simple version of callback first, and then depending on the needs, refine it to have all local variables.
> 
> A problem with passing all local variables is maybe that it may make the callback functionality a little harder to use for most users; in particular we would need to document all the local variables accessible with it. But as a researcher I definitively would use it to monitor line-search for instance, or any other algorithmic details.

Keeping it simple seems reasonable. Perhaps for now it only passes in the factorization result in its current state, in the same form as it is eventually returned. I think that rule would work for all methods.

JeanKossaifi:
Thanks for starting the discussion and opening the PR @aarmey. I'd been playing with a version of callbacks in tensorly-torch but ultimately decided against it as it adds quite a bit of complexity to the code and opportunities to break things. 

**The concerns to me are**:
1. Additional runtime. Speed is an important aspect for most users of TensorLy and that would add an additional loop for each iteration, which could quickly become costly.
2. Complexity - this adds additional complexity to the code and could be potentially harder to debug.
3. Limited concrete uses - what would those be besides logging, display information etc? 

**On the pros side**:
1. Easily customizable by users to do what they want

**One alternative** would be to move towards more scikit-learn like API and have the objects be the first class citizens so we can store things like errors etc in the object itself. 

Might be worth having a discussion on the API and update the pros/cons?

cohenjer:
@JeanKossaifi I am not sure I agree with your downsides, but happy to discuss further of course:

- Runtime: If callback is not used, runtime is virtually not affected (only a simple `callback is not None` is added). If callback is used, it is very unlikely that it would add serious time complexity (e.g. print factors or some metric on them, print time...) but if so that is the user's choice?
- Complexity: appart from Black linting noise, this PR is basically 3 lines of code. We could argue that the user could potentially harm the program, but callback has no return so it has no simple way to affect the rest of the algorithm.
- Usage: I can think of a variety of uses, most useful to me is simply printing time; right now we don't have access to time at each iteration, see for instance #372. It can be something like rescaling the error to match the way it is computed in some other software to compare with, to track the sparsity level of factors for tuning e.g. the rank etc. I already see plenty of usage with the naive callback @aarmey implemented!

MarieRoald:
I agree with @cohenjer, and callbacks would be really nice to have in TensorLy. Whenever I've used TensorLy in my research, I have implemented callbacks locally, and I believe getting it as an official part of the API would be very useful.

This very minimal way of implementing callbacks looks nice. It doesn't change the outputs of the function, will not change runtime unless the callback is provided and if someone want to log a series of metrics, then it's up to them to create a little callback class that implements the `__call__` method. I'd be happy to write a little example for the gallery if this gets merged 🙂

aarmey:
Thanks for the input all. I completely agree with @cohenjer and @MarieRoald's comments. I wanted to add the specific capabilities that this would enable for me, and the alternative that scikit-learn offers since that is a useful comparison.

Specifically, this would enable a few helpful features to me:

1. A tqdm progress bar, so that I can see how the decomposition is progressing without lots of output noise.
2. Tracing the error and timing on a per-iteration basis for comparing different algorithmic approaches.
3. Tracing the factors evolution at each iteration. This is more for research and methods debugging purposes.

It is absolutely true that scikit-learn does not offer an interface like this, @JeanKossaifi. However, I have used the features above with their methods. scikit-learn takes advantage of the class structure to break methods into sub-steps. Consequently, one can override a method to insert their own tracing, etc. For example, in the CP case, we would break the fitting method into a function for one ALS step, the line search step, etc. Then, one could override just the high-level fit function to insert some other code. The benefit of this approach is it is **much** more extensible and flexible. For example, it would make it much easier to insert alternative line search routines, mix and match ALS algorithms, etc. It's worth considering whether this would be a better long-term direction to avoid having a million different specific implementations. The downside is it would require extensive refactoring.

JeanKossaifi:
Thanks all, I am convinced!

I should say, by complexity I refer to the implementation details we need to figure out, not the actual calls to the callback. 
Mainly:
1. Do we pass a dictionary (most powerful) or a fixed set of parameters (that may be confusing as different algorithms have different variables).
2. This will inform this one: what should the callback return (if anything). A dict input would enable modifying variables inplace.
3. Where do we call the callbacks - before the computation? After? We could potentially have multiple levels but that makes the interface more complex.  Enabling, e.g. line-search with callbacks is also doable but requires update-time callbacks.

I have a preference for passing a dict based on your previous comments. 

@aarmey I agree with all your points too. I do think it might be worth moving to the class structure where it makes sense/useful.

yngvem:
My gut feeling is that the callbacks should only take the decomposition as input. A dict of all local variables sounds like it might make the code too complex. By just passing the decomposition objects, we are free to modify how the decomposition functions work without having to worry about breaking changes w.r.t. other people's callbacks. I think that if you need access to all local variables, then you should probably clone TensorLy and make the modifications yourself. 

I'm also sceptical to allowing the callbacks from making significant changes to the decomposition method. I guess we could allow callbacks to stop the iterations (e.g. by returning True, so no return would do nothing), modify the decomposition object, or both. However, I think the best decision is to not have such behaviour at first, and see if there is a need later, and if so, what need.

JeanKossaifi:
I like it @yngvem - so the object would just call the callbacks with self? That means we only use the callbacks within the OO interface. Makes sense given discussion above with @aarmey if we're moving to object first API.

yngvem:
With decomposition classes, I meant the ``CPTensor`` and ``TuckerTensor`` classes, not the classes that run the decomposition. We should probably get a name for these... And for the callback code, I was thinking of something this:

```python
class SSECallback:
    def __init__(self, tensor):
         self.tensor = tensor
         self.log = []

    def __call__(self, cp_tensor):
        sse = tl.sum((self.tensor - cp_to_tensor(cp_tensor))**2)
        self.log.append(sse)


sse_callback = SSECallback(X)
parafac(X, 3, callback=sse_callback)
```

or, in a functional manner

```python
def create_sse_callback(tensor):
    log = []
    def callback(cp_tensor):
        sse = tl.sum((tensor - cp_to_tensor(cp_tensor))**2)
        log.append(sse)
    return log, callback

log, sse_callback = create_sse_callback(X)
parafac(X, 3, callback=sse_callback)
```

that way, we wouldn't need to make any changes to the API, other than adding a callback argument to the decomposition functions.

cohenjer:
Thanks for the update @aarmey. A couple remarks:
- I don't understand what is going on in the last commit about sparse tensors, why does callback has sparse components in the signature? Is this something important in the sparse case?
- Going back to @yngvem's remark on having callback only take the cp_tensor as input, I like better the previous version when we had the error in the inputs as well,
```python
callback(cptensor, rec_error)
```
With the current version, if the user wants to return errors, he can
   -  use return_errors=True (to be deprecated in favor of callback). This comes at almost no additional computational cost since the error uses partial computations in the updates such as MTTKRP
   - use the callback interface. If callback does not depend on `rec_errors`, it will compute the full error on top of the error being computed, at every iterations. This will add significant computational complexity.
Also, I think all (?) the decomposition models have an underlying cost which is computed at every iterations (except maybe Tucker, the cost is there but the algorithm is not iterative and we don't compute the cost), so we could use the previous syntax without loss of generality.

aarmey:
@cohenjer you read my mind. I was going to post about both of these issues:

- The rule I have been following is that the inputs to the callback should match the return signature. When sparse components are included, that is also returned. I thought about leaving it out, but it is also required in the error calculation.
- I agree about including the unnormalized error as a second argument. This has the added benefit that we can provide a standard class for tracking the error and timing regardless of method.

tensorly/decomposition/tests/test_cp.py
aarmey: I've opted to not add this assert as I expect we are getting rid of rec_errors (to ensure the method functions have a more consistent return structure).

Issue #237, opened 2021-03-08
partial_tucker: reorder arguments


JeanKossaifi:
Currently modes is a positional argument while rank is optional, this can be confusing and not consistent with other decompositions. 

https://github.com/tensorly/tensorly/blob/d8e90a600c3c26983777e971ce85ccba071cdfaf/tensorly/decomposition/_tucker.py#L16

PR #418, opened 2022-06-30
reorder modes and ranks in partial_tucker

caglayantuna:
This PR is related to consistency issue in partial_tucker raised in #237 . I reordered modes and rank arguments in signature, doc and tests.

### DIFF
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -76,7 +76,7 @@ def initialize_tucker(tensor, rank, modes, random_state, init='svd', svd='numpy_
     return core, factors
     
 
-def partial_tucker(tensor, modes, rank=None, n_iter_max=100, init='svd', tol=10e-5,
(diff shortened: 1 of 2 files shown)


tensorly/decomposition/_tucker.py
```diff
@@ -76,7 +76,7 @@ def initialize_tucker(tensor, rank, modes, random_state, init='svd', svd='numpy_
     return core, factors
     
 
-def partial_tucker(tensor, modes, rank=None, n_iter_max=100, init='svd', tol=10e-5,
```
aarmey, line 79: Previously rank had a default of None, but will now have no default. I think this also better matches other signatures, but want to confirm this is what you intended.

aarmey (approved):
This looks good to me. See my comment to confirm one thing.

JeanKossaifi:
Great, looks good to me to, thanks @caglayantuna! Fixes #237

PR #419, opened 2022-06-30
Remove descending argument in sorting functions

aarmey:
This removes the descending argument from the sorting functions so that they are consistent with the NumPy interface. Descending sorting was only used in two places, so these have been updated with an additional call to `np.flip`. I know this can easily turn into a game of code golf. However, I think users (1) will expect a NumPy-like interface as this has become standard and (2) simplifying the backend code will make it easier to navigate changes like a more consistent SVD backend. Once we can switch to the NumPy-like interface to PyTorch and tensorflow we should need only a minimal amount of code for the basic backend functions.

### DIFF
--- a/tensorly/backend/core.py
+++ b/tensorly/backend/core.py
@@ -884,7 +884,7 @@ class Backend(object):
         raise NotImplementedError
 
     @staticmethod
-    def sort(tensor, axis, descending = False):
+    def sort(tensor, axis):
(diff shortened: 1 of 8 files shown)


JeanKossaifi:
I agree, thanks @aarmey!

cohenjer (commented):
Looks good to me as well, in particular the proximal part !

PR #420, opened 2022-07-03
Add assert allclose and tests for test utils

MarieRoald:
This pull request adds assert_allclose to `tensorly.testing`. currently, TensorLy supports `assert_array_almost_equal`, which uses `numpy.testing.assert_almost_equal`. However, NumPy now recommends using `assert_allclose` (see [numpy.testing.assert_almost_equal — NumPy v1.22 Manual](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html)) instead, so it would be nice if TensorLy had support for this too.

I also added some simple tests for the `tensorly.testing` module.

### DIFF
--- a/tensorly/testing.py
+++ b/tensorly/testing.py
@@ -14,8 +14,19 @@ def assert_array_almost_equal(a, b, *args, **kwargs):
     np.testing.assert_array_almost_equal(T.to_numpy(a), T.to_numpy(b), *args, **kwargs)
 
 
-def assert_allclose(actual, desired, *args, **kwargs):
-    np.testing.assert_allclose(T.to_numpy(actual), T.to_numpy(desired), *args, **kwargs)
+def assert_allclose(actual, desired, rtol=1e-07, atol=0, equal_nan=True, err_msg='', verbose=True):
+    """Check if two arrays are equal up to a given relevant and absolute tolerance.
+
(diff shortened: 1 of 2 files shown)


caglayantuna:
This is nice feature to have. Maybe, we can specify that  `assert_allclose` has `rtol` and `atol` in signature. It is the main difference and not easy to understand before looking numpy document. Also, we can change `a` and `b` to `actual` and `desired` in other functions to be consistent. 

Do you consider to change all `assert_array_almost_equal` to `assert_allclose` in Tensorly or is it for future test functions?

MarieRoald:
I was thinking that changing the assert_array_almost_equal into assert_allclose is the way to go in the future. It could be a nice issue to mark as a "good first issue", so I'll make that issue once this PR is merged 😀

Good idea of making the signature more explicit! I made the signature explicitly equal to the NumPy signature and added a link to the NumPy documentation in the docstring. I'm hesitant to update the other parts of the testing module for three reasons:
 * The NumPy signatures differ for the different assert functions, `assert_almost_equal`, for example, has `x` and `y` instead of `actual` and `desired`, and we might want to be consistent with NumPy.
 * Such a change should probably be in a separate pull request to make the commit history more transparent.
 * We should probably discuss whether `assert_almost_equal` should be deprecated, and if so, I don't think it's worth it to modify it.

JeanKossaifi:
Perfect, looks great, thanks @MarieRoald! Merging and will let you create the issue!

Issue #308, opened 2021-08-23
Adjust test_svd for randomized SVD


JeanKossaifi:
Currently the test for backend implementation of SVD also runs for the randomized_svd which has a lower accuracy - we can adjust the precision for the randomized version specifically.

caglayantuna:
Following code can be added [here](https://github.com/tensorly/tensorly/blob/7d7c796e1dfa7233fbfc77adb9a16d629d109a80/tensorly/tests/test_backend.py#L135) 

```python
if name == "randomized_svd":
       decimal = 2
else:
       decimal = 3
.
.
.
assert_array_almost_equal(np.abs(S), T.abs(fS), decimal=decimal ...
```

If you thought something like this, I can submit a PR.

Issue #287, opened 2021-06-29
Issue with test_svd


JeanKossaifi:
`test_backend.test_svd` seems to occasionally fail, at least with the MXNet backend, see e.g. https://github.com/tensorly/tensorly/runs/2945783934

Relevant traceback:
```
=================================== FAILURES ===================================
___________________________________ test_svd ___________________________________

    def test_svd():
        """Test for the SVD functions"""
        tol = 0.1
        tol_orthogonality = 0.01
    
        for name, svd_fun in T.SVD_FUNS.items():
            sizes = [(100, 100), (100, 5), (10, 10), (10, 4), (5, 100)]
            n_eigenvecs = [90, 4, 5, 4, 5]
    
            for s, n in zip(sizes, n_eigenvecs):
                matrix = np.random.random(s)
                matrix_backend = T.tensor(matrix)
                fU, fS, fV = svd_fun(matrix_backend, n_eigenvecs=n)
                U, S, V = svd(matrix)
                U, S, V = U[:, :n], S[:n], V[:n, :]
    
>               assert_array_almost_equal(np.abs(S), T.abs(fS), decimal=3,
                    err_msg='eigenvals not correct for "{}" svd fun VS svd and backend="{}, for {} eigenenvecs, and size {}".'.format(
                            name, tl.get_backend(), n, s))

tensorly/tests/test_backend.py:146: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = array([50.15259113,  5.49374303,  5.39877779,  5.32205362,  5.09081317,
        5.03235189,  4.93070913,  4.8058477 , ...691386,  0.81209131,  0.75197604,  0.72356455,
        0.65506754,  0.63010537,  0.57366523,  0.54597993,  0.48835856])
b = array([50.15259113,  5.49374303,  5.39877779,  5.32205362,  5.09081317,
        5.03235189,  4.93070913,  4.8058477 , ...9059,  0.75193308,  0.72347523,
        0.65506358,  0.63002172,  0.57352005,  0.54564354,  0.48644761], dtype=float64)
args = ()
kwargs = {'decimal': 3, 'err_msg': 'eigenvals not correct for "randomized_svd" svd fun VS svd and backend="mxnet, for 90 eigenenvecs, and size (100, 100)".'}

[...]

PR #421, opened 2022-07-08
change decimal for randomized_svd

caglayantuna:
This PR is related to issues #308 and #287. Basically, svd test fails sometimes because of the randomized svd. As @JeanKossaifi suggested, I defined a lower precision specifically for it and ran the test 1000 times with Mxnet backend without failing.


aarmey (approved):
This looks reasonable to me!

Issue #422, opened 2022-07-12
Testing for complex tensors


aarmey:
See #298. Several of the tests for methods should be extended to the complex cases. Perhaps a first step to accomplishing this is to extend the random tensor methods to be able to construct random tensors with complex elements.

aarmey:
Done for the base PARAFAC method. We could consider adding this for the other methods as well.

PR #423, opened 2022-07-12
Add testing for complex values in CP

aarmey:
Here is a basic test setup for complex valued tensors. I will add comments documenting the failure cases.

Addresses #422


tensorly/decomposition/tests/test_cp.py
```diff
@@ -20,13 +20,14 @@
 @pytest.mark.parametrize("init", ['svd', 'random'])
 @pytest.mark.parametrize("normalize_factors", [False, True])
 @pytest.mark.parametrize("random_state", [1, 1234])
-def test_parafac(linesearch, orthogonalise, true_rank, rank, init, normalize_factors, random_state, monkeypatch):
+@pytest.mark.parametrize("complex", [False, True])
+def test_parafac(linesearch, orthogonalise, true_rank, rank, init, normalize_factors, random_state, complex, monkeypatch):
```
aarmey, line 24: This fails under two circumstances. First, the line search routine fails to return an acceptable fit with complex values. I don't see a reason this shouldn't work, so there's possibly an error somewhere. Second, in the complex case the error is not strictly decreasing (but eventually arrives at an ok fit in the absence of line search). Perhaps the error calculation could use a second look.

cohenjer: Interesting; I don't recall anyone trying out line search with complex numbers, perhaps this should be disabled by default since no research paper (as far as I know) as tackled this issue?

For the error computation, I would have to properly check but it is possible we are missing a conjugate in the mttkrp in 
https://github.com/tensorly/tensorly/blob/ce2e5e4bdb1a3d924b26c4f3cd9485e4388f2fd9/tensorly/decomposition/_cp.py#L212

yngvem (commented):
It seems like I forgot to submit my review last week, so some of my comments may be outdated... I cannot check if that's the case now, as I'm out travelling, but hopefully they should be helpful :)

tensorly/decomposition/_cp.py
```diff
@@ -214,6 +214,9 @@ def error_calc(tensor, norm_tensor, weights, factors, sparsity, mask, mttkrp=Non
                 tl.abs(norm_tensor**2 + factors_norm**2 - 2 * iprod)
             )
 
+            errchk, _, _ = error_calc(tensor, norm_tensor, weights, factors, sparsity, mask, mttkrp=None)
```
aarmey, line 217: There is a more general issue here. I added this check—whenever the mttkrp is used it checks this against the error found without using it. The result always matches with a rank 1 problem, but not with *any* rank >1 problems. So, this seems to not be restricted to complex problems—it just happens to show up in that case.

tensorly/cp_tensor.py
```diff
@@ -682,9 +682,7 @@ def cp_norm(cp_tensor):
     """
     _ = _validate_cp_tensor(cp_tensor)
     weights, factors = cp_tensor
-    norm = 1
-    for factor in factors:
-        norm *= T.dot(T.transpose(factor), factor)
+    norm = T.prod([T.dot(T.transpose(f), f) for f in factors])
```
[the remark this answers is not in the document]
aarmey: Good catch. Thanks!

tensorly/cp_tensor.py
```diff
@@ -724,9 +724,7 @@ def cp_norm(cp_tensor):
     """
     _ = _validate_cp_tensor(cp_tensor)
     weights, factors = cp_tensor
-    norm = 1
-    for factor in factors:
-        norm *= T.dot(T.transpose(factor), factor)
+    norm = T.prod([T.dot(T.transpose(f), T.conj(f)) for f in factors], axis=0)
```
aarmey, line 727: Found another missing conjugate call here.

tensorly/decomposition/tests/test_cp.py
```diff
@@ -37,7 +37,8 @@
 @pytest.mark.parametrize("true_rank,rank", [(1, 1), (3, 4)])
 @pytest.mark.parametrize("init", ["svd", "random"])
 @pytest.mark.parametrize("normalize_factors", [False, True])
-@pytest.mark.parametrize("random_state", [1, 1234])
+@pytest.mark.parametrize("random_state", [1, 1236])
```
aarmey, line 40: I changed the random seed here because the random initialization is pathologically bad. It works fine (and I directly checked the MTTKRP error calculation) but ALS gets trapped without line search or orthogonalization.

JeanKossaifi:
Thanks @aarmey! I love how you also casually improved the quality of life for the whole library in the process by speeding up the tests for JAX and TF!

tensorly/decomposition/tests/test_cp.py
```diff
@@ -491,16 +516,18 @@ def test_non_negative_parafac_hals_one_unconstrained():
 @pytest.mark.xfail(tl.get_backend() == "tensorflow", reason="Fails on tensorflow")
 def test_sample_khatri_rao():
     """Test for sample_khatri_rao"""
-
     rng = tl.check_random_state(1234)
     t_shape = (8, 9, 10)
     rank = 3
-    tensor = T.tensor(rng.random_sample(t_shape) + 1)
-    weights, factors = parafac(tensor, rank=rank, n_iter_max=120)
+    _, factors = random_cp(t_shape, rank, full=False)
```
aarmey, line 522: This is an unrelated change to speed up the slower backend tests by ~20 min.

aarmey:
If someone can give this a last look, it should be ready and all the tests are passing now.

tensorly/decomposition/_cp.py
```diff
@@ -209,7 +209,7 @@ def error_calc(tensor, norm_tensor, weights, factors, sparsity, mask, mttkrp=Non
 
             # mttkrp and factor for the last mode. This is equivalent to the
             # inner product <tensor, factorization>
-            iprod = tl.sum(tl.sum(mttkrp * factors[-1], axis=0))
+            iprod = tl.sum(tl.sum(mttkrp * tl.conj(factors[-1]), axis=0))
```
JeanKossaifi, line 212: Why does it have to be the conjugate here?

JeanKossaifi (approved):
Awesome, looks good to me!

tensorly/decomposition/_cp.py
aarmey: @yngvem had this insight, so he might be able to explain. I confirmed that the results are indeed wrong without this.

tensorly/decomposition/tests/test_cp.py
aarmey: Interesting idea. It should be clear by looking at `cp_lstsq_grad()`. It'd be too expensive to run every iteration, but is certainly something you could run at the end to look for issues.

yngvem:
@aarmey @JeanKossaifi I wasn't able to find the button for submitting my comments after it was merged, so I'm answering the mttkrp-question here :)

Since `iprod` is supposed to be the inner product of the estimated and true tensor, it should be equal to 
$\sum\_{ijk} \mathcal{T}\_{ijk} \hat{\mathcal{T}}\_{ijk}^\*$ or $\sum\_{ijk} \mathcal{T}\_{ijk}^\* \hat{\mathcal{T}}\_{ijk}$, where the $^\*$ denotes conjugation. We have already computed $\sum\_{jk} \mathcal{T}\_{ij:} \hat{\mathcal{T}}\_{ij:}^\*$ in the mttkrp, so we need to conjugate the factor matrix of the last mode here to be consistent. 

I think we could conjugate the mttkrp instead of the factor matrix, but I would have to check. Anyways, the point is that we need at least one conjugate on that line for there to be a complex inner product

JeanKossaifi:
Thanks for the clear answer @yngvem! Given that you went through the effort of writing it so nicely, I'd say let's add it to the docstring (in notes) or user guide, what do you think?

PR #424, opened 2022-07-12
Better initialization of CMTF ALS

aarmey:
This PR improves the initialization of CMTF by using the concatenation of both datasets for mode 0 initialization. This adjustment allows the error checks to consistently pass (ran it successfully 200 times). I also ran black linting, added a tolerance argument, and added a convergence failure warning.


aarmey:
> Nice PR @aarmey, the new init and reverse mode swipe makes lot of sense. Looks good to me!
> 
> No idea why the validate_tucker_rank test is failing, this is likely unrelated?

Yes. This is documented in issue #310.

Issue #227, opened 2021-02-26
A clearer interface to SVD options


aarmey:
I think we may want to use the opportunity to add some documentation, perhaps in the user_guide, either in the [backend](https://github.com/tensorly/tensorly/blob/main/doc/user_guide/backend.rst) section or in a new page? We can there explain the difference between the three provided interfaces. 

I'd really rather avoid converting to NumPy as much as possible as it breaks the computational graph in some backends and causes surprising issues (e.g. if running the method on GPU).

On the API side, I'm wondering whether we want to provide a function `partial_svd` (or even just `svd`) and an additional `implementation` (or `method`?) :

```python
def partial_svd(matrix, n_eigenvecs, implementation='truncate')
```
where 
```
implementation : {'truncate', 'numpy', 'eish'}

          - truncate : uses the backend's native SVD function and truncates it
          - numpy : uses the efficient Sparse SVD from Scipy to only compute the required eigenvalues
          - symeig : uses the native backend's symeig to leverage the symmetry in the covariance matrix when computing the SVD
```

_Originally posted by @JeanKossaifi in https://github.com/tensorly/tensorly/issues/217#issuecomment-783381531_

JeanKossaifi:
Added in #429.

PR #429, opened 2022-07-17
Move SVD to a common frontend interface

aarmey:
This aims to address #227 by moving the various SVD functions to `tenderly.tenalg`. A new dispatch function, `svd_funs`, should make it easier to apply modifications like resolving sign indeterminacy and transforming to non-negative form regardless of the algorithm applied. In the future, this new function could also handle other features like missing value imputation.

I also made some adjustments to further unify the backends now that they all are moving to a more consistent NumPy interface.


tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,379 @@
...
+def svd_checks(matrix, n_eigenvecs=None):
+    """Runs common checks to all of the SVD methods.
+
+    Parameters
+    ----------
+    matrix : 2D-array
+    n_eigenvecs : int, optional, default is None
+        if specified, number of eigen[vectors-values] to return
+
+    Returns
```
aarmey, line 165: I still need to add documentation here for the return values.

aarmey:
This is ready for review when someone has a chance to look.

tensorly/__init__.py
```diff
@@ -87,25 +88,19 @@
...
-    randomized_range_finder,
     log,
     log2,
     exp,
     sin,
     cos,
     tan,
-    asin,
-    acos,
-    atan,
```
JeanKossaifi, line 108: Why did you remove those (as well as the hyperbolic versions down)?

doc/development_guide/contributing.rst
```diff
@@ -39,7 +39,7 @@ To contribute code to the TensorLy code-base, you must ensure compatibility with
    This means you should only use TensorLy functions, never directly a function from the backend
    e.g. use ``tl.mean``, **not** ``numpy.mean`` or ``torch.mean``.
 
-To do so, we **only use** functions wrapped in :py:mod:`tensorly.backend`, such as :py:mod:`tensorly.backend.partial_svd`, etc.
+To do so, we **only use** functions wrapped in :py:mod:`tensorly.backend`, such as :py:mod:`tensorly.backend.tensordot`, etc.
```
JeanKossaifi, line 42: Love that you also adjusted the doc! :)

JeanKossaifi (commented):
Thanks @aarmey, this is exactly what we need! My only question is regarding the asinh/arcsinh -- did you remove them on purpose?

One issue I have about the way I organized the `tenalg` backend (and that is unrelated to this PR) is: how can we easily let each backend override some of those functions (e.g. if a backend has an efficient partial_svd, how can we let the backend override our own implementation).

aarmey:
> One issue I have about the way I organized the `tenalg` backend (and that is unrelated to this PR) is: how can we easily let each backend override some of those functions (e.g. if a backend has an efficient partial_svd, how can we let the backend override our own implementation).

@JeanKossaifi

I remember this came up with the backends having `T.kr()` in addition to `tenalg.khatri_rao()`. For SVD, I was thinking that we could allow a function to be passed in as the algorithm selection, in addition to the string options. That way someone could define a backend function and then pass it in and use all the same features.

As for the rest of tenalg, I'm not sure what would be best. Any ideas that come to mind are too complicated.

By the way, is the tenalg dispatching fully setup? I tried switching to the einsum backend once and couldn't find the documented function for setting the backend.

JeanKossaifi:
> As for the rest of tenalg, I'm not sure what would be best. Any ideas that come to mind are too complicated.

Indeed.


> By the way, is the tenalg dispatching fully setup? I tried switching to the einsum backend once and couldn't find the documented function for setting the backend. (I noticed the einsum backend has no coverage.)

Yes, it is setup but I haven't had time to add full coverage, and some backends had, at least at the time, problematic einsum implementations, so didn't document it fully but you should be able to do `tensorly.tenalg.set_backend('einsum')`.

tensorly/__init__.py
JeanKossaifi: I believe functions are needed in some downstream libraries, at least by @j6k4m8, in which case I'd rather leave them to not break other's code -- @j6k4m8 do you need `asinh` as well as `arcsinh`?

j6k4m8: I don't have an attachment to these names in particular; I added `aXXX` along with `arcXXX` versions to maximize compatibility (since the only cost of supporting them is a string match). [Here's an example](https://github.com/tensorly/tensorly/pull/398/files#diff-f605db052e892b4c76b249d0e5751b9b747bee9d1e5d825adab44e43d833ca18R253) of the lookup for the function calls from that PR.

I don't mind if you want to remove them, although a more appropriate treatment might be to simply point `arccos` (for example) to `acos` in the [base backend implementation](https://github.com/tensorly/tensorly/pull/398/files#diff-3e1da4d92e2d739adfbcf25b91c53643851f4da8d7948b314f4e74ff8d93a748R1548); i.e.:

```python

    def arccos(self, x):
       return self.acos(x)

```

...which obviates the need altogether. But I also don't mind just removing them if you prefer to be numpy-API-exclusive!

aarmey: Thanks @j6k4m8. I'll add wrappers as you suggest for now. We'll have plenty of big changes in the next version already!

aarmey:
@JeanKossaifi this should be ready for another look.

tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,423 @@
+import warnings
+import tensorly as tl
+from .proximal import soft_thresholding
+
+
```
JeanKossaifi, line 5: Should we just re-add license and authors here?

tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,423 @@
...
+        U, S, V = truncated_svd(matrix_reduced, n_eigenvecs=n_eigenvecs)
+        U = tl.dot(Q, U)
+
+    return U, S, V
+
+
+SVD_FUNS = ["truncated_svd", "symeig_svd", "randomized_svd"]
+
+
+def svd_funs(
```
JeanKossaifi, line 348: That's perfect! We could perhaps make the param names more explicit. 

Should we also change the name to make this function the core one? e.g. just svd?

tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,423 @@
...
+    matrix,
+    svd_type="truncated_svd",
```
JeanKossaifi, line 350:
```suggestion
    method="truncated_svd",
```
or algorithm?

tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,423 @@
...
+    n_eigenvecs=None,
+    flip_svd=True,
```
JeanKossaifi, line 352: `flip_sign`? It is clear that it is applied to svd from the context?

tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,423 @@
...
+    non_negative=False,
+    nntype="nndsvd",
```
JeanKossaifi, line 354:
```suggestion
    nn_type="nndsvd",
```
maybe we should also make the param of the nnsvd init clearer in a separate PR :)

tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,423 @@
...
+    u_based_decision=True,
```
JeanKossaifi, line 355: Perhaps we can move this below flip_sign and maybe name it `u_based_flip_sign`?

tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,423 @@
...
+        Whether to make the SVD results non-negative.
+    nntype : str, default is 'nndsvd'
+        Algorithm to use for converting U to be non-negative.
+    u_based_decision : bool, optional, default is True
+        Whether the sign indeterminacy should be resolved using U (vs. V).
+    mask : tensor, default is None.
+        Array of booleans with the same shape as ``matrix``. Should be 0 where
+        the values are missing and 1 everywhere else. None if nothing is missing.
+    svd_mask_repeats : int, default is 5
+        Number of repetitions to apply in missing value imputation.
```
JeanKossaifi, line 383: That might not be clear to users (though it is if they go through the code..)

tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,423 @@
...
+        Contains the left singular vectors of `matrix`
+    """
+
+    if svd_type == "truncated_svd":
+        svd_fun = truncated_svd
+    elif svd_type == "symeig_svd":
+        svd_fun = symeig_svd
+    elif svd_type == "randomized_svd":
+        svd_fun = randomized_svd
+    elif callable(svd_type):
```
JeanKossaifi, line 403: ❤️

JeanKossaifi (commented):
Looks great @aarmey, excited to have this merged! Just left some nit-pick comments about parameter names :)

tensorly/tenalg/svd.py
aarmey: My concern with `svd()` is this is a function name provided by the backends. It could be `svd_dispatch()` or `run_svd()`... or something else? I agree that `svd_funs()` is a little strange because it doesn't *list* the SVD functions...

tensorly/tenalg/svd.py
aarmey: Algorithm is harder to spell. :-)

tensorly/tenalg/svd.py
aarmey: I tried to at least provide a slightly better name. Eventually I'd like to provide some documentation on handling missing values where this could be explained.

tensorly/tenalg/svd.py
JeanKossaifi: That's a good point -- `svd_interface`? I am find with overloading `svd` though, like we do for tensordot. 
The idea is that then `tensorly.fun` or `tensorly.backend.fun` will call the backend while `tensorly.tenalg.fun` will call our custom one. What do you think? We would have to be clear in the documentation either way.

tensorly/tenalg/svd.py
```diff
@@ -0,0 +1,429 @@
...
+SVD_FUNS = ["truncated_svd", "symeig_svd", "randomized_svd"]
+
+
+def svd_funs(
+    matrix,
+    method="truncated_svd",
+    n_eigenvecs=None,
+    flip_sign=True,
+    u_based_flip_sign=True,
+    non_negative=False,
```
JeanKossaifi, line 360: I'm just thinking about this now, sorry to add post-hoc but we could merge the two? 
e.g. `non_negative=None` by default but can take an str argument, e.g. `non_negative="nndsvd"` ?

tensorly/tenalg/svd.py
JeanKossaifi: Looks good!

tensorly/tenalg/svd.py
aarmey: This makes sense. I set it up so that you can still provide `non_negative=True` and get a default selection.

tensorly/tenalg/svd.py
aarmey: I think I would find multiple versions of `svd()` confusing, so I went with `svd_interface()`.

aarmey:
@JeanKossaifi care to take one last look?

JeanKossaifi:
Thanks @aarmey, this is a great quality of life improvement! Merging.

JeanKossaifi:
Ops, looks like we broke the doc :/

JeanKossaifi:
Fixed in 416604fa410938cfb0009d37b75289f03949c69a

Issue #432, opened 2022-07-22
Backend-specific functions


yngvem:
@MarieRoald and I have been thinking about how to implement backend-specific functions in TensorLy. We've currently settled on a decorator as a good approach. Here's a draft of the API we're imagining

```python
def _numpy_func(x):
    return 2


@register_backend_specific('numpy', _numpy_func)
def func(x):
    return x

tl.set_backend("pytorch")
print(func(5))
tl.set_backend("numpy")
print(func(5))
```
```
5
2
```

We also have a skeleton implementation of the decorator:

```python
from functools import wraps
import tensorly as tl

def register_backend_specific(backend, backend_func):
    def wrapper(f):
        @wraps(f)
        def wrapped(*args, **kwargs):
            if tl.get_backend() == backend:
                return backend_func(*args, **kwargs)
            return f(*args, **kwargs)
        return wrapped
    return wrapper
```

But, there are also some things to consider:

 1. How do we ensure that the backend-specific functions are properly documented?
 2. What (if any) changes to the function signature should we accept?

Currently, @MarieRoald and I have landed on not allowing any changes to the signature as the answer to both questions. If an argument is only supported by one backend, then that should be explicitly handled by all implementations, and an appropriate exception should be raised for those that do not support it. This way we ensure that there are no "hidden" features in the backend-specific functions that aren't shown in the docstrings or function signature.

What do you think @JeanKossaifi @aarmey @cohenjer @caglayantuna?

JeanKossaifi:
@yngvem @MarieRoald -- I really like it!

Currently I essentially have been relying on a backend system (both for the actual backend and the `tenalg` module). 
I guess the general theme here is that we make the method dynamically dispatchable. The only concern (both for the backend, tenalg backend and this) is the overhead of doing the fetching each time. 

I've tried to do all the backend related "work" once only, when the backend is first set. The rest becomes just a lookup. I wonder if we could use some (optimized?) implementation, e.g. [singledispatch](https://docs.python.org/3.4/library/functools.html#functools.singledispatch) or the more general version [multipledispatch](https://github.com/mrocklin/multipledispatch) (which is not in the stdlib).

aarmey:
[...]

PR #434, opened 2022-07-26
Decorator for backend specific implementations

MarieRoald:
@yngvem and I implemented a decorator for backend specific functions (with tests and documentation) as described in #432

### DIFF
--- a/tensorly/utils/backend_specific.py
+++ b/tensorly/utils/backend_specific.py
@@ -1,0 +1,64 @@
+from functools import wraps
+from .. import get_backend, available_backend_names
+from inspect import getfullargspec
+
+
+def register_backend_specific(backend, backend_func):
+    """A decorator to create a backend specific function.
+
+    Parameters
+    ----------
+    backend : str
+        String that specifies which backend implementation to overload. Must be an entry
+        in the ``tensorly.available_backend_names``-list.
+    backend_func : callable
+        Function implementation for the given backend. The function signature must match
+        the signature of the backend agnostic implementation exactly (including which
+        arguments are keyword-/positional-only and the default arguments) to ensure that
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
I love it @MarieRoald @yngvem!

I wonder whether we would also want to expand it even slightly more for cases where we want to override a function for all backends. A specific usecase I have in mind is einsum -- I've been working on an interface to change the einsum function to i) e.g. use opt-einsum and ii) cache the contraction path. 

Seems like this could be achieved with this PR as well, what do you think?

MarieRoald:
I really like the idea of caching the contraction path/using opt-einsum! Do you have a specific API in mind? Are you for example thinking that the einsum is automatically overriden if opt-einsum is installed?

JeanKossaifi:
No, exactly, I'm thinking it should be optional, and such a decorators might be the way to go? 
e.g. we could have a function like `override_function(existing_fun, new_fun, backend=None)` to apply the decorator, and further provide a facility such as `use_opt_einsum()` or `set_einsum`?

JeanKossaifi:
Another thing: should we add the new function to the list of functions available to that backend when we wrap it? 
Perhaps we can add a Notes section to the docstring too so users know how this is done - that their function is just going the wrapped?

yngvem:
Hmmm. I think we could do something like that :) But I think it would be nice to discuss it in a separate issue/PR to make it easier to see where the different changes were introduced and why we added this feature. I made an issue (https://github.com/tensorly/tensorly/issues/452) for tracking this as best as I understand it, and I think we can start implementing it soon.

JeanKossaifi:
Sounds great - what should we do with this PR? Should we merge as is and iterate?

PR #435, opened 2022-07-28
Tensor PLSR

aarmey:


tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,254 @@
...
+        if (T.ndim(Y) != 1) and (T.ndim(Y) != 2):
+            raise ValueError("Only a matrix (2-mode tensor) Y is allowed.")
+        if T.ndim(Y) == 1:
+            Y = T.reshape(Y, (-1, 1))
+
+        # Mean center the data, record info the object
+        self.X_shape = T.shape(X)
+        self.Y_shape = T.shape(Y)
+        self.X_mean = T.mean(X, axis=0)
+        self.Y_mean = T.mean(Y, axis=0)
```
MarieRoald, line 84: These variable names should end with underscore to follow the scikit-learn naming conventions (https://scikit-learn.org/stable/developers/develop.html#estimated-attributes)

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,254 @@
...
+
+        self.X_factors = [
+            T.zeros((l, self.n_components), **T.context(X)) for l in T.shape(X)
+        ]
+        self.Y_factors = [
+            T.zeros((l, self.n_components), **T.context(X)) for l in T.shape(Y)
+        ]
+
+        ## FITTING EACH COMPONENT
+        for a in range(self.n_components):
```
MarieRoald, line 96: Maybe use a clearer variable name instead of `a`? How about `component` or `component_num`?

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,254 @@
...
+            _X_factors_a = [ff[:, a] for ff in self.X_factors]
+            _Y_factors0_a = Y[:, 0]
+            _old_Y_factors0_a = T.ones(T.shape(_Y_factors0_a)) * T.inf
```
MarieRoald, line 99: These variable names are somewhat unclear to me. There may be some notation I'm missing since I don't have the reference the code is based on. Anyways, these should probably have a bit clearer names.

Some questions:

 * Why the leading underscore? These are local and used later so we don't need to make them hidden.
 * For `_X_factors_a`, could we maybe use something like `current_X_factors`?
 * For `_Y_factors0_a `, is this just a single component vector for the first factor matrix? In that case, maybe it couldbe named `current_coupled_Y_factor` instead?

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,254 @@
...
+
+            for iter in range(self.n_iter_max):
+                Z = T.tensordot(X, _Y_factors0_a, axes=((0,), (0,)))
+                Z_comp = (
+                    tucker(Z, [1] * T.ndim(Z))[1] if Z.ndim >= 2 else [Z / T.norm(Z)]
+                )
```
MarieRoald, line 105: Expand if test to make it clearer what is happening

```suggestion
                if Z.ndim >= 2:
                    Z_comp = tucker(Z, [1] * T.ndim(Z))[1]
                else:
                    Z_comp = [Z / T.norm(Z)]
```

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,254 @@
...
+                for ii in range(Z.ndim):
+                    _X_factors_a[ii + 1] = T.reshape(Z_comp[ii], (-1,))
```
MarieRoald, line 107: Rename `ii` to `mode` and make it more explicit what the code is looping over

```suggestion
                for mode in range(1, X.ndim):  # First mode of Z is collapsed by the above tensordot call
                    _X_factors_a[mode] = T.reshape(Z_comp[mode - 1], (-1,))
```

MarieRoald (commented):
I did a quick first pass on the `cp_plsr.py` file. It looks good so far, just some small comments. I'm not too familiar with CP PLSR, so I think it would be very useful with a reference to the relevant paper/book where the deflation algorithm is from in the class docstring.

I'll try to do another pass later 😀

aarmey:
Thanks @MarieRoald! @cyrillustan, I believe we are using the [N-PLS construction](https://analyticalsciencejournals.onlinelibrary.wiley.com/doi/10.1002/%28SICI%291099-128X%28199601%2910%3A1%3C47%3A%3AAID-CEM400%3E3.0.CO%3B2-C), right? We are also generalizing this to work for >3-mode X. We will add an explanation to the docstring.

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,255 @@
...
+            if T.ndim(Y) == 1:
+                Y = T.reshape(Y, (-1, 1))
+            if self.Y_shape[1:] != T.shape(Y)[1:]:
+                raise ValueError(
+                    f"Training Y has shape {self.Y_shape}, while the new Y has shape {T.shape(Y)}"
+                )
+
+            Y -= self.Y_mean
+            Y_scores = T.zeros((T.shape(Y)[0], self.n_components))
+            for a in range(self.n_components):
```
aarmey, line 224: @cyrillustan why does Y_scores have to be solved component-by-component if it is at most 2D? Shouldn't it just be a simple `Y_scores = T.dot(Y, Y_factors)` like in PLS?

cyrillustan: Sure there is probably a way to do it without solving by component. I just simply copied this code from above.

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,259 @@
...
+        ## FITTING EACH COMPONENT
+        for component in range(self.n_components):
+            comp_X_factors = [ff[:, component] for ff in self.X_factors]
+            comp_Y_factors_0 = Y[:, 0]
+            old_comp_Y_factors_0 = T.ones(T.shape(comp_Y_factors_0)) * T.inf
+
+            for iter in range(self.n_iter_max):
+                Z = T.tensordot(X, comp_Y_factors_0, axes=((0,), (0,)))
+                if Z.ndim >= 2:
+                    Z_comp = tucker(Z, [1] * T.ndim(Z))[1]
```
JeanKossaifi, line 108: random question: if this is just to get the first factor, do we need full Tucker or would just a higher-order SVD across that mode be sufficient?

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,259 @@
...
+                else:
+                    Z_comp = [Z / T.norm(Z)]
+                for mode in range(1, X.ndim):  # First mode of Z is collapsed by the above tensordot call
+                    comp_X_factors[mode] = T.reshape(Z_comp[mode - 1], (-1,))
```
JeanKossaifi, line 112: Would it be clearer to use tensor_to_vec here?

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,259 @@
...
+            X -= CPTensor(
+                (None, [T.reshape(ff, (-1, 1)) for ff in comp_X_factors])
+            ).to_tensor()
+            Y -= T.dot(
+                T.dot(
+                    self.X_factors[0],
+                    T.lstsq(self.X_factors[0], T.reshape(comp_Y_factors_0, (-1, 1)))[0],
+                ),
+                T.reshape(comp_Y_factors_1, (1, -1)),
+            )  # Y -= T pinv(T) u q'
```
JeanKossaifi, line 149: The comment is really welcome here. Would it make sense to add a pinv function to TensorLy, specifying whether we want right or left?

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,259 @@
...
+        return (
+            T.dot(
+                T.dot(T.transpose(scores), estimators), T.transpose(self.Y_factors[1])
+            )
+            + self.Y_mean_
+        )
+
+    def transform(self, X, Y=None):
+        """Apply the dimension reduction.
+        Parameters
```
JeanKossaifi, line 179: missing a new line here, before Parameters, same before returns.

tensorly/regression/cp_plsr.py
```diff
@@ -0,0 +1,259 @@
...
+        estimators = T.lstsq(self.X_factors[0], self.Y_factors[0])[0]
+        return (
+            T.dot(
+                T.dot(T.transpose(scores), estimators), T.transpose(self.Y_factors[1])
+            )
+            + self.Y_mean_
+        )
+
+    def transform(self, X, Y=None):
+        """Apply the dimension reduction.
```
JeanKossaifi, line 178: Would be useful to have a slightly longer, more informative description for users that are going to use this as a black-box :)

tensorly/backend/cupy_backend.py
```diff
@@ -117,4 +117,4 @@ def lstsq(a, b):
 for name in ["svd", "qr", "eigh", "solve"]:
     CupyBackend.register_method(name, getattr(cp.linalg, name))
 
-CupyBackend.regsiter_method("gamma", cp.random.gamma)
+CupyBackend.register_method("gamma", cp.random.gamma)
```
JeanKossaifi, line 120: Thanks for the fix! :)

JeanKossaifi (commented):
Excited to have this in TensorLy @aarmey @cyrillustan, it's looking great!
I left some comments, I'll read up more on the algo itself to review more in detail.

tensorly/regression/cp_plsr.py
aarmey: I could see pinv potentially being useful. Ultimately, we don't need it here. Because we're only using it once, we might as well just use lstsq.

tensorly/regression/cp_plsr.py
```diff
@@ -104,12 +104,10 @@ def fit(self, X, Y):
...
             for iter in range(self.n_iter_max):
                 Z = T.tensordot(X, comp_Y_factors_0, axes=((0,), (0,)))
-                if Z.ndim >= 2:
-                    Z_comp = tucker(Z, [1] * T.ndim(Z))[1]
-                else:
-                    Z_comp = [Z / T.norm(Z)]
+
                 for mode in range(1, X.ndim):  # First mode of Z is collapsed by the above tensordot call
-                    comp_X_factors[mode] = T.reshape(Z_comp[mode - 1], (-1,))
+                    Z_comp = svd_interface(unfold(Z, mode - 1), n_eigenvecs=1)[0]
```
cyrillustan, line 113: @aarmey I just found out that the change from `tucker()` to `svd_interface()` actually changes the result of decomposition. I'm not sure what `svd_interface()` does, but running a proper HOOI (which I assume `tucker()` does) is an integral part of this algorithm. Can you help me check on this function? Thank you!

aarmey: You're right. Since you only need one mode, take a look at partial_tucker. That should be the right choice.

cyrillustan: Hmm, it doesn't seem like there is too much performance difference between `tucker()` and `partial_tucker()` though, as the former pretty much just calls the latter.

aarmey: Performance is not the only consideration here. This code should be clear to read for others trying to understand the algorithm. Readability will also help others verify that the code is correct themselves.

Actually, reading more I'm surprised that the results are different when using the SVD of the unfolded tensor. This is exactly what Tucker uses for initialization, and my understanding is that the subsequent iteration within Tucker doesn't do all that much. Are you perhaps just seeing a sign difference in the results?

cyrillustan: I personally found the readability fine, as `partial_tucker()` actually takes more parameters that we don't need than `tucker()`, but if you find that better I can change it too as they are pretty much equivalent in our case.

For Tucker with ranks difference from the original shape, I think the subsequent iteration does make a difference?  As we don't have orthogonality like in SVD anymore in the tensor case.

Yes, the difference is nontrivial and more than a sign difference.

aarmey: Ok. Use `partial_tucker()` for now. I'll then look into this more.

tensorly/regression/cp_plsr.py
cyrillustan: Turns out that solving all columns together wouldn't give you the same answer. Plugging in the original Y into `transform()`, neither way of solving can recover the Y score at fitting perfectly (just like in regular PLSR), but the one solved component by component is closer.

tensorly/regression/cp_plsr.py
aarmey: @cyrillustan, based on your investigation, change this to CP, then you can resolve this thread. :-)

aarmey:
@JeanKossaifi @MarieRoald this should be ready for your review again!

JeanKossaifi:
I think this broke the doc

aarmey:
@JeanKossaifi, yes, I saw that error. It looks like it is in the COVID example, which this doesn't touch, which is a little strange. Regardless, I'm looking into it.

PR #437, opened 2022-08-03
Warning when fixing last mode

caglayantuna:
This is a small PR for adding missing warning when fixing the last mode in non_negative_tucker_hals. I also extended the test of this function with a fixed_modes option.  

We can support fixing the last mode as it is discussed here #292 but it is better to have consistent warning until then.

### DIFF
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -580,7 +580,8 @@ def non_negative_tucker_hals(
 
     if tl.ndim(tensor) - 1 in fixed_modes:
         warnings.warn(
-            "You asked for fixing the last mode, which is not supported.\n The last mode will not be fixed. Consider using tl.moveaxis()"
+            "You asked for fixing the last mode, which is not supported. The last mode will not be fixed."
+            " Consider using tl.moveaxis() to permute it to another position and keep it fixed there."
         )
         fixed_modes.remove(tl.ndim(tensor) - 1)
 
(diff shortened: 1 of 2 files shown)


tensorly/decomposition/_tucker.py
```diff
@@ -578,6 +578,12 @@ def non_negative_tucker_hals(
     if fixed_modes is None:
         fixed_modes = []
 
+    if tl.ndim(tensor) - 1 in fixed_modes:
+        warnings.warn(
+            "You asked for fixing the last mode, which is not supported.\n The last mode will not be fixed. Consider using tl.moveaxis()"
```
JeanKossaifi, line 583:
```suggestion
            "You asked for fixing the last mode, which is not supported. The last mode will not be fixed."
            " Consider using tl.moveaxis() to permute it to another position and keep it fixed there."
```

JeanKossaifi: Perhaps we should do this ourselves in the future.

JeanKossaifi (commented):
Looks good, thanks @caglayantuna !

tensorly/decomposition/_tucker.py
caglayantuna: We don't do it now because we use last factor for efficient computation. If you agree to transpose data inside the function when the last mode is fixed, I can change this PR.

JeanKossaifi:
Thanks @caglayantuna, merging!

PR #459, opened 2022-11-14
Tenalg einsum backend: cache contraction equation

JeanKossaifi:
Adds a wrapper to cache the contraction equation and just reuse it.

### DIFF
--- a/tensorly/tenalg/einsum_tenalg/caching.py
+++ b/tensorly/tenalg/einsum_tenalg/caching.py
@@ -6,7 +6,7 @@ EINSUM_PATH_CACHE = defaultdict(dict)
 
 def einsum_path_cached(fun):
     def wrapped(key, *args, **kwargs):
-name = f"{fun.__module__}.{fun.__name__}"
+        name = f"{fun.__module__}.{fun.__name__}"
         cache = EINSUM_PATH_CACHE[name]
         try:
             equation = cache[key]
--- a/tensorly/tenalg/einsum_tenalg/caching.py
+++ b/tensorly/tenalg/einsum_tenalg/caching.py
@@ -1,0 +1,20 @@
+EINSUM_PATH_CACHE = dict()
+
+
+def einsum_path_cached(fun):
+    def wrapped(key, *args, **kwargs):
+        name = fun.__name__
+        try:
+            cache = EINSUM_PATH_CACHE[name]
+        except KeyError:
+            EINSUM_PATH_CACHE[name] = dict()
+            cache = EINSUM_PATH_CACHE[name]
+        try:
+            equation = cache[key]
+        except KeyError:
+            equation = fun(*args, **kwargs)
+            cache[key] = equation
+
+        return equation
+
+    return wrapped
--- a/tensorly/tenalg/einsum_tenalg/caching.py
+++ b/tensorly/tenalg/einsum_tenalg/caching.py
@@ -6,7 +6,7 @@ EINSUM_PATH_CACHE = defaultdict(dict)
 
 def einsum_path_cached(fun):
     def wrapped(key, *args, **kwargs):
-        name = fun.__name__
+name = f"{fun.__module__}.{fun.__name__}"
         cache = EINSUM_PATH_CACHE[name]
         try:
             equation = cache[key]
--- a/tensorly/tenalg/einsum_tenalg/caching.py
+++ b/tensorly/tenalg/einsum_tenalg/caching.py
@@ -1,14 +1,13 @@
-EINSUM_PATH_CACHE = dict()
+from collections import defaultdict
+
+
+EINSUM_PATH_CACHE = defaultdict(dict)
 
 
 def einsum_path_cached(fun):
     def wrapped(key, *args, **kwargs):
         name = fun.__name__
-        try:
-            cache = EINSUM_PATH_CACHE[name]
-        except KeyError:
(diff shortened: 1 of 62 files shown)


cohenjer:
Thanks for adding this @JeanKossaifi, I will try to look at it asap but that's slightly outside my expertise.

tensorly/tenalg/einsum_tenalg/caching.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def einsum_path_cached(fun):
+    def wrapped(key, *args, **kwargs):
+        name = fun.__name__
+        try:
+            cache = EINSUM_PATH_CACHE[name]
+        except KeyError:
+            EINSUM_PATH_CACHE[name] = dict()
+            cache = EINSUM_PATH_CACHE[name]
```
yngvem, line 11: Is there a reason why you didn't use defaultdict here?

```suggestion
from collections import defaultdict


EINSUM_PATH_CACHE = defaultdict(dict)


def einsum_path_cached(fun):
    def wrapped(key, *args, **kwargs):
        name = fun.__name__
        cache = EINSUM_PATH_CACHE[name]
```

tensorly/tenalg/einsum_tenalg/caching.py
```diff
@@ -0,0 +1,20 @@
...
+        try:
+            equation = cache[key]
+        except KeyError:
+            equation = fun(*args, **kwargs)
+            cache[key] = equation
```
yngvem, line 16: Is there a reason why you use `try`-`except` instead of checking if the key is in the cache beforehand? Is that more efficient?

tensorly/tenalg/einsum_tenalg/caching.py
```diff
@@ -0,0 +1,20 @@
+EINSUM_PATH_CACHE = dict()
+
+
+def einsum_path_cached(fun):
+    def wrapped(key, *args, **kwargs):
+        name = fun.__name__
```
yngvem, line 6: Should we use

```suggestion
name = f"{fun.__module__}.{fun.__name__}"
```

instead to make sure we don't introduce any bugs if we have two functions with the same name?

tensorly/tenalg/einsum_tenalg/generalised_inner_product.py
```diff
@@ -1,11 +1,12 @@
...
-import numpy as np
+from .caching import einsum_path_cached
 
 # Author: Jean Kossaifi
 # License: BSD 3 clause
 
 
-def inner(tensor1, tensor2, n_modes=None):
+@einsum_path_cached
+def inner_path(tensor1, tensor2, n_modes=None):
```
yngvem, line 9: Will that not set ``key=tensor1``? will that give the correct behaviour for the cache?

tensorly/tenalg/einsum_tenalg/caching.py
JeanKossaifi: That's a good idea - do you know how it does speed-wise? I was focused on trying to minimize any overhead. I guess it's not a big deal as that call will happen once anyway for each new fun.

yngvem (commented):
I did a quick pass on some of the files (nothing big, didn't download and run the code) :)

tensorly/tenalg/einsum_tenalg/caching.py
JeanKossaifi: Yes, getting an item is just O(1) while checking for existence is O(n) and the overhead of the try/except is minimal and only happens at the first call so the overall cost is just O(1).

tensorly/tenalg/einsum_tenalg/caching.py
JeanKossaifi: That's a good idea.

tensorly/tenalg/einsum_tenalg/generalised_inner_product.py
JeanKossaifi: No the key is only used inside the wrapper to retrieve the cached version but not actually passed to the function. I need to document that clearly.

JeanKossaifi:
Thanks @yngvem, great points, as always! :)

JeanKossaifi:
So update, I don't actually see any speedup by caching the einsum equation, not sure whether we want to merge this feature anyway. Any thoughts?

cohenjer:
So if I understood the PR correctly, what you are doing is storing the contracting path in a global dictionary. While this could be interesting for further improvements, this PR alone will not lead to significant speed-up; what is costly is not to compute the einsum path, but to actually compute the contractions along that path. Therefore what we would need to cache is rather the intermediate results of the einsum.

But I do not believe we have access to these partial results easily, so maybe indeed this PR is not useful as such (but the code structure could be useful to start another PR where we cache contraction results along a path? Then we will also need to store the path so we would reuse this code).

No particular comment on the code itself, maybe the doc is not so clear so it took me some time to understand what was going on :p

JeanKossaifi:
@cohenjer I was talking about caching the contraction equation here, not the contraction path. The way I wrote the einsum tenalg backend is that I first check the validity of the operation (e.g. mttkrp) and then generate programmatically the corresponding contraction equation. The idea of this PR is that it is redundant to perform these checks and generate the equation at each call. However the gain seems to be pretty much non-existent so this may be over engineering. 

Caching the optimal contraction path on the other hand (#462) always helps.

PR #460, opened 2022-11-14
Adds opt-einsum plugin

JeanKossaifi:


cohenjer:
Similarly to #459 I will look at it but someone else should also check !

PR #462, opened 2022-11-22
Adds opt-einsum path caching plugin

JeanKossaifi:
Subsumes #460 but include the new plugin only and not the einsum caching


cohenjer:
I did not run the method on my computer but I don't spot any problem. 

Would be curious to see a few examples to check if the path obtained this way is different than the naive one, and if that leads to speedups for some toy contractions.

I also very much like this idea of pluggins, but I'm unsure how to use it myself. Did you have a specific syntax in mind? There is virtually no doc :/

PR #438, opened 2022-08-03
Use f-strings instead of str.format()

bakhtos:
Closes #427 
-All instances of using str.format() for printouts/error messages are changed to f-strings;
-Certain f-strings utilize the property that putting '=' sign after expression prints first the expression and then the result, to avoid duplicate typing of things like f"len(list) = {len(list)}", which is now simply f"{len(list) = }".
-requirements.txt now lists the  minimal version of Python (3.6) to support f-strings.

PR #464, opened 2022-12-09
Finished f-string formatting

aarmey:
Continues #438. I fixed any errors during testing, black formatted, and reviewed the code. Someone should probably quickly review before I merge.


JeanKossaifi:
Thanks @aarmey for finalizing this! It looks good to me. Looks like the tests are failing because of an unrelated issue with `fetch_indian_pines` #465, so happy to merge

PR #467, opened 2022-12-19
Use a secure link to tensorly.org

johnthagen:
The secure link works, so provide that to the user from the README to improve the security posture for users.

Note: Ideally the website link in the project description would also be updated to https://tensorly.org as it currently also points to the non-secure HTTP site.

Screen Shot 2022-12-19 at 8 38 47 AM


JeanKossaifi:
The issue is that the content is hosted on GitHub but we have the tensorly domain and so the link isn't actually secure. Does anyone have experience on how to automatically deploy the website on custom domain while supporting https?

aarmey:
If I am understanding your question, [Github supports HTTPS over custom domains now](https://docs.github.com/en/pages/getting-started-with-github-pages/securing-your-github-pages-site-with-https).

JeanKossaifi:
Great, thanks @johnthagen @aarmey, merging!

PR #474, opened 2023-01-05
Remove in-place projection operations in PARAFAC2

aarmey:
I noticed that the in-place operations here were completely overwriting the existing values, so there wasn't a benefit of keeping this as in-place except maybe reducing allocations. For some backends, such as Jax, in-place operations are extremely slow, and so overall I expect this is both faster and more readable.


JeanKossaifi:
Thanks @aarmey, looks good to me. I really love changes that make the code more readable, which this does!
Merging.

PR #478, opened 2023-01-16
Moderately speeds up and adds testing for CP, PARAFAC2, and Tucker

aarmey:
- In places where there is a lot of duplicated code (e.g., testing CP and NNCP), the two blocks are consolidated with parameterization.
- Some additional assertions to check for more properties of the results.
- Not all tests need to fit to convergence. In those cases I reduced the iteration count.
- In some places the dataset needed not be so large.
- Github actions needed to be updated.
- We now print out the 10 slowest unit tests to highlight slow cases.


JeanKossaifi:
Thanks @aarmey, it looks great to me! I am particularly fond of these PRs that really improve quality of life of the projects!

PR #480, opened 2023-01-20
Fetch kinetic

### LINKED ISSUE #475: Fetch Kinetic dataset
Temporarily skipping the test for the [Kinetic dataset](https://github.com/tensorly/tensorly/blob/8ff7e2950ca939c64641bfe0169e455f7cc59b56/tensorly/datasets/data_imports.py#L856) - the download link and data format seem to have changed. 

https://github.com/tensorly/tensorly/blob/8ff7e2950ca939c64641bfe0169e455f7cc59b56/tensorly/datasets/tests/test_imports.py#L42

cohenjer:
This Pull Request modifies the [Kinetic](https://ucphchemometrics.com/datasets/) (search Kinetic) dataset in tensorly in the following ways:

- add a local copy of the data (permission accorded by the authors, no license). Also add missing values positions.
   -->  more precisely, the added data is X60t, which is a reduced dataset from the original set of measurements. It is 12x smaller than the full measurement tensor (1/4 of entries on mode emission, 1/3 on mode excitation).
- add label/ticks information on all modes. This was slightly tricky because the original publication and the data itself do not contain precise values for the ticks on spectral modes. I extrapolated the values from the plots in the original publication. There is a possibility that these values are slightly off  but the results I have perfectly match the publication graphs.
- Updated the doc to stress that any usage of the data should be accompanied by a reference to the original publication.
- Updated name (fetch -> load) in loaders and tests.

Also this fixes #475 

Here is what I get using `nonnegative_parafac_hals` with random init and rank 2, it is essentially the same as in the original reference (some differences on the time mode, but I think the paper shows the results for X20t which is a variant of this dataset further compressed along time)
Kinetic_factors

edit: forgot the black linting and some last updates, adding another commit sorry !


tensorly/datasets/data_imports.py
```diff
@@ -853,16 +853,14 @@ def load_indian_pines():
     )
 
 
-def fetch_kinetic():
+def load_kinetic():
     """
-    Loads kinetic fluorescence dataset from website and returns it as tensorly tensor without storing the data
-    in the hard drive.The data is well suited for Parafac and multi-way partial least squares regression (N-PLS).
+    Loads the kinetic fluorescence dataset (X60t) as a tensorly tensor. The data is well suited for Parafac and multi-way partial least squares regression (N-PLS). Missing data are replaced by 0s, and a missing value mask is provided.
+    Data is a courtesy of Rasmus Bro and collaborators, it can be originally downloaded at https://ucphchemometrics.com/. Please cite the original reference if you use this data in any way.
```
JeanKossaifi, line 859: Can we add the reference in the doctoring as we do for e.g. CP so it appears in the API?

JeanKossaifi:
That's great, thanks @cohenjer! One minor comment to have the reference appear in the website but looks good to me otherwise!

cohenjer:
Done ! Also added the refs in the same style for other dataset, and updated the api text file for the website (otherwise the dataset were not showing).

JeanKossaifi:
Awesome, thanks @cohenjer, looks great!

Issue #473, opened 2023-01-05
Import error in Ubuntu 22.04 LTS


grassFlamingo:
Dear developers,

When I try import tensorly (0.8.0, 7a252eb, installed via `make install`) in Ubuntu 22.04 LTS with Python 3.7 by
```python
import tensorly
```
it will raise an error
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/xxx/tensorly/tensorly/__init__.py", line 10, in <module>
    from .cp_tensor import (
  File "/xxx/tensorly/tensorly/cp_tensor.py", line 7, in <module>
    from ._factorized_tensor import FactorizedTensor
  File "/xxx/tensorly/tensorly/_factorized_tensor.py", line 2, in <module>
    from . import tenalg
  File "/xxx/tensorly/tensorly/tenalg/__init__.py", line 91, in <module>
    TenalgBackendManager.initialize_backend()
  File "/xxx/tensorly/tensorly/backend/__init__.py", line 291, in initialize_backend
    cls.set_backend(backend_name)
  File "/xxx/tensorly/tensorly/backend/__init__.py", line 335, in set_backend
    backend = cls.load_backend(backend)
  File "/xxx/tensorly/tensorly/tenalg/__init__.py", line 81, in load_backend
    importlib.import_module(f"tensorly.tenalg.{backend_name}_tenalg")
  File "/xxx/anaconda3/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/xxx/tensorly/tensorly/tenalg/core_tenalg/__init__.py", line 4, in <module>
    from .generalised_inner_product import inner
  File "<fstring>", line 1
    (tensor1.shape=)
                  ^
SyntaxError: invalid syntax
```

While checking the codes, it seems that some f-strings are incorrect. They looks like
```python
f"{variable =}"
```
I think that the equal sign is not necessary. Therefore, I have made the following changes.
```
diff --git a/tensorly/tenalg/core_tenalg/generalised_inner_product.py b/tensorly/tenalg/core_tenalg/generalised_inner_product.py
index 7544303..0e1e338 100644
--- a/tensorly/tenalg/core_tenalg/generalised_inner_product.py
+++ b/tensorly/tenalg/core_tenalg/generalised_inner_product.py
@@ -31,7 +31,7 @@ def inner(tensor1, tensor2, n_modes=None):
[...]

PR #482, opened 2023-01-23
Fix and test of compatibility with python 3.7, 3.9 and 3.10

### LINKED ISSUE #479: Error reported when import tensorly (version: 0.8.0) in Python 3.7
In Pyhone 3.7, importing TensorLy, `import tensorly as tl`, will get the following error message:

```
from .generalised_inner_product import inner
  File "<fstring>", line 1
    (tensor1.shape=)
                  ^
SyntaxError: invalid syntax
```

The problem should be caused by the updates in the following lines

https://github.com/tensorly/tensorly/blob/8ff7e2950ca939c64641bfe0169e455f7cc59b56/tensorly/tenalg/core_tenalg/generalised_inner_product.py#L34

https://github.com/tensorly/tensorly/blob/8ff7e2950ca939c64641bfe0169e455f7cc59b56/tensorly/tenalg/core_tenalg/generalised_inner_product.py#L48

which only work in Python 3.8 and 3.9. Will TensorLy stop supporting Python 3.7 and 3.6?

aarmey:
Fixes #479 and #473.


tensorly/utils/tests/test_prod.py
```diff
@@ -1,10 +0,0 @@
-from .._prod import _prod
-from math import prod
-import tensorly as tl
```
aarmey, line 3: I figured, balancing the tradeoffs, that it was better to remove this test but allow testing on Python 3.7. This fallback rarely occurs, has previously been tested, and will be removed soon when 3.7 reaches end-of-life.

JeanKossaifi: Makes sense -- can we instead just comment it out (or better yet, skip for Python >= 3.8.x) and make a note to remove utils.prod altogether when we drop Python 3.7 support?

JeanKossaifi:
Aside from the small comment on that test, looks great to me, thanks @aarmey !

JeanKossaifi (approved):

aarmey:
Thanks @JeanKossaifi. Added back the test based on Python version and am merging!

PR #485, opened 2023-01-24
Fix einsum backend and run full test suite against it

aarmey:
In profiling some code, I tried the einsum backend and noticed an issue. Running the full test suite against it also uncovered a couple other issues. I added the einsum backend as a Github Actions run to prevent future issues, and corrected the issues I uncovered.


JeanKossaifi:
Thanks @aarmey, this is great. I went through the code and it looks good to me, merging!

Issue #487, opened 2023-02-15
cp_permute_factors() sometimes return the wrong permutation


maximeguillaud:
#### Describe the bug
cp_permute_factors() sometimes returns a wrong result due to a flawed definition of the cosine distance for tensors.
Currently, the cosines are added algebraically across the tensor modes, which is wrong. For instance, tensors can have fully colinear factors yielding cosines +1 and -1 in two modes. These cosines cancel out when taking the algebraic sum. The sum across the modes should probably be computed over the absolute values of the cosines.

The sum I am referring to is computed implicitly through the concatenation operation here and the subsequent inner product:
https://github.com/tensorly/tensorly/blob/e5dd2840207e019e1aece7e91928324af0a7857b/tensorly/cp_tensor.py#L680

#### Steps or Code to Reproduce

```
from tensorly.random import random_cp
from tensorly.cp_tensor import cp_permute_factors, cp_to_tensor
shape = (3, 4, 5)
rank = 4
cp_tensor_1 = random_cp(shape, rank, random_state = 0)
cp_tensor_2 = cp_tensor_1.cp_copy()
col_order_2 = [3, 1, 2, 0]
for f in range(3):
    cp_tensor_2.factors[f] = cp_tensor_1.factors[f][:, col_order_2]*(-1)**(f+1)
cp_tensors, permutation = cp_permute_factors(cp_tensor_1, cp_tensor_2)
```
Note that here, in addition to the permutation, factors for modes 0 and 2 are multiplied with -1 to yield cp_tensor_2.

#### Expected behavior

permutation should be equal to col_order_2, i.e.  [3, 1, 2, 0]

#### Actual result

permutation is equal to [2, 1, 3, 0]

#### Versions
Linux-6.1.0-1006-oem-x86_64-with-glibc2.35
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
NumPy 1.24.2
SciPy 1.10.0
TensorLy 0.8.0

cohenjer:
Hi @maximeguillaud, you are right. The best way to do that would be to take the product of the absolute values of the cosines. Feel free to make a small PR to change that if you can!

yngvem:
This is actually already implemented in [TensorLy-Viz](http://tensorly.org/viz/stable/). We should probably migrate some of that functionality over here and add more links to it in the documentation.

@MarieRoald and I could do it, but that would have to wait until we have some spare time again.

[...]

PR #489, opened 2023-02-20
Hotfix permute factors

cohenjer:
This is a fix for the `cp_permute_factors` function. As pointed out by @maximeguillaud the usage of `congruence_coefficient` for CP was to input a concatenated list `[A,B,C]` of factor matrices and compute the congruence on that whole list. But this is not correct because of the sign ambiguity in CP decomposition.

A better way to compute the congruence coefficient is to compute the congruence for A, B, and C and then use the product of the absolute value of the outcomes to perform the linear assignment. 

I modified the `congruence_coefficient` function accordingly by allowing two list of matrices as inputs rather than only two matrices. The old code should run without issues since the previous inputs (matrix, matrix) is also supported. All tests are passing locally, and the bug raised in #487 is fixed:

```python 
import tensorly as tl
from tensorly.random import random_cp
from tensorly.cp_tensor import cp_permute_factors, cp_to_tensor
shape = (3, 4, 5)
rank = 4
cp_tensor_1 = random_cp(shape, rank, random_state = 0)
cp_tensor_1[1][0] *= 2
cp_tensor_2 = cp_tensor_1.cp_copy()
col_order_2 = [3, 1, 2, 0]
for f in range(3):
    cp_tensor_2.factors[f] = cp_tensor_1.factors[f][:, col_order_2]*(-1)**(f+1)
cp_tensors, permutation = cp_permute_factors(cp_tensor_1, cp_tensor_2)
```
produces the correct permutation [3, 1, 2, 0].

I also updated the doc of congruence coefficient to reflect that the returned permutation should be applied on the second input matrix/list of matrices.

(and sorry for the noisy commits, I can squash them if needed)

### DIFF
--- a/tensorly/metrics/factors.py
+++ b/tensorly/metrics/factors.py
@@ -14,12 +14,17 @@ def congruence_coefficient(matrix1, matrix2, absolute_value=True):
         \\frac{\\mathbf{v}_1^T \\mathbf{v}_1^T}{\\|\\mathbf{v}_1^T\\| \\|\\mathbf{v}_1^T\\|}
 
     When we compute the congruence between two matrices, we find the optimal permutation of
-    the columns and return the mean congruence and the permutation.
+    the columns and return the mean congruence and the permutation. The output permutation is the one
+    that permutes the columns of matrix2 onto the closest columns in matrix1.
+
+    If a list of matrices is provided for each input, we define the congruence coefficient as the
+    product of the absolute values of pairs of matrices. The lists must therefore have the same size.
+    The output permutation also applies to each matrix of the lists.
 
     Parameters
     ----------
-    matrix1 : tensorly.Tensor
-    matrix2 : tensorly.Tensor
+    matrix1 : tensorly.Tensor or list of tensorly.Tensor
+    matrix2 : tensorly.Tensor of list of tensorly.Tensor to permute.
     absolute_value : bool
         Whether to take the absolute value of all vector products before finding the optimal permutation.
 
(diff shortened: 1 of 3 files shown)


cohenjer:
And we have the usual Tensorflow bug (tensors are immutable in Tensorflow). Sure I can change the test but we can also skip Tensorflow in the test since alternatives are typically cumbersome.

tensorly/metrics/factors.py
```diff
@@ -28,16 +33,33 @@ def congruence_coefficient(matrix1, matrix2, absolute_value=True):
...
-        raise ValueError("Matrices must have same shape")
-
-    matrix1 = matrix1 / T.norm(matrix1, axis=0)
-    matrix2 = matrix2 / T.norm(matrix2, axis=0)
-    all_congruences = T.dot(T.transpose(matrix1), matrix2)
-    if absolute_value:
-        all_congruences = T.abs(all_congruences)
-    all_congruences = T.to_numpy(all_congruences)
+    # Check if matrix1 and matrix2 are lists of the same length
+    if isinstance(matrix1, list):
```
JeanKossaifi, line 37: I think it is safer to do it the other way around: if tl.is_tensor(matrix1), make it into a list, otherwise, check that both have the same len (alternatively we could also check if it is iterable to not limit to list but accept list-like objects or tuples, but that would be slightly more involved/convoluted)

tensorly/metrics/factors.py
```diff
@@ -28,16 +33,33 @@ def congruence_coefficient(matrix1, matrix2, absolute_value=True):
...
+        if not isinstance(matrix2, list) or len(matrix1) != len(matrix2):
+            raise ValueError("Input lists of matrices must have the same length")
+    else:
+        matrix1 = [matrix1]
+        matrix2 = [matrix2]
+    all_congruences_list = []
+    # Store an arbitrary column dimension, they should all be the same
+    num_col_mat1 = T.shape(matrix1[0])[1]
```
JeanKossaifi, line 45: We can also check all matrices have the same number of columns  
```python
columns = [T.shape(m) for m in matrix1] + [T.shape(m) for m in matrix2]
if len(np.unique(np.array(columns)) > 1:
    raise ValueError(...)
```

tensorly/metrics/factors.py
```diff
@@ -28,16 +33,33 @@ def congruence_coefficient(matrix1, matrix2, absolute_value=True):
...
+    for mat1, mat2 in zip(matrix1, matrix2):
+        num_cols = T.shape(mat1)[1]
+        if T.shape(mat1) != T.shape(mat2):
+            raise ValueError("Matrices must have same shape")
+        if T.shape(mat1)[1] != num_col_mat1 or T.shape(mat2)[1] != num_col_mat1:
+            # check that all matrices have the same number of columns
+            raise ValueError("Matrices must have the same number of columns")
+
+        mat1 = mat1 / T.norm(mat1, axis=0)
```
JeanKossaifi, line 54: Should we add an `eps` (we can use tl.eps) or check for zero norm?

JeanKossaifi (commented):
Thanks @cohenjer, looks great to me, left a few comments.
I think the Black error might be from the previously merged PR. 

I agree, let's skip TensorFlow. We might want to just deprecate that backend at some point..

tensorly/metrics/factors.py
cohenjer: Great snippet thanks ! I did check that by comparing all sizes to the first one, but your code is cleaner.

tensorly/metrics/factors.py
cohenjer: I guess we should not allow 0 norm in the columns of inputs, so I will add a check for that.

cohenjer:
> Thanks @cohenjer, looks great to me, left a few comments. I think the Black error might be from the previously merged PR.
> 
> I agree, let's skip TensorFlow. We might want to just deprecate that backend at some point..

I skipped Tensorflow in the test using the same code snippet as in test_cp_tensor.py. The black lining error is indeed from a previous PR.

tensorly/metrics/factors.py
```diff
@@ -33,24 +34,24 @@ def congruence_coefficient(matrix1, matrix2, absolute_value=True):
...
-        if T.shape(mat1)[1] != num_col_mat1 or T.shape(mat2)[1] != num_col_mat1:
-            # check that all matrices have the same number of columns
-            raise ValueError("Matrices must have the same number of columns")
-
+        if T.shape(mat1)[0] != T.shape(mat2)[0]:
+            raise ValueError("Pairs of matrices must have the same number of rows")
+        # Check if any norm is exactly zero to avoid singularity
+        if T.prod(T.norm(mat1, axis=0)) * T.prod(T.norm(mat2, axis=0)) == 0:
+            raise ValueError("Columns of all matrices should have nonzero l2 norm")
         mat1 = mat1 / T.norm(mat1, axis=0)
```
JeanKossaifi, line 55: I guess we could even just wrap those two divisions in a try/except and just catch `ZeroDivisionError` -- better ask for forgiveness than permission.

tensorly/metrics/factors.py
JeanKossaifi: I guess we could even just wrap those two divisions in a try/except and just catch `ZeroDivisionError` -- better ask for forgiveness than permission.

cohenjer: Turns out e.g. with numpy backend we don't have a ZeroDivisionError but a RuntimeWarning; and mat1 and mat2 are filled with NaNs. I am not confident in my knowledge of warnings with the different backends, so I am not sure how to catch that properly :/

JeanKossaifi: I guess we can just explicitly check if T.norm(...) > 0

JeanKossaifi:
Looks good to me -- could you run black to fix that last CI fail before we merge?

cohenjer:
> Looks good to me -- could you run black to fix that last CI fail before we merge?

I think the remaining black error comes from another PR... the files from this PR are fine in the test.

JeanKossaifi:
Thanks, merging!

PR #491, opened 2023-02-27
Add logsumexp function

braun-steven:
This adds the `logsumexp` function using the logsumexp trick which is the numerical stable version of applying `log(sum(exp(x)))`. I chose to default to the backend implementation if available (pytorch, numpy/scipy, tensorflow, jax) and implemented the numerical stable computation of `log(sum(exp(x)))` when the backend did not provide a logsumexp function (mxnet, cupy).

Additionally, I've added a unit test that checks against the numpy/scipy baseline with some 3d data tensor over all axis.

### DIFF
--- a/tensorly/backend/__init__.py
+++ b/tensorly/backend/__init__.py
@@ -114,6 +114,7 @@ class BackendManager(types.ModuleType):
         "acosh",
         "atanh",
         "partial_svd",
+        "logsumexp",
     ]
     _attributes = [
         "int64",
--- a/tensorly/backend/core.py
+++ b/tensorly/backend/core.py
@@ -1162,6 +1162,11 @@ class Backend(object):
         raise
 
     @staticmethod
+    def logsumexp(x):
+        """Calculate the log of the sum of exponentials of input elements in a numerically stable way."""
+        raise NotImplementedError
+
+    @staticmethod
     def exp(x):
         """Calculate the exponential of all elements in the input array."""
         raise NotImplementedError
--- a/tensorly/backend/core.py
+++ b/tensorly/backend/core.py
@@ -1162,8 +1162,22 @@ class Backend(object):
         raise
 
     @staticmethod
-    def logsumexp(x):
-        """Calculate the log of the sum of exponentials of input elements in a numerically stable way."""
+    def logsumexp(x, axis=None):
+        """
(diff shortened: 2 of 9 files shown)


JeanKossaifi:
Thanks @braun-steven! Is there anywhere it could be used in the current codebase or is this just for new functionalities?

braun-steven:
I haven't found any use cases in the current code base. We need the functionality for our library [SPFLow](https://github.com/SPFlow/SPFlow) which we are currently rewriting to be used with tensorly as backend. We could have also written it as part of our codebase but thought that pushing the added functionality upstream would be good for tensorly as well.

JeanKossaifi:
Indeed, thanks for contributing @braun-steven ! Looks good to me, merging.

JeanKossaifi:
As a side note @braun-steven let us know if you need more features on the tensorly side -- we can also showcase SPFlow somewhere on the website, e.g. in an ecosystem page or something!

braun-steven:
> As a side note @braun-steven let us know if you need more features on the tensorly side

Will do! But I think we will probably be able to implement most functionality that we need and is missing ourselves and push it upstream. In case of issues/questions/problems we will definitely get back at you :)


The current status is that the rewrite of SPFlow is quite far and we wanted to initially support a numpy and a torch backend which was almost done, with support for TensorFlow and Jax in future (>1.0.0) versions. Then we found out about tensorly and investigated if it is worth the effort to replace the numpy backend with tensorly to get all the backends "for free" (with the exception of some specific modules such as probability distributions etc). So the current status is that we are replacing numpy with tensorly and see how far we come and into how many issues we run on our side. 

> we can also showcase SPFlow somewhere on the website, e.g. in an ecosystem page or something!

That sounds awesome! Once the backend rewrite is done and we are ready to release our 1.0.0 version built on top of tensorly, I would gladly come back at you :)

felixdivo:
I think this wasn't added to the docs, or did I just miss it?

https://tensorly.org/stable/search.html?q=logsumexp

JeanKossaifi:
Indeed, it hasn't been added, good catch @felixdivo! Do you want to make a small PR and add it to the doc (in the [API doc](https://github.com/tensorly/tensorly/blob/main/doc/modules/api.rst))? 
I can add it later otherwise.

PR #492, opened 2023-03-04
Fix CP Partial Least Square

cyrillustan:
This fixes some implementation issues in `CP_PLSR()`. I also added the calculation of $R^2$ of X and Y, very useful parameters to inspect the performance of the model.

### DIFF
--- a/tensorly/metrics/tests/test_regression.py
+++ b/tensorly/metrics/tests/test_regression.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 import tensorly as tl
-from ..regression import MSE, RMSE, correlation
+from ..regression import MSE, RMSE, R2_score, correlation
 from ...testing import assert_array_almost_equal
 
 
@@ -19,6 +19,14 @@ def test_RMSE():
     y_pred = tl.tensor([0, -1, 1, -1])
     true_mse = 1
     assert_array_almost_equal(RMSE(y_true, y_pred), true_mse)
+
+
+def test_R2_score():
+    """Test for RMSE"""
+    X_original = np.random.rand(5, 4, 3)
+    assert R2_score(X_original, X_original) == 1.0
+    assert R2_score(X_original, X_original * 2) == 0.0
+    assert R2_score(X_original, np.zeros_like(X_original)) == 0.0
 
 
 def test_correlation():
--- a/tensorly/metrics/tests/test_regression.py
+++ b/tensorly/metrics/tests/test_regression.py
@@ -26,7 +26,7 @@ def test_R2_score():
     X_original = tl.randn((5, 4, 3))
     assert R2_score(X_original, X_original) == 1.0
     assert R2_score(X_original, X_original * 2) == 0.0
-    assert R2_score(X_original, np.zeros_like(X_original)) == 0.0
+    assert R2_score(X_original, tl.zeros_like(X_original)) == 0.0
 
 
 def test_correlation():
--- a/tensorly/metrics/tests/test_regression.py
+++ b/tensorly/metrics/tests/test_regression.py
@@ -23,7 +23,7 @@ def test_RMSE():
 
 def test_R2_score():
     """Test for RMSE"""
-    X_original = np.random.rand(5, 4, 3)
+    X_original = tl.randn(5, 4, 3)
     assert R2_score(X_original, X_original) == 1.0
     assert R2_score(X_original, X_original * 2) == 0.0
     assert R2_score(X_original, np.zeros_like(X_original)) == 0.0
--- a/tensorly/metrics/tests/test_regression.py
+++ b/tensorly/metrics/tests/test_regression.py
@@ -23,7 +23,7 @@ def test_RMSE():
 
 def test_R2_score():
     """Test for RMSE"""
-    X_original = tl.randn(5, 4, 3)
+    X_original = tl.randn((5, 4, 3))
     assert R2_score(X_original, X_original) == 1.0
     assert R2_score(X_original, X_original * 2) == 0.0
     assert R2_score(X_original, np.zeros_like(X_original)) == 0.0
--- a/tensorly/regression/cp_plsr.py
+++ b/tensorly/regression/cp_plsr.py
@@ -86,6 +86,8 @@ class CP_PLSR:
         # Mean center the data, record info the object
         self.X_shape_ = T.shape(X)
         self.Y_shape_ = T.shape(Y)
+        self.original_X_ = T.copy(X)
+        self.original_Y_ = T.copy(Y)
         self.X_mean_ = T.mean(X, axis=0)
         self.Y_mean_ = T.mean(Y, axis=0)
         X -= self.X_mean_
@@ -97,6 +99,7 @@ class CP_PLSR:
         self.Y_factors = [
             T.zeros((l, self.n_components), **T.context(X)) for l in T.shape(Y)
         ]
+        self.coef_ = T.zeros((self.n_components, self.n_components), **T.context(X))
 
         ## FITTING EACH COMPONENT
         for component in range(self.n_components):
@@ -151,6 +154,13 @@ class CP_PLSR:
                 self.Y_factors[1], T.index[:, component], comp_Y_factors_1
             )
 
+            B = T.lstsq(self.X_factors[0], T.reshape(comp_Y_factors_0, (-1, 1)))[0]
+            self.coef_ = T.index_update(
+                self.coef_,
+                T.index[:, component],
+                T.reshape(B, (-1,)),
+            )
+
             # Deflation
             X -= CPTensor(
                 (None, [T.reshape(ff, (-1, 1)) for ff in comp_X_factors])
@@ -158,10 +168,10 @@ class CP_PLSR:
             Y -= T.dot(
                 T.dot(
                     self.X_factors[0],
-                    T.lstsq(self.X_factors[0], T.reshape(comp_Y_factors_0, (-1, 1)))[0],
+                    T.reshape(B, (-1, 1)),
                 ),
                 T.reshape(comp_Y_factors_1, (1, -1)),
-            )  # Y -= T pinv(T) u q'
+            )  # Y -= T b q' = T pinv(T) u q'
 
         return self
 
@@ -178,16 +188,31 @@ class CP_PLSR:
                 f"Training X has shape {self.X_shape_}, while the new X has shape {T.shape(X)}"
             )
         X -= self.X_mean_
-        factors_kr = khatri_rao(self.X_factors, skip_matrix=0)
-        unfolded = unfold(X, 0)
-        scores = T.lstsq(factors_kr, T.transpose(unfolded))[0]  # = Tnew
-        estimators = T.lstsq(self.X_factors[0], self.Y_factors[0])[0]
-        return (
-            T.dot(
-                T.dot(T.transpose(scores), estimators), T.transpose(self.Y_factors[1])
+        X_projection = T.zeros((T.shape(X)[0], self.n_components), **T.context(X))
+        for component in range(self.n_components):
+            X_projection = T.index_update(
+                X_projection,
+                T.index[:, component],
+                multi_mode_dot(
+                    X,
+                    [factor[:, component] for factor in self.X_factors[1:]],
+                    range(1, T.ndim(X))
+                )
             )
-            + self.Y_mean_
-        )
+            X -= CPTensor(
+                (
+                    None,
+                    [T.reshape(X_projection[:, component], (-1, 1))] +
+                    [T.reshape(factor[:, component], (-1, 1)) for factor in self.X_factors[1:]]
+                )
+            ).to_tensor()
+        return T.dot(
+            T.dot(
+                X_projection,
+                self.coef_
+            ),
+            T.transpose(self.Y_factors[1])
+        ) + self.Y_mean_
 
     def transform(self, X, Y=None):
         """Apply the dimension reduction from fitting to a new tensor.
@@ -256,11 +281,11 @@ class CP_PLSR:
 
                 Y -= T.dot(
                     T.dot(
-                        T.lstsq(T.transpose(X_scores), T.transpose(X_scores))[0],
-                        Y_scores[:, [component]],
+                        X_scores,
+                        T.reshape(self.coef_[:, component], (-1, 1)),
                     ),
-                    T.transpose(self.Y_factors[1][:, [component]]),
-                )  # Y -= T pinv(T) u q'
+                    T.transpose(T.reshape(self.Y_factors[1][:, component], (-1, 1)))
+                )
             return X_scores, Y_scores
 
         return X_scores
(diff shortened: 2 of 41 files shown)


tensorly/regression/cp_plsr.py
```diff
@@ -97,6 +99,7 @@ def fit(self, X, Y):
         self.Y_factors = [
             T.zeros((l, self.n_components), **T.context(X)) for l in T.shape(Y)
         ]
+        self.coef_ = T.zeros((self.n_components, self.n_components), **T.context(X))
```
aarmey, line 102: What is `coef_`? Can you provide a comment?

tensorly/regression/cp_plsr.py
```diff
@@ -178,14 +188,29 @@ def predict(self, X):
...
-        factors_kr = khatri_rao(self.X_factors, skip_matrix=0)
-        unfolded = unfold(X, 0)
-        scores = T.lstsq(factors_kr, T.transpose(unfolded))[0]  # = Tnew
-        estimators = T.lstsq(self.X_factors[0], self.Y_factors[0])[0]
-        return (
-            T.dot(
-                T.dot(T.transpose(scores), estimators), T.transpose(self.Y_factors[1])
+        X_projection = T.zeros((T.shape(X)[0], self.n_components), **T.context(X))
+        for component in range(self.n_components):
+            X_projection = T.index_update(
```
aarmey, line 193: Does this mean that the Bro paper algorithm did in fact work in the end? I'm not seeing you subtracting from X_orig like we had previously discussed...

tensorly/regression/tests/test_cp_plsr.py
```diff
@@ -207,3 +213,49 @@ def test_optimized_covariance(n_latent):
         )
 
     assert_allclose(max_cov, pls_cov)
+
+
+@pytest.mark.parametrize("n_response", [5, 7, 9])
```
aarmey, line 218: I don't see you using `n_response`. Can this parameterization be removed?

tensorly/regression/cp_plsr.py
```diff
@@ -34,7 +34,7 @@ class CP_PLSR:
     """
 
     def __init__(
-        self, n_components, tol=1.0e-9, n_iter_max=100, random_state=None, verbose=False
+        self, n_components, tol=1.0e-8, n_iter_max=100, random_state=None, verbose=False
```
aarmey, line 37: The interaction within a component was going on up to `n_iter_max` with the old tolerance and the tensorflow backend. I'm not sure why the accuracy is different in a backend-specific way, but this was helping with testing taking a more reasonable amount of time.

aarmey (changes requested):
@cyrillustan thanks for fixing this. Here are a few questions to start review.

tensorly/regression/cp_plsr.py
cyrillustan: These are the coefficients of the linear model, the same as $\beta$ in ordinary least square $Y = X \beta + \epsilon$. This was notated in Bro's paper as $\mathbf{b}$. This parameter has always existed in Bro's paper, but was previously dismissed by me as trivial since there was no explanation for it.

tensorly/regression/cp_plsr.py
cyrillustan: Yes. The subtracting from the original Y turned out to be not a problem at all given the 0 columns elsewhere. Every component explains exactly what we have subtracted each round, in this case, the project of X to Y. Therefore, the definition of $R^2$ is rightfully different for X and Y; they are not symmetrical (surprise, surprise).

tensorly/regression/tests/test_cp_plsr.py
cyrillustan: I have added it in the function so it will make sense now.

aarmey:
Thanks @cyrillustan. I also made an adjustment to the CP factorization so that it is faster. This is a pretty big change, so I will wait for one other person to review before merging.

tensorly/regression/cp_plsr.py
```diff
@@ -86,6 +98,8 @@ def fit(self, X, Y):
         # Mean center the data, record info the object
         self.X_shape_ = T.shape(X)
         self.Y_shape_ = T.shape(Y)
+        self.original_X_ = T.copy(X)
```
JeanKossaifi, line 101: I'm not sure I'd want to store the full X and Y, it seems unnecessary and potentially very expensive (memory-wise): i'd let the user take care of that

tensorly/regression/cp_plsr.py
```diff
@@ -151,17 +174,24 @@ def fit(self, X, Y):
                 self.Y_factors[1], T.index[:, component], comp_Y_factors_1
             )
 
+            B = T.lstsq(self.X_factors[0], T.reshape(comp_Y_factors_0, (-1, 1)))[0]
+            self.coef_ = T.index_update(
```
JeanKossaifi, line 178: As I'm reviewing this, given your other PR @aarmey I guess if we deprecate tensorflow we'll be able to remove those pesky index_updates (I think JAX supports the syntactic sugar now)

tensorly/regression/cp_plsr.py
```diff
@@ -178,14 +208,29 @@ def predict(self, X):
...
+            X_projection = T.index_update(
+                X_projection,
+                T.index[:, component],
+                multi_mode_dot(
+                    X,
+                    [factor[:, component] for factor in self.X_factors[1:]],
+                    range(1, T.ndim(X)),
+                ),
             )
+            X -= CPTensor(
```
JeanKossaifi, line 222: Creating a CP tensor for this will add unnecessary overhead: we can directly use [`tl.tenalg.outer`](https://tensorly.org/dev/modules/generated/tensorly.tenalg.outer.html#tensorly.tenalg.outer) instead.

tensorly/regression/cp_plsr.py
```diff
@@ -256,11 +301,11 @@ def transform(self, X, Y=None):
 
                 Y -= T.dot(
                     T.dot(
-                        T.lstsq(T.transpose(X_scores), T.transpose(X_scores))[0],
-                        Y_scores[:, [component]],
+                        X_scores,
+                        T.reshape(self.coef_[:, component], (-1, 1)),
```
JeanKossaifi, line 305: Do we need the reshape here?

tensorly/regression/cp_plsr.py
```diff
@@ -256,11 +301,11 @@ def transform(self, X, Y=None):
...
                     ),
-                    T.transpose(self.Y_factors[1][:, [component]]),
-                )  # Y -= T pinv(T) u q'
+                    T.transpose(T.reshape(self.Y_factors[1][:, component], (-1, 1))),
```
JeanKossaifi, line 307: Same, do we need the reshape? It's just a vector.

tensorly/regression/cp_plsr.py
```diff
@@ -283,3 +328,33 @@ def fit_transform(self, X, Y):
             Return `x_scores` if `Y` is not given, `(x_scores, y_scores)` otherwise.
         """
         return self.fit(X, Y).transform(X, Y)
+
+    def X_r2_score(self):
```
JeanKossaifi, line 332: I would explicitly pass X here, and not store it, see first comment.

tensorly/regression/cp_plsr.py
```diff
@@ -283,3 +328,33 @@ def fit_transform(self, X, Y):
...
+            the amount of variance in X explained by the model
+
+        """
+        X_original = self.original_X_ - self.X_mean_
+        X_reconstructed = CPTensor([None, self.X_factors]).to_tensor()
+        return (
+            1 - T.norm(X_reconstructed - X_original) ** 2.0 / T.norm(X_original) ** 2.0
+        )
+
+    def Y_r2_score(self):
```
JeanKossaifi, line 347: Similarly, I think we can pass Y explicitly.

JeanKossaifi (commented):
Thanks, this looks good. 

I left some comments. I think we can move the R2 to the metrics subpackage (it's the same for both Y and X and can be useful for other classes too), and just compute it internally and store the results rather than the input X and Y. 

For the API, either:
1. we pass both X, X_mean and X_pred
2. we directly pass in an (optionally) centered X
3. we make X_mean optional and with a default of None

What do you think? I'd suggest following scikit-learn for this.

tensorly/regression/cp_plsr.py
cyrillustan: We can just calculate these numbers and save them in the object, but not save the `original_X` (only use them on the run). This is similar to how sklearn does.

tensorly/regression/cp_plsr.py
cyrillustan: `T.dot` will somehow report error on shape if it is not in a matrix.

tensorly/regression/cp_plsr.py
cyrillustan: Same, but `T.transpose(T.reshape)` is redundant indeed.

tensorly/regression/cp_plsr.py
cyrillustan: Okay, I'll change it to `cp_to_tensor()`

cyrillustan:
I separate the R2 function out of this class to `metric`. Hopefully it is in the appropriate file. I just removed the functions of R2 X and Y entirely and make them attributes, as it makes more sense to calculate them along the fitting process, and we don't have to save or pass the original tensor. This resembles how sklearn does variance explained, too.

tensorly/regression/cp_plsr.py
```diff
@@ -86,6 +103,8 @@ def fit(self, X, Y):
         # Mean center the data, record info the object
         self.X_shape_ = T.shape(X)
         self.Y_shape_ = T.shape(Y)
+        original_X_ = T.copy(X)
+        original_Y_ = T.copy(Y)
```
JeanKossaifi, line 107: Can we get around the creation of those copy? Can we just work always on the centered versions of X and Y (we can uncenter if needed as we have the mean)?

tensorly/regression/cp_plsr.py
JeanKossaifi: This is quite useful -- could we add it to the docstrings (e.g. in Notes) or the doc?

tensorly/regression/cp_plsr.py
```diff
@@ -151,17 +181,40 @@ def fit(self, X, Y):
...
+                self.coef_,
+                T.index[:, component],
+                T.reshape(B, (-1,)),
+            )
+
             # Deflation
-            X -= CPTensor(
-                (None, [T.reshape(ff, (-1, 1)) for ff in comp_X_factors])
-            ).to_tensor()
+            X -= cp_to_tensor((None, [T.reshape(ff, (-1, 1)) for ff in comp_X_factors]))
```
JeanKossaifi, line 192: Why not use tl.tenalg.outer? It's just a rank-1 CP, i.e. a single outer product of vectors. That may be faster as avoiding the checks and creation of None weights to handle.

tensorly/regression/cp_plsr.py
```diff
@@ -177,15 +230,30 @@ def predict(self, X):
...
+                    range(1, T.ndim(X)),
+                ),
             )
+            X -= cp_to_tensor((
+                    None,
+                    [T.reshape(X_projection[:, component], (-1, 1))]
+                    + [
+                        T.reshape(factor[:, component], (-1, 1))
+                        for factor in self.X_factors[1:]
+                    ],
```
JeanKossaifi, line 252: Same here, we could just do the tl.tenalg.outer if it proves faster? Might be more readable too?

JeanKossaifi (commented):
Thanks @cyrillustan! The R2 is in the right place indeed and you even added a unit-test, perfect!

Looks good to me, just a minor thing about using outer-product rather than cp_to_tensor, which might be a little faster/more readable as we're only substracting a rank-1 tensor, i.e. a simple outer-product of vectors.

tensorly/metrics/tests/test_regression.py
```diff
@@ -21,6 +21,14 @@ def test_RMSE():
     assert_array_almost_equal(RMSE(y_true, y_pred), true_mse)
 
 
+def test_R2_score():
+    """Test for RMSE"""
+    X_original = np.random.rand(5, 4, 3)
```
JeanKossaifi, line 26:
```suggestion
    X_original = tl.rand(5, 4, 3)
```

tensorly/regression/cp_plsr.py
cyrillustan: Hmm, I'm not sure how to avoid making a copy, either the original or the mean-centered version (what I changed into) of X and Y, if we still want to keep the $R^2$ calculation though. @aarmey do you have any idea?

aarmey: I think Jean's point is that right now you have two persistent copies—of both the original and mean-centered. Just keep a copy of one, and then build the other if/where you need it.

cyrillustan: Oh, then the latest change should make everything only mean-centered.

tensorly/metrics/tests/test_regression.py
```diff
@@ -21,6 +21,14 @@ def test_RMSE():
     assert_array_almost_equal(RMSE(y_true, y_pred), true_mse)
 
 
+def test_R2_score():
+    """Test for RMSE"""
+    X_original = tl.randn((5, 4, 3))
+    assert R2_score(X_original, X_original) == 1.0
+    assert R2_score(X_original, X_original * 2) == 0.0
+    assert R2_score(X_original, np.zeros_like(X_original)) == 0.0
```
JeanKossaifi, line 29:
```suggestion
    assert R2_score(X_original, tl.zeros_like(X_original)) == 0.0
```

JeanKossaifi:
Thinking about this more, how about calling the R2 `r2_score` for consistency with sklearn, and, as they do, just extracting its computation from the regression class itself @aarmey @cyrillustan?

We can then either 
* let the users compute the r2 score themselves `r2_score(ground-truth, prediction)` 
* and/or add a `score` method that takes as input both the prediction and ground-truth and essentially just acts as an alias to r2 (probably can handle centering etc as well)

cyrillustan:
I think my edits should fix the problem in `score()`. Let me know if there is anything else I should do. I see that some tests failed, but they don't seem to be related to `CP_PLSR` edits.

tensorly/regression/cp_plsr.py
```diff
@@ -283,3 +317,18 @@ def fit_transform(self, X, Y):
...
+
+        Parameters
+        ----------
+        X : ndarray
+            tensor data of shape (n_samples, N1, ..., NS), same dimension as the X
+            in self.fit() all except the first dimension
+        Y : 2D-array of shape (n_samples, n_predictions)
+            the ground truth labels associated with each sample
+        """
+        from ..metrics.regression import R2_score
```
aarmey, line 332: Is there a reason this is imported just at the function level?

aarmey (changes requested):
One more question, then I am comfortable merging this.

XiaoXiao-Woo:
作业已收到，祝同学生活快乐

tensorly/regression/cp_plsr.py
cyrillustan: Oh this is just mimicking how [sklearn did it](https://github.com/scikit-learn/scikit-learn/blob/9aaed498795f68e5956ea762fef9c440ca9eb239/sklearn/base.py#L720).

PR #494, opened 2023-03-05
Add mxnet warning and remove old deprecations

aarmey:
I wanted to post this for feedback. First, this removes some old deprecation warnings. Second, I propose to deprecate both the mxnet and tensorflow backends and remove them in the next major version. We have talked about doing this for tensorflow for a while. With mxnet, we have repeatedly run into problems with the backend and it appears that development of the project has stopped. There have been essentially no commits to the project over the last 6 months. I doubt users are using these backends and, if they are, the warning should allow us to get feedback before removing them.

### DIFF
--- a/doc/user_guide/tensor_decomposition.rst
+++ b/doc/user_guide/tensor_decomposition.rst
@@ -137,7 +137,7 @@ The advantage of the TT/tensor-train decomposition is that both of its number of
 Implementations
 +++++++++++++++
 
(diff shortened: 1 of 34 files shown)


JeanKossaifi:
Hmm yes, that's a tough one, I've been pondering it for a while. I'm surprised by the low number of commits on MXNet, I expected activity to increase now that they are fully in the Apache community and with v.2.0 upcoming for so long.

A warning that at least it may be deprecated in the future might be the way to go. Also a good opportunity to then clean up a little the repo and make the code more readable.

aarmey:
Ok great. I'll reword the warning for mxnet and take out TensorFlow from here.

JeanKossaifi:
Thanks @aarmey, much needed cleanup! Merging.

PR #495, opened 2023-03-20
Move choices about Khatri-Rao to tenalg backend

aarmey:
I know we discussed this previously, but now with the tenalg backend it feels like the choice of reshape- vs. einsum-based KR should occur in one place. This makes that change by removing `kr()` from the main backend and consolidating some code.


tensorly/backend/jax_backend.py
```diff
@@ -58,29 +58,6 @@ def lstsq(a, b):
...
-                m if i else m * self.reshape(weights, (1, -1))
-                for i, m in enumerate(matrices)
-            ]
-
-        m = mask.reshape((-1, 1)) if mask is not None else 1
-        return np.einsum(operation, *matrices).reshape((-1, n_columns)) * m
-
-    @staticmethod
-    def logsumexp(tensor, axis=0):
-        return jax.scipy.special.logsumexp(tensor, axis=axis)
```
JeanKossaifi, line 82: Why did you remove the logsumexp? I believe it was needed by another package downstream.

JeanKossaifi (commented):
Thanks @aarmey! The reason to leave it there was in case backend would provide efficient Khatri-Rao implementations, I think having it in tenalg is good.

I think in Jax the PR removed another function.

I am slightly conflicted by the _validate_khatri_rao. 
I agree it is a little cleaner to have it shared but equally, it's pretty clear and I'm not a fan of the side effect -- e.g. it doesn't just validate the inputs it also changes them. We could mix in the tests for the sizes within the computation itself (at least for the non-einsum-based version) to make it more efficient. 

What do you think?

tensorly/backend/jax_backend.py
aarmey: Oops! Missed that resolving a merge conflict.

aarmey:
Thanks for the thoughts. I'm not sure there would be a significant difference in efficiency from merging the checks into the computation itself because (if I follow), it would just eliminate obtaining the shape a few times. This has got to be much faster than any sort of actual data operation. `einsum()` is naturally going to perform these size checks, so we could eliminate them altogether in several cases, but I think these checks provide a more understandable error message.

One option would be to rename the function to `prepare_khatri_rao()`, which I think better reflects that it modifies the data.

Another possibility would be to just pull out the `skip_matrix` functionality into the parent functions. `reverse` could be pulled out similarly, although it is never used outside of testing so perhaps we drop it?

JeanKossaifi:
> I'm not sure there would be a significant difference in efficiency from merging the checks into the computation itself because (if I follow), it would just eliminate obtaining the shape a few times. This has got to be much faster than any sort of actual data operation. einsum() is naturally going to perform these size checks, so we could eliminate them altogether in several cases, but I think these checks provide a more understandable error message.

Yes, I agree with both points -- speed is not really a concern here. The checks are more to offer helpful error messages to the users and their overhead minimal. I meant more that I'd be fine duplicating the code rather than having a separate function just for checks as it's quite specific to Khatri-Rao. 

> One option would be to rename the function to prepare_khatri_rao(), which I think better reflects that it modifies the data.

Makes sense, though this feels like this may be over-engineering almost and make the code actually harder to read to have the logic separate from the code -- it's just checks, not complicated code :)

> Another possibility would be to just pull out the skip_matrix functionality into the parent functions. reverse could be pulled out similarly, although it is never used outside of testing so perhaps we drop it? 

The main reason to leave it was to be able to use Kolda-style unfolding where the ordering in the Khatri-Rao is reversed. I'd be fine removing it.

aarmey:
Good points. Should be reworked here based on this discussion.

PR #502, opened 2023-04-19
More efficient error calculation in PARAFAC2

aarmey:
I have been using this, borrowed from Marie Roald's matcouply package. We currently reconstruct the entire dataset before calculating the error; this frequently takes up the most time in my profiling. The approach here avoids doing so by looking at the inner product, which ends up being much more efficient.

### DIFF
--- a/tensorly/decomposition/_parafac2.py
+++ b/tensorly/decomposition/_parafac2.py
@@ -6,12 +6,10 @@ from tensorly.random import random_parafac2
 from tensorly import backend as T
 from . import parafac, non_negative_parafac_hals
 from ..parafac2_tensor import (
-    parafac2_to_slice,
     Parafac2Tensor,
     _validate_parafac2_tensor,
 )
 from ..cp_tensor import CPTensor
-from ..base import unfold
 from ..tenalg.svd import svd_interface
 
 # Authors: Marie Roald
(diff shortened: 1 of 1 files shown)


JeanKossaifi (commented):
Looks great! Perhaps we can add a short comment (or even better, note in the doctoring) to explain what is done, like in https://github.com/tensorly/tensorly/blob/b648944fe94548765537a10cedd621871f7a2ffd/tensorly/decomposition/_cp.py#L217

aarmey:
Definitely—I will add a docstring with some description.

I can't for the life of me figure out why the `test_parafac2_normalize_factors` test consistently fails on just PyTorch with this change, given this only would affect the stopping criteria (and the fitting runs to the maximum number of iterations on all the backends)...

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -228,25 +229,26 @@ def test_parafac2_nn():
...
 
     random_parafac2_tensor = random_parafac2(
-        shapes=[(15, 30) for _ in range(25)], rank=rank, random_state=1234
+        shapes=[(15, 30) for _ in range(25)], rank=rank, random_state=rng
     )
     tensor = parafac2_to_tensor(random_parafac2_tensor)
     slices = parafac2_to_slices(random_parafac2_tensor)
 
     slice_rec = parafac2(
-        slices, rank, random_state=1234, normalize_factors=False, n_iter_max=2
```
aarmey, line 240: Because the decomposition was being set to the same seed as the random tensor generation, the starting point was exactly at the solution. This was causing the error to be very slightly negative (just due to roundoff) and causing a warning about a negative sqrt in the code. These changes make the test a little better, in principle, because it is not "cheating" by starting at the exact same random point. It also resolves the warning.

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -270,12 +272,13 @@ def test_parafac2_normalize_factors():
...
-    )
-    assert tl.max(tl.abs(T.norm(normalized_rec.factors[0], axis=0) - 1)) < 1e-5
-    assert abs(tl.max(norms) - tl.max(normalized_rec.weights)) / tl.max(norms) < 1e-2
-    assert abs(tl.min(norms) - tl.min(normalized_rec.weights)) / tl.min(norms) < 1e-2
+    normalized_rec = parafac2(slices, rank, random_state=rng, normalize_factors=True)
+
+    assert_array_almost_equal(T.norm(normalized_rec.factors[0], axis=0), tl.ones(rank))
+    assert_array_almost_equal(T.norm(normalized_rec.factors[1], axis=0), tl.ones(rank))
+    assert_array_almost_equal(T.norm(normalized_rec.factors[2], axis=0), tl.ones(rank))
+    assert abs(tl.max(norms) - tl.max(normalized_rec.weights)) / tl.max(norms) < 0.05
```
aarmey, line 280: The PyTorch failure was just an accuracy issue—probably primarily due to using float32. Adjusted here to compensate.

tensorly/backend/pytorch_backend.py
```diff
@@ -35,6 +35,10 @@ def context(tensor):
 
     @staticmethod
     def tensor(data, dtype=torch.float32, device="cpu", requires_grad=False):
+        if isinstance(data, torch.Tensor):
```
aarmey, line 38: This is just to silence a warning from PyTorch about not abusing `torch.tensor`.

tensorly/backend/pytorch_backend.py
```diff
@@ -200,12 +204,6 @@ def sign(tensor):
         """torch.sign does not support complex numbers."""
         return torch.sgn(tensor)
 
-    @staticmethod
-    def svd(matrix, full_matrices=True):
```
aarmey, line 204: This wasn't being used because we now import `torch.linalg.svd`.

tensorly/decomposition/tests/test_parafac2.py
JeanKossaifi: Great catch, this was a really problematic issue, basically making the test useless!

cohenjer: We may check other tests to see if the same issue happens elsewhere, I recall having a similar issue with admm on a local branch

Issue #503, opened 2023-05-24
Partial Tucker with PyTorch backend fails on GPU 


AtomicCactus:
#### Describe the bug
Decomposing a 2D tensor along both modes, while specifying two ranks (different or same) results in an error because internally there's a tensor created on the CPU as part of the process, which cannot be concatenated with the GPU.

Works fine when the rank is specified as an integer, but not as a list:

- `rank=16` works
- `rank=[16,16]` crashes

Works fine on the CPU, but performance is not the same.

#### Steps or Code to Reproduce

```python
import torch
import tensorly as tl
tl.set_backend('pytorch')

x = torch.randn((72, 27)).cuda()
result = tl.decomposition.partial_tucker(x, rank=[16, 32], modes=[0, 1])
```

#### Expected behavior

Tucker decomposition should not fail when ranks are provided as an array of values.

#### Actual result

```
Traceback (most recent call last):
  File "/home/yuri/Source/tensorly_bug.py", line 6, in <module>
    result = tl.decomposition.partial_tucker(x, rank=[16, 32], modes=[0, 1])
  File "/home/yuri/.local/lib/python3.10/site-packages/tensorly/decomposition/_tucker.py", line 166, in partial_tucker
    core, factors = initialize_tucker(
  File "/home/yuri/.local/lib/python3.10/site-packages/tensorly/decomposition/_tucker.py", line 64, in initialize_tucker
    U, _, _ = svd_interface(
  File "/home/yuri/.local/lib/python3.10/site-packages/tensorly/tenalg/svd.py", line 426, in svd_interface
    U, V = svd_flip(U, V, u_based_decision=u_based_flip_sign)
  File "/home/yuri/.local/lib/python3.10/site-packages/tensorly/tenalg/svd.py", line 42, in svd_flip
    signs = tl.concatenate((signs, tl.ones(tl.shape(V)[0] - tl.shape(U)[1])))
  File "/home/yuri/.local/lib/python3.10/site-packages/tensorly/backend/__init__.py", line 206, in wrapped_backend_method
    return getattr(
  File "/home/yuri/.local/lib/python3.10/site-packages/tensorly/backend/pytorch_backend.py", line 153, in concatenate
    return torch.cat(tensors, dim=axis)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument tensors in method wrapper_CUDA_cat)
```

#### Versions

```
[...]

PR #504, opened 2023-05-25
Fixed svd_flip() when used on GPU w/ PyTorch backend

AtomicCactus:
Fixes a problem where one of the tensors is being initialized on CPU and concatenated with a GPU tensor. See #503

### DIFF
--- a/tensorly/tenalg/svd.py
+++ b/tensorly/tenalg/svd.py
@@ -39,7 +39,7 @@ def svd_flip(U, V, u_based_decision=True):
         )
         U = U * signs
         if tl.shape(V)[0] > tl.shape(U)[1]:
-            signs = tl.concatenate((signs, tl.ones(tl.shape(V)[0] - tl.shape(U)[1])))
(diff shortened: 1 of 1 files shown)


tensorly/tenalg/svd.py
```diff
@@ -39,7 +39,7 @@ def svd_flip(U, V, u_based_decision=True):
         )
         U = U * signs
         if tl.shape(V)[0] > tl.shape(U)[1]:
-            signs = tl.concatenate((signs, tl.ones(tl.shape(V)[0] - tl.shape(U)[1])))
+            signs = tl.concatenate((signs, tl.ones(tl.shape(V)[0] - tl.shape(U)[1], device=signs.device)))
```
JeanKossaifi, line 42: We can directly pass `**tl.context(signs)` instead of just the device so the type is also preserved.

JeanKossaifi (commented):
Thanks @AtomicCactus! We can pass the full context instead. One option would be to test for different dtypes rather than devices.

aarmey:
Thanks for catching this, @AtomicCactus. Merging.

PR #505, opened 2023-05-26
Use math.pi

JeanKossaifi:
Solve a common issue in PyTorch.


cohenjer:
Looks ok to me, not sure what the error was, however.

JeanKossaifi:
It's an error only for people with an older version of PyTorch: I'm debating whether we need this at all. 

They get: 
`AttributeError: module 'torch' has no attribute 'pi'`

If we add this PR, I also need to explicitly check for torch version when I load `pi` it in the backend.

PR #511, opened 2023-07-10
Add randomized sampling-based ALS method for tensor ring decomposition

OsmanMalik:
This is an implementation of the algorithm TR-ALS-Sampled proposed in the paper
```
O. A. Malik, S. Becker. A sampling-based method for tensor ring decomposition. 
International Conference on Machine Learning (ICML), PMLR 139, pp. 7400-7411, 2021.
```
The code is written to mirror that of the deterministic ALS method for tensor ring decomposition in `tensor_ring_als` (in `tensorly/decomposition/_tr.py`) as closely as possible, but making appropriate modifications to make the method randomized. A wrapper class for the new function and unit tests have also been added.

The method extracts entries from tensors via lists of indices, e.g., `my_data_vector[my_index_vector]` to get a new vector whose i-th entry is `my_data_vector[my_index_vector[i]]`. Tensorflow doesn't seem to like this kind of indexing. The implementation therefore doesn't work with the Tensorflow backend (I added appropriate xfail statements for the unit tests). The function `randomised_parafac` uses a similar kind of indexing, and seems to have the same problem with Tensorflow since its tests also generate xfail messages when using the Tensorflow backend.

Any feedback is appreciated!


JeanKossaifi:
Thanks @OsmanMalik, this is a great feature! Quite busy at the moment but try to review soon - would be great if someone else can have a look @merajhashemi @cohenjer maybe?

tensorly/decomposition/_tr.py
```diff
@@ -268,6 +270,260 @@ def tensor_ring_als(
     return tr_decomp
 
 
+def compute_lev_score_dist(matrix):
```
JeanKossaifi, line 273: I would rename it just `leverage_score_dist` - the compute is implied and it's clearer than lev? Can we also add a reference? Would it be useful outside of TR? If so we could move it to `tensorly.metrics`

JeanKossaifi:
About the indexing issue we were discussing deprecating TensorFlow (and MXNet) anyway so your solution is great!

tensorly/decomposition/_tr.py
OsmanMalik: @JeanKossaifi Thank you for the suggestions! I updated the name and added a reference. This could be useful outside of TR (e.g., for sampled CP decomposition), so I moved the function to `tensorly.metrics`.

OsmanMalik:
@JeanKossaifi Is there anything additional I can do with this PR before it can be merged?

JeanKossaifi:
Thanks @OsmanMalik! It looks like the tests are failing

OsmanMalik:
> Thanks @OsmanMalik! It looks like the tests are failing

Thanks @JeanKossaifi for pointing that out. I have fixed the issue now and it looks like all the tests are passing.

OsmanMalik:
@JeanKossaifi Is there anything additional I can do before this can be merged?

JeanKossaifi:
Thanks @OsmanMalik -- apologies for the delay! It looks good to me! 
One minor thing is perhaps we want to split the tr into two files, one for each main algorithm but no strong opinion! 
Happy to merge!

OsmanMalik:
> nor thing is perhaps we want to split the tr into two files, one for each main algorithm but no strong opinion!
> Happy to merge!

@JeanKossaifi : Are you thinking one file for SVD-based algorithms and one for ALS-based ones? I think that would make sense.

JeanKossaifi:
Yes exactly!

OsmanMalik:
> Yes exactly!

@JeanKossaifi : I've split up `_tr.py` into two files `_tr_svd.py` and `_tr_als.py`. I split the pytest scripts in a similar fashion.

The failing test for numpy 3.11 looks unrelated to my updates.

JeanKossaifi:
Looks great, thank you @OsmanMalik! It might be worse adding a small interactive [example](https://github.com/tensorly/tensorly/tree/main/examples/decomposition) to showcase the new methods?

Issue #448, opened 2022-09-05
Improving the test suite


MarieRoald:
@yngvem and I have discussed the test suite a bit, and we have thought of some ways that might improve it. We don't need to adress all these points in one go, but we could try to be mindful about it for new additions and maybe incrementally split up the tests we have into smaller tests? What do you guys think? 

### Test structure
Currently, TensorLy mainly has monolithic tests that check many things at once. However, best practices are splitting large tests into smaller tests that (ideally) only test one thing at a time. This will make it easier to see what fails and what works.

### Better use of fixtures
If we split the monolithic tests into smaller tests, we should probably use fixtures more actively to avoid having to rewrite the same boilerplate code for each of the smaller test functions. So, for example, we could have fixtures for the different decompositions.

### Clearer separation of unit tests and integration tests
There is currently little difference between unit tests (which test a single unit of code) and integration tests (which test how several code units work together). It can be challenging to create a hard line separating the different types of tests, but it is a good strategy to be mindful of the difference and have both. 

### Better handling of random numbers
Currently, we use fixed random seeds. We can use [pytest-randomly](https://github.com/pytest-dev/pytest-randomly) to get new random seeds every time we run the test suite, with the possibility to specify the seed manually and re-run the last seed from added command line arguments to the `pytest` command.

PR #513, opened 2023-07-22
Miscellaneous housekeeping improvements

aarmey:
Apologies that this does not involve just one change. This resolves some various minor inconsistencies, deprecations, and performance issues:

- In both the PLSR and HALS implementations, this reduces indexing which improves performance, particularly for Jax and TensorFlow
- Some methods were not being exported or had missing arguments in their docstrings
- Most backends provide the `np.lstsq` interface, so all four return values are passed on now
- As suggested in #448, this adds `pytest-randomly`. This sets the random seeds of the default random generators and reports it in output. I've found this is very helpful when dealing with inconsistent errors.
- Python v3.8 is now EOL, so I removed some code ensuring its support. The default Python version is now 3.11 in testing.

I will further annotate each of the changes so that this is simpler to review.


tensorly/__init__.py
```diff
@@ -58,16 +58,21 @@
...
     transpose,
     moveaxis,
     arange,
+    any,
     ones,
     zeros,
     zeros_like,
     eye,
     where,
+    conj,
```
aarmey, line 73: These methods were not being exported, and I didn't see a reason for being left out.

tensorly/backend/core.py
```diff
@@ -527,30 +544,36 @@ def clip(tensor, a_min=None, a_max=None):
...
         -------
-        scalar
+        scalar or tensor
+            If axis is None, returns a scalar. Otherwise, returns a tensor of scalars.
         """
         raise NotImplementedError
 
     @staticmethod
-    def min(tensor):
+    def min(tensor, axis=None):
```
aarmey, line 564: The axis argument is actively used, so it should be represented here.

tensorly/base.py
```diff
@@ -1,5 +1,5 @@
+from math import prod
 from . import backend as tl
-from .utils import prod
```
aarmey, line 2: `prod` was only here for Python v3.8 support.

tensorly/decomposition/_cp.py
```diff
@@ -426,8 +426,6 @@ def parafac(
                 tl.solve(tl.conj(tl.transpose(pseudo_inverse)), tl.transpose(mttkrp))
             )
             factors[mode] = factor
-            if normalize_factors and mode != modes_list[-1]:
```
aarmey, line 429: I think the decision to do this only at the end came out of #264, but was in a closed PR. I need to dig it up.

tensorly/decomposition/tests/test_cmtf_als.py
```diff
@@ -10,8 +10,8 @@
 def test_coupled_matrix_tensor_3d_factorization():
     I = 21
     J = 12
-    K = 8
-    M = 7
+    K = 16
+    M = 14
```
aarmey, line 14: To help with occasional testing failures.

tensorly/regression/cp_plsr.py
```diff
@@ -122,24 +121,18 @@ def fit(self, X, Y):
...
-                        init=Z_comp_CP,
-                        tol=None,
-                        normalize_factors=False,
-                        n_iter_max=10,
-                    )
-
-                    # Normalize separately—this was profiling very slow in parafac
-                    Z_comp = cp_normalize(Z_comp_CP)[1]
+                    for mode in range(len(Z_comp)):
+                        factor = multi_mode_dot(Z, Z_comp, skip=mode)
```
aarmey, line 133: Tucker is equivalent to CP when rank 1. This is more performant, and this spot was consistently one of the slowest parts of testing, particularly for indexing-slow backends.

tensorly/tenalg/proximal.py
```diff
@@ -987,37 +987,23 @@ def hals_nnls(
...
 
-                else:  # without sparsity
-                    deltaV = tl.where(
-                        (UtM[k, :] - tl.dot(UtU[k, :], V)) / UtU[k, k] > -V[k, :],
-                        (UtM[k, :] - tl.dot(UtU[k, :], V)) / UtU[k, k],
-                        -V[k, :],
-                    )
-                    V = tl.index_update(V, tl.index[k, :], V[k, :] + deltaV)
+                # Modifying the function for sparsification
+                if sparsity_coefficient is not None:
```
aarmey, line 993: This code should be unchanged, but (1) reuses some indexing results for indexing-slow backend performance, and (2) I find it a bit more readable by removing repetitive segments and breaking the expressions up into logical units.

tensorly/tests/test_tr_tensor.py
```diff
@@ -46,24 +46,10 @@ def test_tr_to_tensor():
...
-                    tl.dot(factors[0][:, i, :], factors[1][:, j, :]),
-                    factors[2][:, k, :],
-                )
-                # TODO: add trace to backend instead of this
-                tensor = tl.index_update(
-                    tensor,
-                    tl.index[i, j, k],
-                    tl.sum(product * tl.eye(product.shape[0])),
-                )
+    tensor = tl.einsum("iaj,jbk,kci->abc", *factors)
```
aarmey, line 49: We have trace in the backend now (see TODO comment), but I thought this was cleaner.

JeanKossaifi (commented):
Thanks @aarmey! Always great to have improvements and more readable code!

tensorly/regression/cp_plsr.py
aarmey: Here, I think this mostly just came down to calling CP, with a lot of extra code contained within, when it is not strictly needed. This call is inside two loops, and so it can be called thousands of times. Within the profiling, I believe most of the time was spent on the MTTKRP, but other parts of the CP call weren't negligible.

tensorly/decomposition/_tr.py
```diff
@@ -216,7 +216,7 @@ def tensor_ring_als(
 
             if ls_solve == "lstsq":
                 # Solve least squares problem directly
-                sol, _ = tl.lstsq(design_mat, tensor_unf)
+                sol = tl.lstsq(design_mat, tensor_unf)[0]
```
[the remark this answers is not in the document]
aarmey: I didn't know `*_` was possible! Agreed this will be more readable.

PR #514, opened 2023-07-25
feat: add `pip` caching to CI

SauravMaheshkar:
Changes made by this PR can be summarized as follows:

- Set the `cache` and `cache-dependency-path` argument to enable caching to speed up CI times.

Request for Review: @JeanKossaifi


SauravMaheshkar:
Huh, it seems like coverage is failing, but this PR simply changes the CI not the codebase, maybe it's just a codecov error ?

SauravMaheshkar:
Yes @JeanKossaifi, For reference https://github.com/actions/setup-python#caching-packages-dependencies, outlines that cache won't be restored if a newer package version is available. In case of a new version, it downloads it and updates the cache.

SauravMaheshkar:
Gentle ping @JeanKossaifi

JeanKossaifi:
Thanks @SauravMaheshkar !
It looks good to me - the test doesn't use the doc requirements. Perhaps we can update the PR cache only reqs here and add doc reqs caching to the doc CI script?

SauravMaheshkar:
Makes sense, addressed in e9970122ec073be6b2a3b03256dadea123cd4e53

SauravMaheshkar:
Looks like everything passed 😄

JeanKossaifi:
Thanks @SauravMaheshkar, looks great to me! Could you maybe add a small note in the documentation about this for future users/maintainers?

SauravMaheshkar:
> Thanks @SauravMaheshkar, looks great to me! Could you maybe add a small note in the documentation about this for future users/maintainers?

Yeah sure, will do

PR #517, opened 2023-08-02
Also provide H back from make_svd_non_negative()

### LINKED ISSUE #515: make_svd_non_negative only returns the updated U matrix
#### Describe the bug
I'm not sure if this is a bug or if this is actually how this function is supposed to behave but currently, it only returns the updated U value. Incase it's not a bug, I don't understand why the H matrix is computed in this function (https://github.com/tensorly/tensorly/blob/6c7ecfbfbfc088cb1a423e6a5970e19565844651/tensorly/tenalg/svd.py#L65)

Looking forward to your response.

aarmey:
Fixes #515

### DIFF
--- a/tensorly/tenalg/svd.py
+++ b/tensorly/tenalg/svd.py
@@ -119,15 +119,17 @@ def make_svd_non_negative(tensor, U, S, V, nntype=True):
 
     if nntype == "nndsvd":
         W = soft_thresholding(W, eps)
+        H = soft_thresholding(H, eps)
     elif nntype == "nndsvda":
         avg = tl.mean(tensor)
         W = tl.where(W < eps, tl.ones(tl.shape(W), **tl.context(W)) * avg, W)
+        H = tl.where(H < eps, tl.ones(tl.shape(H), **tl.context(H)) * avg, H)
     else:
         raise ValueError(
             f'Invalid nntype parameter: got {nntype} instead of one of ("nndsvd", "nndsvda")'
         )
 
-    return W
+    return W, H
 
 
 def randomized_range_finder(A, n_dims, n_iter=2, random_state=None):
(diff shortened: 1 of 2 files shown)


hello-fri-end:
Hello @aarmey! Maybe we should loop till len(S) in the following line since the no. of columns of U may not be equal to the number of rows of V:
https://github.com/tensorly/tensorly/blob/b99cb2c4909db2e2285c460d66eec155ca2be704/tensorly/tenalg/svd.py#L90

JeanKossaifi:
Thank you @aarmey - it looks good to me!

@hello-fri-end that's a good point but it shouldn't matter - that potential "extra" block should be zero and associated with just a 1 on the diagonal of S right?

hello-fri-end:
> Thank you @aarmey - it looks good to me!
> 
> @hello-fri-end that's a good point but it shouldn't matter - that potential "extra" block should be zero and associated with just a 1 on the diagonal of S right?

Hey @JeanKossaifi, thanks for the reply! Assuming, `U`, `S`,  and `V` are the output of economy SVD, it will never fail. I was just thinking if `U.shape == ( m x m )`,` S.shape == ( m x n )`  and `V.shape == ( n  x n )`, given a matrix of shape `(m, n)` with `m > n`, then `V[j, :]` might throw an error in the following line:
https://github.com/tensorly/tensorly/blob/b99cb2c4909db2e2285c460d66eec155ca2be704/tensorly/tenalg/svd.py#L90-L91

JeanKossaifi:
I see - this is a good point we don't know what was used to get U, S, V, should we loop until min(m, n) then? 
Could you maybe add a test case for this?

aarmey:
I've fixed this issue, as well as #522, and added a test.

JeanKossaifi:
Thanks a lot @aarmey! Looks great, merging.

aarmey:
Thanks, @JeanKossaifi. Looks like there is a typing error with mxnet. I can fix that and then merge.

JeanKossaifi:
Yes I just noticed :)

PR #524, opened 2023-08-21
PARAFAC2: Avoid reprojection of X during the error calculation

aarmey:
In many (most?) applications of PARAFAC2, the slices of X are much larger than the rank along both dimensions. For very large slices, the majority of time is spent on any operations that touch the full X. This necessarily occurs in three places:

1. SVD initialization (optionally).
2. Solving for the projections matrices.
3. Projecting X into the projected tensor.
4. The reconstruction error calculation.

With 4, this can be avoided because the inner product operates with the projected tensor. Additionally, the second path of the previous if statement would only be used if the rank is greater than the size along C. In practice, this does not generally happen and, if it did, would be for a small dataset. Consequently, I removed the previous if statement and replaced it with this optional reuse of the projected tensor.

For _very_ large slices of X (e.g., 4,000 x 8,000) this can save ~30% of the runtime because these three later multiplications dominate. It also tends to help more so with a line search routine, because you add an additional OP and reconstruction error calculation at the expense of avoiding another round of CP.

### DIFF
--- a/tensorly/decomposition/_parafac2.py
+++ b/tensorly/decomposition/_parafac2.py
@@ -93,7 +93,9 @@ def _project_tensor_slices(tensor_slices, projections):
     return tl.stack(slices)
 
 
-def _parafac2_reconstruction_error(tensor_slices, decomposition, norm_matrices=None):
+def _parafac2_reconstruction_error(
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
Awesome, thanks @aarmey, merging!

PR #525, opened 2023-09-04
Bro's line search for PARAFAC2

aarmey:
This implements Bro's line search strategy for PARAFAC2. I know that this introduces a non-trivial amount of additional complexity. However, PARAFAC2 often undergoes a large number of iterations because CP is embedded within the Orthogonal Procrustes projection. In my experience, the benefits from line search have been even greater than with CP—frequently 5–10X faster convergence because essentially no line search steps are rejected.


MarieRoald:
Just one question: Which line search method are you using? In [[1]], Rasmus compares different extrapolation algorithms, and from a quick glance at the code, it seems like you implemented the method that Rasmus calls *PSL_Toolbox line search*? Is this correct? I'm wondering because his benchmark seems to show that what he calls *All-at-once Nesterov-like extrapolation* (which is a variation of the extrapolation scheme in [[2]]) could be faster.

I'll try to review the code properly on Tuesday!

PS. On a slightly related note, I'm also working on speeding up PARAFAC2 but specifically for cases whenever J ≪ K (for an I × J × K tensor) Essentially, we can fit the PARAFAC2 model based on only the right singluar vectors of the X_i matrices without losing any accuracy. So by computing one SVD at the beginning of the fitting process, we can compress the data matrices so we have an I × K × K tensor instead of an I × J × K tensor. 

[1]: https://www.sciencedirect.com/science/article/abs/pii/S0169743921000800
[2]: https://hal.science/hal-02143969/document

aarmey:
Thanks @MarieRoald! Yes, I think that this is the PLS Toolbox line search implementation. (Though I don't have access this week to check.) It is also the line search implementation developed within Rasmus's thesis, if that helps. There are most certainly faster implementations, but I felt like this might be a good trade off of speed and complexity for TensorLy.

I would be _very_ interested to try out the advancement you mention! While this line search has given us sufficient performance with a high-end GPU, we have an application for which many folks will want to use a CPU implementation. Also, we're bumping up against problems due to datasets that are >150 GB and no longer fit into RAM. Using randomized SVD as a first step, and then working off of that, would be an outstanding improvement.

JeanKossaifi:
As a side note, I think we should bring in incremental/online (batched) SVD to trade memory for compute when that is the issue, especially on GPU.

tensorly/decomposition/_parafac2.py
```diff
@@ -348,27 +365,59 @@ def parafac_updates(X, w, f):
...
+
+        # Start line search if requested.
+        if line_iter:
+            jump = iteration ** (1.0 / acc_pow)
+
+            ls_weights = weights_last + (weights - weights_last) * jump
+            ls_factors = [
+                factors_last[ii] + (factors[ii] - factors_last[ii]) * jump
+                for ii in range(3)
+            ]
```
MarieRoald:
```suggestion
            ls_factors = [
                factors_last[ii] + (factors[ii] - factors_last[ii]) * jump
                for ii, _ in enumerate(factors)
            ]
```
Maybe use enumerate so it's clear that this happens with all factor matrices?

tensorly/decomposition/_parafac2.py
```diff
@@ -348,27 +365,59 @@ def parafac_updates(X, w, f):
...
+
+            # Clip if the mode should be non-negative
+            if nn_modes:
+                if 0 in nn_modes:
+                    ls_factors[0] = tl.abs(ls_factors[0])
+                if 2 in nn_modes:
+                    ls_factors[2] = tl.abs(ls_factors[2])
```
MarieRoald:
```suggestion
                if 0 in nn_modes:
                    ls_factors[0] = tl.clip(ls_factors[0], 0)
                if 2 in nn_modes:
                    ls_factors[2] = tl.clip(ls_factors[2], 0)
```

Was there a reason to use `tl.abs` instead of `tl.clip`?

tensorly/decomposition/_parafac2.py
```diff
@@ -348,27 +365,59 @@ def parafac_updates(X, w, f):
...
+
+            ls_rec_error = _parafac2_reconstruction_error(
+                tensor_slices, (weights, factors, projections), norm_tensor
+            )
```
MarieRoald: In Rasmus' recent paper, [[1]], he says that we should also update the projection matrices before computing the reconstruction error for the line search. Have you tested that to see if it makes much of a difference?

Here's the quote from [[1]]:
> In all investigated PARAFAC2 algorithms, matrices A, C and the common matrix U are extrapolated. The orthogonal matrix Pk is not to be extrapolated. Instead, after each extrapolation step, the Pk matrices are calculated according to Algorithm 1 - Step 2 and used for calculating the PARAFAC2 loss function error in order to judge the extrapolation step in the loop.

[1]: https://www.sciencedirect.com/science/article/abs/pii/S0169743921000800

MarieRoald (commented):
Overall, this looks very nice! My main comment is that it would maybe be preferable if we could have the line search procedure in its own function. Maybe something like:

```python
PLS_LINE_SEARCH_DEFAULTS = dict(
     acc_pow=2.0, # Extrapolate to the iteration^(1/acc_pow) ahead  
     acc_fail=0, # How many times acceleration has failed	 
     max_fail=4, # Increase acc_pow with one after max_fail failure
 )

def _init_parafac2_line_search(line_search):
    if line_search is None:
        return {}
    elif line_search == "PLS_Toolbox":
        return PLS_LINE_SEARCH_DEFAULTS.copy()
    else:
        raise ValueError(f"Invalid line search value: {line_search}")

def _parafac2_line_search(weights, factors, weights_last, factors_last, rec_error, line_search_params):
    ...
    # Code to do the extrapolation 
    return weights_ls, factors_ls, rec_error, line_search_params
```

I think having such a function could make the code more readable, easier to maintain and test, and it could also make it easier to add implementations of other line search algorithms later. What do you think?

aarmey:
Great stuff. Thanks @MarieRoald! I'll work on finishing this next week.

cohenjer:
Happy to review this as well when it's ready!

tensorly/decomposition/_parafac2.py
```diff
@@ -348,27 +365,59 @@ def parafac_updates(X, w, f):
...
-                    tl.ones(tl.shape(norms), **tl.context(factors[0])),
-                    norms,
-                )
+            weights, factors = cp_normalize((weights, factors))
+
+        # Start line search if requested.
+        if line_iter:
+            jump = iteration ** (1.0 / acc_pow)
+
+            ls_weights = weights_last + (weights - weights_last) * jump
```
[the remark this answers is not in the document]
aarmey: I think so. By this point, we have run the weights and factors at least once through PARAFAC, which should fill these in as ones.

tensorly/decomposition/_parafac2.py
aarmey: Yes, there is a call to compute_projections doing this just before. (I fixed this in a more recent commit, so you may have reviewed code without it.) If you don't do this, the line search is almost always rejected, so it is indeed necessary.

aarmey:
> Overall, this looks very nice! My main comment is that it would maybe be preferable if we could have the line search procedure in its own function. Maybe something like:
> 
> ```python
> PLS_LINE_SEARCH_DEFAULTS = dict(
>      acc_pow=2.0, # Extrapolate to the iteration^(1/acc_pow) ahead  
>      acc_fail=0, # How many times acceleration has failed	 
>      max_fail=4, # Increase acc_pow with one after max_fail failure
>  )
> 
> def _init_parafac2_line_search(line_search):
>     if line_search is None:
>         return {}
>     elif line_search == "PLS_Toolbox":
>         return PLS_LINE_SEARCH_DEFAULTS.copy()
>     else:
>         raise ValueError(f"Invalid line search value: {line_search}")
> 
> def _parafac2_line_search(weights, factors, weights_last, factors_last, rec_error, line_search_params):
>     ...
>     # Code to do the extrapolation 
>     return weights_ls, factors_ls, rec_error, line_search_params
> ```
> 
> I think having such a function could make the code more readable, easier to maintain and test, and it could also make it easier to add implementations of other line search algorithms later. What do you think?

I like the idea of refactoring out the line search so that this could be adjusted. Ideally, the code here would also be shared with the line search in CP—the approach is almost identical. Perhaps we can settle on an interface and then introduce this in a separate PR, though? I am not sure what the best interface looks like. For example, here, it seems like the type of line search needs to be an entry in the parameter dict, so that the line search routine can look up what it needs to run. Also, we would probably need a _parafac2_line_search(), which then would call some _PLS_toolbox_line_search() to interface with the parameters.

Again, I really like this idea, because line search could then be a capability across almost all the decomposition methods, but part of that will be getting the interface exactly right.

aarmey:
@MarieRoald @cohenjer, this should be all set to review again.

tensorly/decomposition/_parafac2.py
```diff
@@ -340,35 +348,78 @@ def parafac_updates(X, w, f):
...
+        # Will we be performing a line search iteration?
+        if linesearch and iteration % 2 == 0 and iteration > 5:
+            line_iter = True
+            factors_last = [tl.copy(f) for f in factors]
+            factors_last[1] = factors_last[1] * T.reshape(weights, (1, -1))
+        else:
+            line_iter = False
+
         factors[1] = factors[1] * T.reshape(weights, (1, -1))
         weights = T.ones(weights.shape, **tl.context(tensor_slices[0]))
```
MarieRoald, line 361:
```suggestion
        factors[1] = factors[1] * T.reshape(weights, (1, -1))
        weights = T.ones(weights.shape, **tl.context(tensor_slices[0]))
        
        # Will we be performing a line search iteration?
        if linesearch and iteration % 2 == 0 and iteration > 5:
            line_iter = True
            factors_last = [tl.copy(f) for f in factors]
        else:
            line_iter = False
```

Could we change here the order to avoid multiplying in the weights twice?

MarieRoald (commented):
I see your point, and it would be really nice to get a unified API for CP and PARAFAC2 line-search! However, I'm wary of increasing the complexity of the PARAFAC2 code further now, with the intent to improve it later. Instead, I think maybe the best solution is to arrive at a very simple way to abstract the PARAFAC2 line-search into a function for now. This function could be even simpler than what I sketched out above, e.g. just having a `_bro_thesis_line_search()` function. We shouldn't let the perfect be the enemy of the good 🙂

I think the PARAFAC2 implementation will be more readable and easier to follow for new users if it doesn't have a large block of line-search logic in the middle of it. And it will also make it much easier to test the line-search implementation thoroughly. For example, if we have a separate line-search function, then we could mock the `_parafac2_reconstruction_error`-function to control whether a line-search should be accepted and test whether:

- line-searches are accepted if and only if the reconstruction error decreases
- clipping is implemented correctly 
- acc_fail is increased when line-searches are rejected
- no change occurs if factors = factors_last

Also, the way I see it, we can always move from a PARAFAC2-specific to a more general implementation that also works for CP later. Since it is a purely internal API, it will not affect the end user if we change it.

aarmey:
@MarieRoald good points. I will make an attempt at separating out the line search code as you suggest. Thanks!

aarmey:
@MarieRoald I am interested in what you think about this code structure. It seems more natural to me to include the line search state within a class, as opposed to a separate dictionary. There are also several quantities that are static during the fitting process and so are only passed along once to reduce the number of arguments. Depending on how this structure looks, I will add the various documentation and then let you review.

tensorly/decomposition/_parafac2.py
```diff
@@ -95,6 +95,70 @@ def _project_tensor_slices(tensor_slices, projections):
     return tl.stack(slices)
 
 
+class _bro_thesis_line_search:
```
MarieRoald, line 98:
```suggestion
class _BroThesisLineSearch:
```

Let's use the CapWords convention mentioned in [PEP 8](https://peps.python.org/pep-0008/#class-names).

tensorly/decomposition/_parafac2.py
```diff
@@ -95,6 +95,70 @@ def _project_tensor_slices(tensor_slices, projections):
...
+    def __init__(self, norm_tensor, svd, verbose=False, nn_modes=None):
+        self.norm_tensor = norm_tensor
+        self.svd = svd
+        self.verbose = verbose
+        self.acc_pow = 2.0  # Extrapolate to the iteration^(1/acc_pow) ahead
+        self.acc_fail = 0  # How many times acceleration have failed
+        self.max_fail = 4  # Increase acc_pow with one after max_fail failure
+        self.nn_modes = nn_modes
```
MarieRoald, line 106:
```suggestion
    def __init__(self, norm_tensor, svd, verbose=False, nn_modes=None, acc_pow=2.0, max_fail=4):
        self.norm_tensor = norm_tensor
        self.svd = svd
        self.verbose = verbose
        self.acc_pow = acc_pow  # Extrapolate to the iteration^(1/acc_pow) ahead
        self.max_fail = max_fail  # Increase acc_pow with one after max_fail failure
        self.acc_fail = 0 # How many times acceleration have failed
        self.nn_modes = nn_modes
```

Maybe we want to add the possibility of setting different starting parameters?

tensorly/decomposition/_parafac2.py
```diff
@@ -303,9 +367,10 @@ def parafac2(
     if absolute_tol is None:
         absolute_tol = tl.eps(factors[0].dtype) * 1000
 
-    acc_pow = 2.0  # Extrapolate to the iteration^(1/acc_pow) ahead
-    acc_fail = 0  # How many times acceleration have failed
-    max_fail = 4  # Increase acc_pow with one after max_fail failure
+    if linesearch:
+        line_search = _bro_thesis_line_search(
+            norm_tensor, svd, verbose=verbose, nn_modes=nn_modes
+        )
```
MarieRoald, line 373:
```suggestion
if linesearch and not isinstance(linesearch, _BroThesisLineSearch):
    linesearch = _BroThesisLineSearch(
        norm_tensor, svd, verbose=verbose, nn_modes=nn_modes
    )
```

It could be nice to support sending in a line search object (keeping this option undocumented so it's not officially a part of the public API yet). Then we also don't need a separate variable for the `_BroThesisLineSearch` instance. What do you think?

tensorly/decomposition/_parafac2.py
```diff
@@ -365,48 +430,15 @@ def parafac_updates(X, w, f):
...
-                if 0 in nn_modes:
-                    factors_ls[0] = tl.clip(factors_ls[0], 0)
-                if 2 in nn_modes:
-                    factors_ls[2] = tl.clip(factors_ls[2], 0)
-
-            projections_ls = _compute_projections(tensor_slices, factors_ls, svd)
-
-            ls_rec_error = _parafac2_reconstruction_error(
-                tensor_slices, (weights, factors_ls, projections_ls), norm_tensor
+            factors, projections, rec_errors[-1] = line_search.line_step(
```
MarieRoald, line 433:
```suggestion
            factors, projections, rec_errors[-1] = linesearch.line_step(
```

See comment above

MarieRoald (commented):
It looks really good so far! I think it's a lot easier to follow the logic now, and I agree that it might be more natural as a class 🙂

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -116,7 +120,43 @@ def test_parafac2(monkeypatch, normalize_factors, init):
...
+        rank,
+        init="svd",
+        return_errors=True,
+        n_iter_max=10,
+        linesearch=True,
+    )
+    ls_error = err[-1]
+    assert ls_error < standard_error
+
+
```
MarieRoald, line 157: A nice thing with having the line search this way is that it' a lot easier to test its behaviour in an isolated way. Here's one example with a test that fails if the line search accepts a decomposition that worsens the error:


```suggestion
def test_linesearch_accepts_only_improved_fit():
    rng = tl.check_random_state(123)
    rank = 4

    weights, factors, projections = random_parafac2(
        shapes=[(25 + rng.randint(5), 300) for _ in range(15)],
        rank=rank,
        random_state=rng,
    )
    slices = parafac2_to_slices((weights, factors, projections))

    # Create dummy variable for the previous iteration
    previous_iteration = random_parafac2(
        shapes=[(25 + rng.randint(5), 300) for _ in range(15)],
        rank=rank,
        random_state=rng,
    )

    # Test with line search where the reconstruction error would worsen if accepted
    line_search = _BroThesisLineSearch(norm_tensor=1, svd="truncated_svd")
    ls_factors, ls_projections, ls_rec_error = line_search.line_step(
        iteration=10,
        tensor_slices=slices,
        factors_last=previous_iteration[1],
        weights=weights,
        factors=factors,
        projections=projections,
        rec_error=0,
    )

    # Assert that the factor matrices, projection and reconstruction error all
    # are unaffected by the line search
    for fm, ls_fm in zip(factors, ls_factors):
        assert_array_almost_equal(fm, ls_fm)
    for proj, ls_proj in zip(projections, ls_projections):
        assert_array_almost_equal(proj, ls_proj)
    assert ls_rec_error == 0
    assert line_search.acc_fail == 1

    # Test with line search where the reconstruction error would improve if accepted
    line_search = _BroThesisLineSearch(norm_tensor=1, svd="truncated_svd")
    ls_factors, ls_projections, ls_rec_error = line_search.line_step(
        iteration=10,
        tensor_slices=slices,
        factors_last=previous_iteration[1],
        weights=weights,
        factors=factors,
        projections=projections,
        rec_error=float('inf'),  # float('inf') to force accepting the line search
    )
    # Assert that the factor matrices, projection and reconstruction error all
    # are changed by the line search
    for fm, ls_fm in zip(factors, ls_factors):
        assert_(tl.norm(fm - ls_fm) > 1e-5)
    for proj, ls_proj in zip(projections, ls_projections):
        assert_(tl.norm(proj - ls_proj) > 1e-5)
    assert 0 < ls_rec_error < float('inf')
    assert line_search.acc_fail == 0


```

MarieRoald (changes requested):
Looks good to me, only a minor nitpick about the reference in the docstring and I added a suggestion for an additional test 🙂

aarmey:
Thanks @MarieRoald! I incorporated these changes. Merging!

cohenjer:
Sorry I didn't find the time to review before merging; I will have a look in the coming weeks and raise issues if I see some problems!

aarmey:
Thanks @cohenjer! I'm happy to make adjustments if you come across any issues.

PR #530, opened 2023-09-23
SVD compression of PARAFAC2

MarieRoald:
A big downside with PARAFAC2 is that it is slow. This is exacerbated by the local minima issue, which means that we need to fit models with several different initialisations to avoid ending up in a local minima.

However, we can speed up the fitting procedure drastically in two cases: when J ≪ K (for an I × J × K tensor) or when the I horisontal slices have approximately low rank. In these cases, we can preprocess the data by computing the singular value decomposition, which we can think of as a data compression step. Then, we fit the PARAFAC2 model to the right singular vectors multiplied by the singular values, which will have a dimensionality of K × K (or L × K, with L < K if the data is approximately low rank). Finally, to decompress the decomposition to model the original, uncompressed data, we left-multiply the B_i-matrices with the left singular vectors of the data matrices.

However, we can speed up the fitting procedure drastically in two cases: when J ≪ K (for an I × J × K tensor) or when the I lateral slices have approximately low rank. In these cases, we can preprocess the data by computing the singular value decomposition. Then, we fit the PARAFAC2 model to the right singular vectors multiplied by the singular values. This is a data compression step since these matrices will have a size K × K (or L × K, with L < K if the data is approximately low rank and we truncate the SVD). Finally, to decompress the decomposition to model the original, uncompressed data, we left-multiply the B_i-matrices with the left singular vectors of the data matrices.

For more information about how this works, see the documentation of `tensorly.decomposition.preprocessing.svd_compress_tensor_slices` from this PR. This code also includes an example for the documentation demonstrating how to use this compression scheme with PARAFAC2.

**Note:**
Since we would typically like to do this once and share the compressed data for each initialisation (and each model in a grid search), we thought it made more sense to keep this in a separate preprocessing module instead of adding more flags to the PARAFAC2 function. This workflow makes it more apparent to the user that this step should be done once and cached, which is important since the SVD of the tensor slices can be time-consuming.

### DIFF
--- a/examples/decomposition/plot_parafac2_compression.py
+++ b/examples/decomposition/plot_parafac2_compression.py
@@ -1,0 +1,293 @@
+"""
+Speeding up PARAFAC2 with SVD compression
+=========================================
+
+PARAFAC2 can be very time-consuming to fit. However, if the number of rows greatly
+exceeds the number of columns or the data matrices are approximately low-rank, we can
+compress the data before fitting the PARAFAC2 model to considerably speed up the fitting
+procedure.
+
+The compression works by first computing the SVD of the tensor slices and fitting the
+PARAFAC2 model to the right singular vectors multiplied by the singular values. Then,
+after we fit the model, we left-multiply the :math:`B_i`-matrices with the left singular
+vectors to recover the decompressed model. Fitting to compressed data and then
+decompressing is mathematically equivalent to fitting to the original uncompressed data.
+
+For more information about why this works, see the documentation of
+:py:meth:`tensorly.decomposition.preprocessing.svd_compress_tensor_slices`.
+"""
+from time import monotonic
+import tensorly as tl
+from tensorly.decomposition import parafac2
+import tensorly.preprocessing as preprocessing
+
+
+##############################################################################
+# Function to create synthetic data
+# ---------------------------------
+#
+# Here, we create a function that constructs a random tensor from a PARAFAC2
+# decomposition with noise
+
+rng = tl.check_random_state(0)
+
+
+def create_random_data(shape, rank, noise_level):
+    I, J, K = shape  # noqa: E741
+    pf2 = tl.random.random_parafac2(
+        [(J, K) for i in range(I)], rank=rank, random_state=rng
+    )
+
+    X = pf2.to_tensor()
+    X_norm = [tl.norm(Xi) for Xi in X]
+
+    noise = [rng.standard_normal((J, K)) for i in range(I)]
+    noise = [noise_level * X_norm[i] / tl.norm(E_i) for i, E_i in enumerate(noise)]
+    return [X_i + E_i for X_i, E_i in zip(X, noise)]
+
+
+##############################################################################
+# Compressing data with many rows and few columns
+# -----------------------------------------------
+#
+# Here, we set up for a case where we have many rows compared to columns
+
+n_inits = 5
+rank = 3
+shape = (10, 10_000, 15)  # 10 matrices/tensor slices, each of size 10_000 x 15.
+noise_level = 0.33
+
+uncompressed_data = create_random_data(shape, rank=rank, noise_level=noise_level)
+
+##############################################################################
+# Fitting without compression
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# As a baseline, we see how long time it takes to fit models without compression.
+# Since PARAFAC2 is very prone to local minima, we fit five models and select the model
+# with the lowest reconstruction error.
+
+print("Fitting PARAFAC2 model without compression...")
+t1 = monotonic()
+lowest_error = float("inf")
+for i in range(n_inits):
+    pf2, errs = parafac2(
+        uncompressed_data,
+        rank,
+        n_iter_max=1000,
+        nn_modes=[0],
+        random_state=rng,
+        return_errors=True,
+    )
+    if errs[-1] < lowest_error:
+        pf2_full, errs_full = pf2, errs
+t2 = monotonic()
+print(
+    f"It took {t2 - t1:.1f}s to fit a PARAFAC2 model a tensor of shape {shape} "
+    + "without compression"
+)
+
+##############################################################################
+# Fitting with lossless compression
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Since the tensor slices have many rows compared to columns, we should be able to save
+# a lot of time by compressing the data. By compressing the matrices, we only need to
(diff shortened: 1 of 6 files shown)


aarmey:
This is outstanding! Thanks @MarieRoald. I'll review this soon.

@JacksonLChin—If you have time to apply this for the hackathon, this is a better solution for making your dataset into a manageable size. It'd also be helpful to hear about any snags you hit when running it.

examples/decomposition/plot_parafac2_compression.py
```diff
@@ -0,0 +1,280 @@
...
+##############################################################################
+# We see that we saved a lot of time by compressing the data before fitting the model.
+
+##############################################################################
+# Fitting with lossy compression
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# 
+# We can try to speed the process up even further by accepting a slight discrepancy 
+# between the model obtained from compressed data and a model obtained from uncompressed
+# data.Specifically, we can truncate the singular values at some threshold, essentially
```
aarmey, line 135:
```suggestion
# data. Specifically, we can truncate the singular values at some threshold, essentially
```

tensorly/decomposition/_parafac2.py
```diff
@@ -115,6 +117,9 @@ def _parafac2_reconstruction_error(tensor_slices, decomposition, norm_matrices=N
         * projections : List of projection matrices used to create evolving factors.
     norm_matrices : float, optional
         The norm of the data. This can be optionally provided to avoid recalculating it.
+    projected_tensor : ndarray, optional
+        The projections of X into an aligned tensor for CP decomposition. This can be optionally
```
aarmey, line 121: If you can, rebasing onto TensorLy master will get rid of these changes that aren't associated with this PR.

tensorly/decomposition/_parafac2.py
```diff
@@ -364,7 +370,10 @@ def parafac_updates(X, w, f):
 
         if tol:
             rec_error = _parafac2_reconstruction_error(
-                tensor_slices, (weights, factors, projections), norm_tensor
+                tensor_slices,
+                (weights, factors, projections),
+                norm_tensor,
```
aarmey, line 375: I saw the black formatting test failed. Run `black tensorly` to reformat everything. I don't know where the actual formatting issue is—this is just a place to throw the comment.

tensorly/preprocessing.py
```diff
@@ -0,0 +1,132 @@
+from tensorly import backend as T
+
+from .parafac2_tensor import Parafac2Tensor
+from .tenalg.svd import svd_interface
+
+
+def svd_compress_tensor_slices(
+    tensor_slices, compression_threshold=0, svd="truncated_svd"
```
aarmey, line 8:
```suggestion
    tensor_slices, compression_threshold=0.0, svd="truncated_svd"
```

tensorly/preprocessing.py
```diff
@@ -0,0 +1,132 @@
...
+        to the scores.
+
+    References
+    ----------
+    .. [1] Helwig, N. E. (2017). Estimating latent trends in multivariate longitudinal
+           data via Parafac2 with functional and structural constraints. Biometrical
+           Journal, 59(4), 783-803. doi: 10.1002/bimj.201600045
+
+    .. [2] Jang JG, Kang U. Dpar2: Fast and scalable parafac2 decomposition for
+           irregular dense tensors. In2022 IEEE 38th International Conference on Data
```
aarmey, line 66: Check the formatting of the reference here. "In" became part of the conference name.

tensorly/preprocessing.py
```diff
@@ -0,0 +1,132 @@
...
+           Engineering (ICDE) 2022 May 9 (pp. 2454-2467). IEEE.
+
+    """
+    loading_matrices = [None for _ in tensor_slices]
+    score_matrices = [None for _ in tensor_slices]
```
aarmey, line 71: I like to use `X.append()` rather than pre-filling the list, but this is just a matter of style.

aarmey (changes requested):
This looks great!

In addition to the comments here, it looks like there are some test failures dependent on the backend. I don't think these are issues with your code here. PyTorch in particular is always frustrating due to the lower accuracy of its SVD. (Jean knows more about the details of this.) One possible easier solution for fixing the tests could be to force the arrays to be float64 (e.g., with PyTorch they default to float32). If this doesn't work, we will have to look into different sizing of the tensors, or adjusting the tolerance checks depending on the backend.

tensorly/decomposition/_parafac2.py
yngvem: Something weird happened when we rebased before submitting the PR so we got many other commits included as well (which we fixed with a new rebase just now). However we did make the changes in b12844ff49684741899247f5df7f81d44f9f864b (commit 1) here. They were small enough that we didn't bother to create a separate PR for them.

What do you think? Is it fine to include the first commit in this PR?

tensorly/preprocessing.py
yngvem: We arbitrarily landed at this because we use None as default value if we're not compressing the data. Also, this way we'd get an error downstream if we for some reason forgot to add a score matrix (we would get a type error instead of just a slightly smaller tensor).

aarmey: Right. Now understanding how `None` is handled later, this seems like a good strategy.

aarmey (approved):
Looks great!

tensorly/tests/test_preprocessing.py
```diff
@@ -0,0 +1,140 @@
...
+    # Second slice is not compressed
+    assert compressed_slices[1] is tensor_slices[1]
+    assert loadings[1] is None
+
+
+def test_svd_compression_compresses_uses_compression_threshold():
+    """If compression threshold is set, small singular values are truncated"""
+    rng = tl.check_random_state(1234)
+    slice = tl.tensor(rng.standard_normal((10, 6)), dtype=tl.float64)
+    slice[:, -1] = slice[:, -2] + 1e-10
```
aarmey, line 104: Take a look at the build errors for Jax and Tensorflow. This is an array assignment, so you want to use something like `slice = tl.index_update(slice, -1, slice[:, -2] + 1e-10)`.

aarmey (changes requested):
Missed one build error.

JeanKossaifi:
Very neat @MarieRoald @yngvem, and very clean PR as always!

Jax test should be an easy fix, as @aarmey suggested - it should also fix the tensorflow test but if it doesn't happy to just skip it.

MarieRoald:
Thanks! It passes all tests for all backends now 🙂

aarmey:
Excellent. Merging!

PR #532, opened 2023-10-19
Fully deprecate mxnet

aarmey:
The mxnet project has been [formally archived](https://whimsy.apache.org/board/minutes/MXNet.html). Therefore, I think it is reasonable to fully deprecate mxnet, before it stops working on various new Python and numpy versions.

I will also open a draft PR for removing mxnet that we can keep open until a new TensorLy version is issued.

Searching on Github, I only found mxnet being used in copies of the tutorials and one project that tests against all backends, so I think that the impact will be minimal.

### DIFF
--- a/tensorly/backend/jax_backend.py
+++ b/tensorly/backend/jax_backend.py
@@ -16,7 +16,6 @@ except ImportError as error:
     raise ImportError(message) from error
 
 import numpy
-import copy
 
 from .core import (
     Backend,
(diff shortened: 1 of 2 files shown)


tensorly/backend/jax_backend.py
```diff
@@ -98,14 +105,5 @@ def logsumexp(tensor, axis=0):
...
     JaxBackend.register_method(name, getattr(np.linalg, name))
 
-if Version(jax.__version__) >= Version("0.3.0"):
-
-    def index_update(tensor, indices, values):
-        return tensor.at[indices].set(values)
-
-    JaxBackend.register_method("index_update", index_update)
-else:
-    JaxBackend.register_method(name, getattr(jax.ops, name))
```
aarmey, line 108: I'm pretty sure the `else` statement just never worked here? It looks like `name` is undefined.

JeanKossaifi: Oh dear - you're right. That change was early days of Jax

PR #534, opened 2023-10-22
Simplify proximal operator code

aarmey:
Indexing operations are slow on the Jax and Tensorflow backends. This aims to reduce the places where indexing occurs. It should also make it possible to use the Tensorflow backend.


JeanKossaifi:
@aarmey I love it! This is so much cleaner!! Maybe we can just add a comment for someone reading that code later.

PR #536, opened 2023-11-07
Add max rank argument to svd_compress_tensor_slices

aarmey:
This adds a maximum rank argument to `svd_compress_tensor_slices()`. With very large matrices, calculating the full SVD can be very slow. This allows one to pair this with randomized SVD solving for better performance.


JeanKossaifi:
Thanks @aarmey, it looks good to me!

Issue #537, opened 2023-11-08
Tensor Conversion in TensorLy Does Not Preserve PyTorch Tensor dtype and device Attributes


braun-steven:
#### Describe the bug
The TensorLy library does not preserve the `dtype` and `device` attributes of a PyTorch tensor when using `tl.tensor()` to convert a PyTorch tensor into a TensorLy tensor. Instead, all output tensors are converted to `dtype=torch.float32` and `device=cpu`, regardless of the original tensor's attributes.

#### Steps or Code to Reproduce
Execute the following code snippet which demonstrates the incorrect conversion behavior:

```python
import tensorly as tl
import torch

tl.set_backend("pytorch")

def print_info(x: torch.Tensor):
    o = tl.tensor(x)
    print(f"In: dtype={str(x.dtype):14} device={str(x.device):10} -- Out: dtype={str(o.dtype):14} device={str(o.device):10}")

dtypes = [torch.int, torch.long, torch.float16, torch.float32, torch.float64]
devices = ["cpu", "cuda"]

for dtype in dtypes:
    for device in devices:
        x = torch.zeros(1, dtype=dtype, device=device)
        print_info(x)
```

#### Expected behavior
The dtype and device of the output tensor should match those of the input tensor. For instance, an input tensor with `dtype=torch.int32` and `device=cuda:0` should result in an output tensor with the same `dtype` and `device`.

#### Actual result

The output tensor always has `dtype=torch.float32` and is set to `device=cpu`, which is unexpected. Here is the output:

```
In: dtype=torch.int32    device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.int32    device=cuda:0     -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.int64    device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.int64    device=cuda:0     -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float16  device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float16  device=cuda:0     -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float32  device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float32  device=cuda:0     -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float64  device=cpu        -- Out: dtype=torch.float32  device=cpu
[...]

PR #538, opened 2023-11-08
Fix torch tensor creation dtype/device

braun-steven:
The previous implementation simply took the input data and forced it to be on the given device (`cpu` default) of the given dtype (`torch.float32` default).

This lead to `tl.tensor(torch.zeros(1, device="cuda:0"))` creating a `torch.float32` tensor on `cpu` while the expected behavior should be an `torch.int32` tensor on `cuda:0`.

Using the snippet provided in #537:

```python
import tensorly as tl
import torch

tl.set_backend("pytorch")

def print_info(x: torch.Tensor):
    o = tl.tensor(x)
    print(f"In: dtype={str(x.dtype):14} device={str(x.device):10} -- Out: dtype={str(o.dtype):14} device={str(o.device):10}")

dtypes = [torch.int, torch.long, torch.float16, torch.float32, torch.float64]
devices = ["cpu", "cuda"]

for dtype in dtypes:
    for device in devices:
        x = torch.zeros(1, dtype=dtype, device=device)
        print_info(x)
```

**Before**:
```
In: dtype=torch.int32    device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.int32    device=cuda:0     -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.int64    device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.int64    device=cuda:0     -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float16  device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float16  device=cuda:0     -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float32  device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float32  device=cuda:0     -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float64  device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float64  device=cuda:0     -- Out: dtype=torch.float32  device=cpu
```

**After**:
```
In: dtype=torch.int32    device=cpu        -- Out: dtype=torch.int32    device=cpu
In: dtype=torch.int32    device=cuda:0     -- Out: dtype=torch.int32    device=cuda:0
In: dtype=torch.int64    device=cpu        -- Out: dtype=torch.int64    device=cpu
In: dtype=torch.int64    device=cuda:0     -- Out: dtype=torch.int64    device=cuda:0
In: dtype=torch.float16  device=cpu        -- Out: dtype=torch.float16  device=cpu
In: dtype=torch.float16  device=cuda:0     -- Out: dtype=torch.float16  device=cuda:0
In: dtype=torch.float32  device=cpu        -- Out: dtype=torch.float32  device=cpu
In: dtype=torch.float32  device=cuda:0     -- Out: dtype=torch.float32  device=cuda:0
In: dtype=torch.float64  device=cpu        -- Out: dtype=torch.float64  device=cpu
In: dtype=torch.float64  device=cuda:0     -- Out: dtype=torch.float64  device=cuda:0
```

Resolves: #537

### DIFF
--- a/tensorly/backend/pytorch_backend.py
+++ b/tensorly/backend/pytorch_backend.py
@@ -35,21 +35,19 @@ class PyTorchBackend(Backend, backend_name="pytorch"):
 
     @staticmethod
     def _from_torch(data, dtype, device, requires_grad):
-        if dtype is None:
-            # Infer dtype from data
-            dtype = data.dtype
+        # Clone tensor as recommended by PyTorch
+        tensor = data.clone()
 
-        if requires_grad is None:
-            # Infer requires_grad from data
-            requires_grad = data.requires_grad
+        if dtype is not None:
+            tensor = tensor.type(dtype)
 
-        if device is None:
-            # Infer device from data
-            device = data.device
+        if requires_grad is not None:
+            tensor.requires_grad_(requires_grad)
 
-        return torch.tensor(
-            data.clone(), dtype=dtype, device=device, requires_grad=requires_grad
-        )
+        if device is not None:
+            tensor = tensor.to(device=device)
+
+        return tensor
 
     @staticmethod
     def _from_numpy(data, dtype, device, requires_grad):
--- a/tensorly/backend/pytorch_backend.py
+++ b/tensorly/backend/pytorch_backend.py
@@ -34,16 +34,88 @@ class PyTorchBackend(Backend, backend_name="pytorch"):
         }
 
     @staticmethod
-    def tensor(data, dtype=torch.float32, device="cpu", requires_grad=False):
-        if isinstance(data, torch.Tensor):
-            with torch.device(device):
-                return data.clone().detach().type(dtype).requires_grad_(requires_grad)
+    def _from_torch(data, dtype, device, requires_grad):
+        if dtype is None:
+            # Infer dtype from data
+            dtype = data.dtype
 
-        if isinstance(data, np.ndarray):
-            data = data.copy()
+        if requires_grad is None:
+            # Infer requires_grad from data
+            requires_grad = data.requires_grad
+
+        if device is None:
+            # Infer device from data
+            device = data.device
+
         return torch.tensor(
-            data, dtype=dtype, device=device, requires_grad=requires_grad
+            data.clone(), dtype=dtype, device=device, requires_grad=requires_grad
         )
+
+    @staticmethod
+    def _from_numpy(data, dtype, device, requires_grad):
+        # Use from_numpy provided by PyTorch
+        t = torch.from_numpy(data)
+
+        # Clone the tensor to avoid memory sharing between torch tensor and numpy array
+        t = t.clone()
+
+        # If specific dtype is given, set the dtype, else keep the dtype of the numpy array
+        if dtype is not None:
+            t = t.type(dtype)
+
+        t.requires_grad_(False if requires_grad is None else requires_grad)
+        return t.to(device=device)
+
+    @staticmethod
+    def _from_python(data, dtype, device, requires_grad):
+        t = torch.tensor(data)
+
+        # Set dtype if given, else keep the inferred dtype from data
+        if dtype is not None:
+            t = t.type(dtype)
+
+        # Set device if given, else keep the default device ("cpu")
+        if device is not None:
+            t = t.to(device=device)
+
+        # Set requires_grad if given, else keep the default requires_grad (False)
+        if requires_grad is not None:
+            t.requires_grad_(requires_grad)
+
+        return t
+
+    @staticmethod
+    def tensor(data, dtype=None, device=None, requires_grad=None):
+        """
+        Tensor constructor for the PyTorch backend.
+
+        Parameters
+        ----------
+        data : array-like
+            Data for the tensor.
+        dtype : torch.dtype, optional
+            Data type of the tensor. If None, the dtype is inferred from the data.
+        device : Union[str, torch.device], optional
+            Device on which the tensor is allocated. If None, the device is inferred from the data.
+        requires_grad : bool, optional.
+            If autograd should record operations on the returned tensor. If None, requires_grad is inferred from the
+            data.
+
+        """
+        if isinstance(data, torch.Tensor):
+            # Input is already a torch tensor
+            return PyTorchBackend._from_torch(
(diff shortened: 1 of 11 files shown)


braun-steven:
Btw., I wanted to write testcases for this but found that there are currently no backend specific test suits which I could have extended. According to the GitHub CI, all tests are run with all backends.

Do you have a suggestions where I should implement the relevant test cases for the torch backend?

braun-steven:
The following test seems to fail now. I haven't dug into the code, but I guess there is some implicit assumption about a tensor dtype being `float32` present (as this was the previous default no matter what was thrown into `tl.tensor(...)`) but the tensor is now `float64`: https://github.com/tensorly/tensorly/blob/1fe1fc7ef575fc43abd0deab1366529470f5c0a6/tensorly/decomposition/tests/test_constrained_parafac.py#L364

aarmey:
> Btw., I wanted to write testcases for this but found that there are currently no backend specific test suits which I could have extended. According to the GitHub CI, all tests are run with all backends.
> 
> Do you have a suggestions where I should implement the relevant test cases for the torch backend?

Thanks @braun-steven! This is an important fix.

So far, tests have only been constrained to specific backends by using `purest.mark.skipif()` to skip irrelevant backends. For example, [look here](https://github.com/tensorly/tensorly/blob/main/tensorly/contrib/decomposition/tests/test_mps_decomposition_cross.py#L11). Would this work in your case?

aarmey:
> The following test seems to fail now. I haven't dug into the code, but I guess there is some implicit assumption about a tensor dtype being `float32` present (as this was the previous default no matter what was thrown into `tl.tensor(...)`) but the tensor is now `float64`:
> 
> https://github.com/tensorly/tensorly/blob/1fe1fc7ef575fc43abd0deab1366529470f5c0a6/tensorly/decomposition/tests/test_constrained_parafac.py#L364

I'm happy to take a look at this if you can't find the issue. It is probably because the tensor context is not being passed to some of the variables used within the method. For example, I saw some calls to `tl.ones()` and `tl.zeros()` that don't have `**tl.context(tensor)`.

braun-steven:
> So far, tests have only been constrained to specific backends by using purest.mark.skipif() to skip irrelevant backends. For example, [look here](https://github.com/tensorly/tensorly/blob/main/tensorly/contrib/decomposition/tests/test_mps_decomposition_cross.py#L11). Would this work in your case?

Yes, that worked! Although I've removed the specification again, see below.

> It is probably because the tensor context is not being passed to some of the variables used within the method. For example, I saw some calls to tl.ones() and tl.zeros() that don't have **tl.context(tensor).

Found the culprits and added the missing tensor contexts.

I've now also added the proper tensor initialization for Tensorflow. I haven't touched Tensorflow in quite a while so I'm not 100% certain if I didn't miss anything -- maybe someone with more expertise in Tensorflow can double check this commit (https://github.com/tensorly/tensorly/commit/1a320eba9536c8a610b66642fa20b5067cf3cc04)? At least the tests seem to be passing with appropriate test fixes (https://github.com/tensorly/tensorly/commit/a3721d94ec6d2132b843013eae6a14fc7514d56b).

JeanKossaifi:
Thanks @braun-steven - when I first wrote this, pytorch was not creating a copy but it seems that it now does according to the doc so the simplest would actually to just remove the clone and just call `torch.tensor`.

braun-steven:
You're right -- I'm not sure why I wrapped the `tensor.clone()` call with an another `torch.tensor()` call. Fixed in e32701d.

JeanKossaifi:
What I meant is that we shouldn't need any specific code anymore, we should now be able to just call `torch.tensor` and it should correctly convert from numpy/list etc.

braun-steven:
Yes, we only need to catch the case where the input is already a tensor as recommended by PyTorch:

```
In [2]: torch.tensor(torch.rand(1))
<ipython-input-2-c2d0a58e0307>:1: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
  torch.tensor(torch.rand(1))
Out[2]: tensor([0.0470])
```

JeanKossaifi:
I think it's the dtype that you hard coded in the tests - I'm not sure we should add this. 
We can add an option to override the default global dtype and add a unit-test to make sure dtype is preserved correctly?

braun-steven:
That's not it (I think). Before the latest commit https://github.com/tensorly/tensorly/commit/3104f84f6ec070e0c1bd8136b49dc4a15e242643 all tests passed, so I'm confused as to why changing something in the torch backend would lead to tests failing for other backends.

> I'm not sure we should add this.

The issue with the test cases where I added a dtype is, that before my changes the following 

```python
x = tl.tensor([1, 2, 3])
y = tl.tensor([1.0, 2.0, 3.0])
z = x + z
```

would all be done in `tl.float32` since, at least for torch and TensorFlow, the `tl.tensor` method explicitly cast the input to `tl.float32` due to the default of `dtype=tl.float32` in the `tl.tensor()` method. 

With the changes I introduced (adopting the dtype from the input data if not specified), `x` would be of dtype `tl.int` and `y` dtype `tl.float32` which causes a mismatch of dtype for operations for those backends that don't explicitly cast to a matching dtype (e.g. TensorFlow: `tensorflow.python.framework.errors_impl.InvalidArgumentError: cannot compute AddV2 as input #1(zero-based) was expected to be a int64 tensor but is a double t
ensor [Op:AddV2] name:`)

So the solution to this is to specify the dtype in the test cases.

> We can add an option to override the default global dtype

Well, that would go against the idea of this PR since there was a default dtype previously (at least for the torch/TensorFlow backends) which lead to the weird behavior I reported in #537. Or did I misunderstand you?

I think a default dtype can make sense for situations where data is generated from an RNG but this should not affect the dtype of tensor construction with input data which already has a dtype.

JeanKossaifi:
Apologies this fell of my radar - thanks for the changes @braun-steven, it looks good to me! The tests logs are no longer available, let me see if I can relaunch them.

aarmey:
Merging! Thanks @braun-steven for resolving this.

PR #539, opened 2023-11-08
Avoid large concatenation within PARAFAC2 upon SVD initialization

aarmey:
Currently with SVD initialization, we concatenate all the matrices into one large tensor. This requires essentially copying the entire dataset. This PR adds a check for whether assembling the cross-product would be smaller, and then follows that route if so.


aarmey:
Good points! I'll make these changes.

JeanKossaifi:
Thank you @aarmey, apologies for the delay, this looks great! Merging.

PR #542, opened 2024-02-01
Reworking nonnegative algorithm and solvers

cohenjer:
First of all, I am sorry for the massive PR with so many small commits (to be squashed upon merge hopefully). I have been working on this for quite some time, and cutting this work into pieces seemed artificial and challenging.

Anyways I hope to make the nonnegative things in Tensorly a little cleaner and efficient with this!

# Improvements for NNLS
- Fixed Nonnegative Least Squares (NNLS) solvers (HALS, FISTA) that had bugs in them.
- Added ridge and sparse regularization support for all NNLS solvers and HALS algorithms.
- Replaced return_errors with callback for all nonnegative solvers (also Multiplicative Updates solvers). Added documentation which was missing for callback.
- Added the option to have unconstrained modes in ``non_negative_tucker_hals`` (this is now similar to ``non_negative_parafac_hals`` API)
- The error computation (math and code) were different in the various nonnegative decomposition algorithms. I chose to unify them, computing the loss function. This is not what is done in the CP function (we compute the square-root of the loss); this is open to discussion, but showing the loss is what makes the most sense to me. I updated the CP functions as well to have at least the same math for ``tensor_norm`` everywhere.
- Updated API for HALS function to allow some degree of control over the inner HALS nnls solver. But I also simplified the nnls HALS API so these new inputs should be self-explanatory.
- Reworked documentation (a lot) to add some missing arguments, and improve clarity.
- Updated the nonnegative decomposition examples to make them easier to follow and simpler. Also they now explain random tensor generations and parameter tuning in the algorithms.
- Of course, modified the tests where needed, and added new ones for the solvers/ (see below)

# Overall improvements
- Reverted MTTKRP back to the faster matrix-matrix product. I added the current memory-efficient version too, with some snippet of code in the documentation to show how to switch from one to the other using the backend system. This makes all CP algorithms significantly faster as we discussed here #442 
- Created a solvers/ folder in tensorly, to move all the optimization methods from tenalg/ to a dedicated place. solvers/ has four modules (``nnls.py``, ``admm.py``, ``proximal.py`` and ``penalizations.py``, this last one being mostly empty but I plan to populate it in a coming PR).

# Others
- Adding the maximum() function in the backend, useful for nonnegative projections.

# Pending
[...]

PR #549, opened 2024-04-03
Provide faster implementation of the MTTKRP

aarmey:
Copied from #542, with just some minor word smithing.

### DIFF
--- a/tensorly/metrics/tests/test_entropy.py
+++ b/tensorly/metrics/tests/test_entropy.py
@@ -1,3 +1,4 @@
+import pytest
 import tensorly as tl
 from ..entropy import vonneumann_entropy
 from ..entropy import tt_vonneumann_entropy, cp_vonneumann_entropy
@@ -131,6 +132,8 @@ def test_tt_vonneumann_entropy_mixed_state():
     assert_array_almost_equal(tl_vne, actual_vne, decimal=3)
 
 
+# Overfactoring causes a singular matrix error.
+@pytest.mark.xfail()
 def test_cp_vonneumann_entropy_mixed_state():
     """Test for cp_vonneumann_entropy on CP tensors.
     This test checks that the VNE of mixed states is calculated correctly.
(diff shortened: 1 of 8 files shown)


tensorly/metrics/tests/test_entropy.py
```diff
@@ -170,11 +170,9 @@ def test_cp_vonneumann_entropy_mixed_state():
         / 2.0
```
aarmey, line 170: I am confused about what `test_cp_vonneumann_entropy_mixed_state()` is trying to do. It seems like this test is correctly erroring with a singular matrix error, since the true rank of the tensor is 1.

cohenjer: I have no idea either... thought you were the one who wrote it !

cohenjer:
Thanks @aarmey ! This looks perfect to me. "Only" need to sort this error with the entropy test...

aarmey:
Ok, thanks @cohenjer. I'll mark it as broken and then proceed with this PR.

JeanKossaifi:
@aarmey the reason we had changed this is that the faster version is not sparse safe - I guess the sparse backend needs a fair amount of work anyway.

cohenjer:
@JeanKossaifi I was well aware of that that. That is why both methods are still in the package, and we allow the user to change the method easily with the backend manager. The procedure is in the documentation of both versions of mttkrp but I paste it here for further discussions: 
(here to switch from the khatri-rao version to the old version which I named ` unfolding_dot_khatri_rao_memory`)
```python
from tensorly.tenalg.core_tenalg.mttkrp import unfolding_dot_khatri_rao_memory
tl.tenalg.register_backend_method("unfolding_dot_khatri_rao", unfolding_dot_khatri_rao_memory)
tl.tenalg.use_dynamic_dispatch()
```
Using the backend manager, an expert user who requires the sparse backing (I guess this concerns very few people) can easily configure the mttkrp in his script to be the memory-efficient/sparse-safe/slow method. We can work on improving the sparse backend while enjoying fast dense computations by default. If you want we can swap back the default mttkrp, but in my opinion dense + fast should be the default one.

aarmey:
The sparse backend could make this switch for the user, too.

Do the sparse backend tests not run? It would be helpful to codify which code is expected to be sparse safe that way.

PR #550, opened 2024-04-10
moving solvers related code to solvers/* directory, improving fista and nnls_hals

cohenjer:
This is a copy of the part of #542 related to solvers.

What this PR solves:
----------------------------

- Currently, in Tensorly we have a large chunk of code in tenalg/proximal.py which barely relates to tensor algebra but rather numerical optimization. Other functions in Tensorly perform subtasks for numerical optimization such as nonnegative least squares and admm. 
- The nonnegative least squares solver nnls_hals has a bug related to sparsity (the formula is incorrect). Also, we should support ridge penalization since it is also quite simple. 
- The documentation of nnls_hals and fista was not friendly, and the code was not clear. I tried to improve upon these issues.

What this PR does:
--------------------------
(I left the old proximal file with the modified nnls_hals and fista solvers to show the changes, we can remove it before merging.)

- Move optimization-related code from various places to a dedicated solvers/ directory.  (only copy-pasting files, no changes)
- Correct the nnls hals solver when sparsity is required and clean up the syntax. 
- Adding support for ridge regularization in fista and in nnls_hals.
- Reworking documentation for fista and nnls_hals.
- Update the import paths where it is necessary.
-  Add a module  `solvers/penalizations.py` with a utility function to process input ridge/sparsity regularization (transform 1d input to list of correct length, avoid no regularization on only some modes which makes the factorization ill-posed).

Possible improvements:
----------------------------------

- move the SVD functions also in solvers/ ?

Note: If this PR is merged, I will proceed with the nonnegative algorithms improvements in #542.


cohenjer:
Not sure why the pytorch test for contained_parafac is returning an error, I did not change the contents of the code there... seems to be a type mismatch between floats and integers but I have no idea why it happens.

cohenjer:
@aarmey @JeanKossaifi @MarieRoald @yngvem if you can find some time to review this PR at your own pace that would be fantastic :)

tensorly/tenalg/proximal.py
```diff
@@ -879,11 +879,13 @@ def hals_nnls(
...
     V=None,
     n_iter_max=500,
-    tol=10e-8,
+    tol=1e-8,
     sparsity_coefficient=None,
-    normalize=False,
+    ridge_coefficient=None,
     nonzero_rows=False,
     exact=False,
+    epsilon=0,
```
aarmey, line 887:
```suggestion
    epsilon=0.0,
```

Many tools will infer this to be an int if 0 is the default.

tensorly/tenalg/proximal.py
```diff
@@ -1065,33 +1080,42 @@ def fista(
...
     if x is None:
         x = tl.zeros(tl.shape(UtM), **tl.context(UtM))
     if lr is None:
-        lr = 1 / (tl.truncated_svd(UtU)[1][0])
+        lr = 1 / (tl.truncated_svd(UtU)[1][0] + 2 * ridge_coef)
     # Parameters
-    momentum_old = tl.tensor(1.0)
-    norm_0 = 0.0
+    momentum_old = 1.0  # tl.tensor(1.0)
+    norm_0 = 0
```
aarmey, line 1118:
```suggestion
    norm_0 = 0.0
```

Same issue.

tensorly/solvers/proximal.py
```diff
@@ -0,0 +1,860 @@
...
+        else:
+            for i in range(len(constraints)):
+                constraints[i] = "non_negative"
+    if l1_reg:
+        if isinstance(l1_reg, dict):
+            modes = list(l1_reg)
+            for i in range(len(modes)):
+                if constraints[modes[i]] is not None:
+                    raise ValueError(
+                        "You selected two constraints for the same mode. Consider to check your input"
```
aarmey, line 83: Can the two constraints for the same mode check be done in a constraint-general way? In other words, does it have to be repeated within each if-else block? It might be more readable if it can be moved out.

tensorly/solvers/proximal.py
```diff
@@ -0,0 +1,860 @@
...
+    elif constraint == "normalized_sparsity":
+        return normalized_sparsity_prox(tensor, parameter)
+    elif constraint == "soft_sparsity":
+        return soft_sparsity_prox(tensor, parameter)
+    elif constraint == "smoothness":
+        return smoothness_prox(tensor, parameter)
+    elif constraint == "monotonicity":
+        return monotonicity_prox(tensor)
+    elif constraint == "hard_sparsity":
+        return hard_thresholding(tensor, parameter)
```
aarmey, line 413: Maybe an else with a runtime error here, to catch no matches?

aarmey (changes requested):
Looks great! Here are some minor suggestions.

aarmey:
@cohenjer the PyTorch test failure may have to do with a typing issue here, too.

tensorly/solvers/tests/test_proximal.py
```diff
@@ -1,13 +1,10 @@
 import numpy as np
+import tensorly as tl
```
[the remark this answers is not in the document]
cohenjer: This was my understanding as well, frankly, I just imitate whatever is the current style of the file I am updating.

tensorly/solvers/tests/test_admm.py
```diff
@@ -0,0 +1,26 @@
...
+
+# Author: Jean Kossaifi
+skip_tensorflow = pytest.mark.skipif(
+    (tl.get_backend() == "tensorflow"),
+    reason=f"Indexing with list not supported in TensorFlow",
+)
+
+
+def test_admm():
+    """Test for admm operator"""
```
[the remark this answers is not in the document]
cohenjer: Added some description !

cohenjer (commented):
Thanks a lot @aarmey; apart from the pytorch issue which I have yet to solve, I accounted for your comments.

aarmey:
@cohenjer by the way, `main` has a fix for the python_version >= "3.6" build error.

cohenjer:
There is a remaining issue with Tensorflow solve in the backend, when the inputs are tf.Variables (which is what we use for the factors in the CP class). Indeed, tl.solve calls fac.ndim, which is undefined for tf.Variables.
This was not watched before, because probably only the smoothness prox uses svd on these objects. I can fix this in a separate commit/PR by catching the resulting AttributeError, like this:
```python
    def solve(lhs, rhs):
        squeeze = False
        try:
            if rhs.ndim == 1:
                squeeze = [-1]
                rhs = tf.reshape(rhs, (-1, 1))
            res = tf.linalg.solve(lhs, rhs)
            if squeeze:
                res = tf.squeeze(res, squeeze)
            return res
        except AttributeError:
            # Variable inputs
            res = tf.linalg.solve(lhs, rhs)
            return res
```

aarmey:
Interesting. `solve()` should be using `tl.ndim(rhs)`, not `rhs.ndim`. It sounds like that should resolve the issue.

cohenjer:
> Interesting. `solve()` should be using `tl.ndim(rhs)`, not `rhs.ndim`. It sounds like that should resolve the issue.

Thanks for the tip (using `tnp` instead of `tl`). It works properly now.

tensorly/tenalg/proximal.py
```diff
@@ -913,7 +911,7 @@ def hals_nnls(
         Initialized V array
         By default, is initialized with one non-zero entry per column
         corresponding to the closest column of U of the corresponding column of M.
-    n_iter_max: Postivie integer
+    n_iter_max: Postive integer
```
aarmey, line 914:
```suggestion
    n_iter_max: Positive integer
```

aarmey:
This looks great! Thanks @cohenjer. Merging.

cohenjer:
Wait I need to remove the duplicate solvers code from proximal (I duplicated so we could compare) !

cohenjer:
I'll make the ablation of the duplicate code tomorrow in another quick PR

aarmey:
Thanks! Sorry I didn't catch that.

cohenjer:
@JeanKossaifi I reverted the PR so you can review #550 , will open a new PR cloning #550

PR #551, opened 2024-04-11
add normalization method to tucker_tensor class (similar to cp_tensor)

cohenjer:
A small PR to add normalization as a class method for `tucker_tensor`, which is something we already have for cp_tensor and is quite convenient in my humble opinion.

I can write a test but we already test tucker_normalize and this is just a wrapper.

This was also part of #542.


aarmey (approved):
I think it's fine to forgo a test here.

Issue #543, opened 2024-02-10
PARAFAC2 for missing data


cchatzis:
Hi! I am currently working with PARAFAC2 and I noticed there is no support for missing data (PARAFAC has the `mask` argument). Are there any plans for missing data support for PARAFAC2? If not, I would like to work on the classic EM approach that is present in [https://www.mathworks.com/matlabcentral/fileexchange/1089-parafac2](https://www.mathworks.com/matlabcentral/fileexchange/1089-parafac2).

Thank you for your time :)

MarieRoald:
Hi @cchatzis! PARAFAC2 with missing data would be very useful to have in TensorLy, so a PR for that sounds excellent  :)

PR #563, opened 2024-07-02
Parafac2 ALS EM

cchatzis:
Enable missing data imputation via the `mask` argument as done in CP, as brought up in #543 .

The changes are outlined as follows (all lines refer to `_parafac2.py`):

1. Added new argument `mask` to the `parafac2` function.
2. Initial slice-wise missing data imputation (lines 480-501)
3. Norm tensor is adjusted to include only the non-missing entries in the presence of missing data (lines 510-523)
4. Update imputation after a full factor update circle (lines 601-611)
5. Adjusted `_parafac2_reconstruction_error` to include the `mask` argument in case `norm_matrices` is not given (lines 260, 297-306)
6. Adjusted `_BroThesisLineSearch` to include `mask`, required when computing the reconstruction error.
7. Added the `mask` docstring whenever required.


MarieRoald:
Hi @cchatzis Great! I have my hands full with a paper deadline right now, but I can take a look when my plate clears a little 🙂

JeanKossaifi:
Thanks @cchatzis for opening the PR, this will be great to have!

For the failing tests, you can also try to run the tests locally using `pytest` to check things work as expected and quick debugging.

Regarding the masking, it would be more efficient to do the computation in a vectorized way, e.g. you can first multiply the slice by mask array of 0 and 1 and then take the mean rather than loop over individual elements.

Thanks @MarieRoald, will be great to get your feedback too!

tensorly/decomposition/_parafac2.py
```diff
@@ -544,6 +602,18 @@ def parafac_updates(X, w, f):
...
 
+        # Update imputations
+        if mask is not None:
+
+            reconstructed_tensor = parafac2_to_tensor((weights, factors, projections))
+
+            for idx in indices_missing:
+                tl.index_update(
+                    tensor_slices,
+                    tl.index[idx[0], idx[1], idx[2]],
```
cohenjer, line 613: This will not work for order>3 tensors; there should be a better way do this?

In other functions we write something like tensor = tensor*mask + reconstructed_tensor * (1-mask)
which I think would also apply here.

cohenjer (commented):
Thanks @cchatzis for this PR !
Looks correct to me, I have few minor comments.

cchatzis:
Thanks again everyone for taking the time to look and review this.

In the latest commits, I updated the code to include your latest feedback. Relevant tests are passed now - I will create a small test for the _update_imputed function in the near future.

tensorly/decomposition/_parafac2.py
```diff
@@ -458,15 +481,50 @@ def parafac2(
     non-negative, then :math:`B` will be non-negative, but not the orthogonal `P_i` matrices.
     Consequently, the `B_i` matrices are unlikely to be non-negative.
     """
+    if mask is not None:
+        # Fill in missing values with the mean of the mode-2 slices of observed
+        indices_missing = []
+        for slice_no, (slice, slice_mask) in enumerate(zip(tensor_slices, mask)):
+            slice_mean = tl.mean(
```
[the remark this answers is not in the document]
cchatzis: I added a specific function called `_update_imputed` in 22343365a71051260aab7ff9d447809ae32239ee, kindly let me know if you think this solves this issue.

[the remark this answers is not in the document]
cohenjer: Looks good to me, thanks :)

cohenjer:
Thanks @cchatzis for the updates, this version looks good to me.

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -472,3 +473,47 @@ def test_parafac2_to_tensor():
...
+    assert_allclose(imputed_tensor, tensor)
+
+    # test mode='mode-2' imputation
+
+    tensor = np.array(tensor)
+
+    tensor[mask == 0] == np.nan
+
+    for slice in tensor:
+        slice[slice == 0] = np.nanmean(slice)
```
cchatzis: Is casting to a numpy array acceptable for this test?

tensorly/decomposition/_parafac2.py
```diff
@@ -277,6 +342,9 @@ def _parafac2_reconstruction_error(
     projected_tensor : ndarray, optional
         The projections of X into an aligned tensor for CP decomposition. This can be optionally
         provided to avoid recalculating it.
+    mask : ndarray, optional
+        array of booleans with the same shape as tensor should be 0 where the values are
+        missing and 1 everywhere else.
```
yngvem, line 347:
```suggestion
        An array with the same shape as the tensor. It should be 0 where there are
        missing values and 1 everywhere else.
```
This text is a bit unclear. Maybe something like this instead? :)

tensorly/decomposition/_parafac2.py
```diff
@@ -553,6 +672,7 @@ def parafac_updates(X, w, f):
                 (weights, factors, projections),
                 norm_tensor,
                 projected_tensor,
+                mask,
```
MarieRoald, line 675: It's not obvious to me that this is correct. The code is reusing the projected tensor (`P_i @ X_i`) computed on line 641. However, on line 659, the missing values of `X_i` are imputed. I think that means that you need to "re-project" the tensors before you compute the loss if `mask is not None`?

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -472,3 +473,47 @@ def test_parafac2_to_tensor():
...
+
+    rng = tl.check_random_state(1234)
+
+    random_parafac2_tensor = random_parafac2(
+        shapes=[(15, 30) for _ in range(25)],
+        rank=3,
+        random_state=rng,
+    )
+
+    # test mode='factors' imputation
```
MarieRoald, line 488: I think it would be easier to understand if this comment (and the comment on line 503) was a complete sentence explaining what the test is checking

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -472,3 +473,47 @@ def test_parafac2_to_tensor():
         Bi = T.dot(projections[i], factors[1])
         manual_tensor = T.einsum("r,jr,kr", factors[0][i], Bi, factors[2])
         assert_(tl.max(tl.abs(constructed_tensor[i, :, :] - manual_tensor)) < 1e-6)
+
+
+def test_update_imputed():
```
MarieRoald, line 478: The unit test is nice, but it would also be good with an integration test where we mask out some values and check that we get the correct components by running PARAFAC2 ALS EM

MarieRoald (changes requested):
Great work! Only a couple of minor comments on my end :)

tensorly/decomposition/_parafac2.py
```diff
@@ -17,8 +18,57 @@
...
+
+    Parameters
+    ----------
+    tensor_slices : Iterable of ndarray
+    mask : ndarray
+        array of booleans with the same shape as tensor should be 0 where the values are
+        missing and 1 everywhere else.
+    decomposition : Parafac2Tensor, optional
+    method : string
+        One of 'mode-2' or 'factors'. 'mode-2' updated imputed values according to mean of each mode-2 slice. If 'factors' is chosen, update imputationa according to reconstruction given from 'decomposition'.
```
yngvem, line 33: I think it would be nice if this docstring also stated that "mode-2" was only used for initialisation and not used during model fitting.

tensorly/decomposition/_parafac2.py
```diff
@@ -17,8 +18,57 @@
 #          Yngve Mardal Moe
 
 
+def _update_imputed(tensor_slices, mask, decomposition=None, method="mode-2"):
```
yngvem, line 21: I'm not sure if we need these default values. In my experience, default values for arguments can make it very easy to "trip up" and forget to pass a value -- especially when we're refactoring. We can be explicit about `decomposition` being `None` when that's needed :)
```suggestion
def _update_imputed(tensor_slices, mask, decomposition, method):
```

tensorly/decomposition/_parafac2.py
```diff
@@ -458,15 +536,46 @@ def parafac2(
...
     )
     factors = list(factors)
 
+    # Initial missing imputation
+    if mask is not None:
+
+        if init == "random" or init == "svd":
+
+            tensor_slices = _update_imputed(
+                tensor_slices=tensor_slices, mask=mask, method="mode-2"
```
yngvem, line 551: We can explicitly pass `decomposition=None` here
```suggestion
                tensor_slices=tensor_slices, mask=mask, decomposition=None, method="mode-2"
```

yngvem (commented):
Just adding a couple of minor suggestions here :)

tensorly/decomposition/_parafac2.py
cohenjer: Actually @MarieRoald made me realize that the P2 error computation function with mask is probably wrong. 
For now it works as follows:
1. compute the l2 norm of the tensor, at only non-masked values. This works fine.
2. Use the precomputed quantities to compute inner products <T, decomp> and the l2 norm of the model.
This second step is bugged as @MarieRoald mentioned because the projection quantities are outdated, but furthermore, the loss should account for the mask in step 2 as well (the data tensor and the model are compared only at 1-mask values).

The rationale behind steps 1 and 2 is the identity $\|a - b\|_2^2 = \|a\|_2^2 + \|b\|_2^2 - 2<a,b> $, but in the presence of missing data we are computing $\|w*(a-b)\|_2^2\| = \|w*a\|_2^2 + \|w*b\|_2^2 - 2<w*a,w*b>$. I think right now the code computes $\|w*a\|_2^2 + \|b\|_2^2 - 2<w*a_old, b>$ with outdated $a_old$ in the scalar product.

For comparison, in the CP error calculation function, we simply compute the loss naively in the presence of missing data. Because we need to recompute quantities such as the projections before computing the error, it might be similarly effective to just compute the batch cost?

tensorly/decomposition/_parafac2.py
cchatzis: Thank you for looking into this. This was addressed in fe0acb5fb91b54d00926258756e732187223b33b.

tensorly/decomposition/tests/test_parafac2.py
cchatzis: Thanks for the input @MarieRoald! This was addressed in bcb0fe51b06f07d48ab37217f17f7d148495796e.

tensorly/decomposition/_parafac2.py
cchatzis: Thanks! This was addressed in a20b562526de762ff7e1f6296d92136e8348dd5b.

tensorly/decomposition/_parafac2.py
cchatzis: Also addressed in a20b562526de762ff7e1f6296d92136e8348dd5b.

tensorly/decomposition/_parafac2.py
cchatzis: Also addressed in a20b562526de762ff7e1f6296d92136e8348dd5b.

tensorly/decomposition/_parafac2.py
cchatzis: Thank you both for the detailed feedback.
@MarieRoald I think you are right as well, I addressed this on b0edd5d6465a6219a7182c1254a33092bf395d65.
@cohenjer I will experiment with both implementations in terms of efficiently and come back at you.

cchatzis: @cohenjer I created this small test to investigate the performance, while the output plot is shown below. **Approach (1)** is the existing one that includes recomputing the projections while **approach (2)** "naively" computes the error. There was nothing printed from the `isclose` clause. I guess this means we should  use the latter in the case of missing data? Kindly let me know what you think.
```
def create_random_data(shape, rank):
    I, J, K = shape  # noqa: E741
    pf2 = tl.random.random_parafac2(
        [(J, K) for i in range(I)], rank=rank
    )
    return pf2

import time
import math

tolerance = 1e-8

times1 = []
times2 = []

for i in range(1000000):

    decomposition = create_random_data((3, 80, 25), 3)
    tensor_slices = tl.tensor([np.random.rand(80, 25) for i in range(3)])
    mask = tl.tensor([np.random.randint(0, 2, (80, 25)) for i in range(3)])

    start1 = time.time()

    error1 = _parafac2_reconstruction_error_v1(tensor_slices=tensor_slices, decomposition=decomposition, mask=mask)

    end1 = time.time()

    times1.append(end1 - start1)

    start2 = time.time()

    error2 = _parafac2_reconstruction_error_v2(tensor_slices=tensor_slices, decomposition=decomposition, mask=mask)

    end2 = time.time()

    times2.append(end2 - start2)

    if not math.isclose(error1, error2, abs_tol=tolerance):
        print(f'Error {error1} != {error2} within tolerance {tolerance}')

# do a boxplot of each times1 and times2
plt.boxplot([times1, times2])
plt.yscale('log')
plt.show()
```
image

For completeness, here is what **approach (2)** looks like:
```
def _parafac2_reconstruction_error_v2(
    tensor_slices, decomposition, norm_matrices=None, projected_tensor=None, mask=None
):
    reconstructed_tensor = parafac2_to_tensor(decomposition)
    total_error = 0
    for i, (slice, slice_mask) in enumerate(zip(tensor_slices, mask)):
        total_error += tl.norm(slice_mask * slice - slice_mask * reconstructed_tensor[i])**2
    return tl.sqrt(total_error)
```

cohenjer: Thanks for the precise testing @cchatzis. I assume version 1 is the one you had implemented already in the PR? I think your experiments prove that v1 and the naive error v2 are not the same, and since v2 is obviously the correct loss, this confirms that (1) was not correct. 

Even if it is slower, you should use the v2 error computation in the PR.

cchatzis: @cohenjer thank you for your input! v1 refers to the "corrected" version of the one currently in the PR (its not pushed yet). You and @MarieRoald correctly mentioned the need for re-computations and this refers to this "adjusted" version that includes these.

Should I then discard all pre-computations implemented and change the current code so it only computes the error like this (naively) or have an if clause that checks that and if there is no `mask` given utilize the precomputations?

cohenjer: You cannot use pre-computations because of the mask, as I explained earlier. If there is not mask then the precomputations, corrected (as in v1) should be fine.

tensorly/decomposition/tests/test_parafac2.py
cchatzis: Added this test on 174eca66f9bcb0daca88df7862e48e805817214e. Thanks for the suggestion! :)

tensorly/decomposition/_parafac2.py
cchatzis: @cohenjer addressed this in the latest coments, let me know what you think :)

tensorly/decomposition/_parafac2.py
```diff
@@ -48,21 +48,26 @@ def _update_imputed(tensor_slices, mask, decomposition, method):
...
                 tl.prod(slice.shape) - len(tl.where(slice_mask == 0)[0])
             )
+
             for i in range(slice.shape[0]):
                 for j in range(slice.shape[1]):
                     if slice_mask[i, j] == 0:
                         tl.index_update(
-                            tensor_slices, tl.index[slice_no, i, j], slice_mean
+                            tensor_slices[slice_no], tl.index[i, j], slice_mean
                         )
```
yngvem: I believe we could use [`tl.where`](https://tensorly.org/stable/modules/generated/tensorly.where.html#tensorly.where) here:

```suggestion
            tensor_slices[slice_no] = tl.where(slice_mask, tensor_slices[slice_no], tensor_slices[slice_no], slice_mean)
```

I'm not entirely sure though (I always need to double check how `where` works), but I think it should work :)

tensorly/decomposition/_parafac2.py
```diff
@@ -48,21 +48,26 @@ def _update_imputed(tensor_slices, mask, decomposition, method):
...
-            + (tl.ones(shape=mask.shape) - mask) * reconstructed_tensor
-        )
+        for slice_no, (slice, rec_slice, slice_mask) in enumerate(
+            zip(tensor_slices, reconstructed_slices, mask)
+        ):
+
+            tensor_slices[slice_no] = (
+                slice_mask * slice
+                + (tl.ones(shape=slice_mask.shape) - slice_mask) * rec_slice
+            )
```
yngvem: If we end up using `tl.where` above, then we should probably also use it here for consistency :)

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -518,3 +518,57 @@ def test_update_imputed():
...
+    )
+
+    random_parafac2_tensor.factors[0] = random_parafac2_tensor.factors[0] + 0.01
+
+    slices = parafac2_to_slices(random_parafac2_tensor)
+
+    # Form the full data and a mask with ~10% missing values
+
+    slice_mask = [rng.binomial(1, 0.9, size=T.shape(slice)) for slice in slices]
+
```
yngvem, line 543: It would be nice to apply the slice mask to the data as well so the test would fail if we completely ignored the mask

```suggestion

    # Apply the mask, setting missing values to zero
    slices = [tl.where(slice_mask, slice, 0) for slice in slices]

```

tensorly/decomposition/_parafac2.py
```diff
@@ -45,8 +45,9 @@ def _update_imputed(tensor_slices, mask, decomposition, method):
     if method == "mode-2":
 
         for slice_no, (slice, slice_mask) in enumerate(zip(tensor_slices, mask)):
+
             slice_mean = tl.sum(slice * slice_mask) / (
-                tl.prod(slice.shape) - len(tl.where(slice_mask == 0)[0])
+                int(T.prod(T.tensor(slice.shape))) - len(T.where(slice_mask == 0)[0]) # tensorflow required this casting to int
             )
```
yngvem, line 51: Since the mask is 1 for all observed values and 0 for all non-observed values, we could probably simplify this a bit. Maybe this will work as a shorter alternative?
```suggestion
            slice_mean = tl.sum(slice * slice_mask) / int(T.sum(slice_mask))  # Cast to int due to TensorFlow
```

tensorly/decomposition/_parafac2.py
yngvem: Really strong stuff @cchatzis, great job testing the two methods 💪

yngvem (changes requested):
Overall, this looks very good! Just a couple of more changes and I think it's good to go :)

tensorly/decomposition/_parafac2.py
cchatzis: Absolutely true. Thanks a lot for bringing this to my attention! I addressed that in a424e3205d5c239e6825c4cdb4a23b63f61f3751.

tensorly/decomposition/_parafac2.py
cchatzis: Done! :)

tensorly/decomposition/tests/test_parafac2.py
cchatzis: Added that on ce46c43d022536ce871e4f6dde02e4a823da79cb. Thanks a lot for the nice feedback!

tensorly/decomposition/_parafac2.py
cchatzis: Addressed that one as well :) Good eye!

tensorly/decomposition/_parafac2.py
cohenjer: looks good to me as well !

JeanKossaifi: Great comparisons and investigation, thank you!

cchatzis:
Hi everyone and thank you very much for your kind feedback and help!

Maybe I am missing something but I cannot seem to reproduce the issue with the paddle backend because when i try to set it I set it for pytest (i.e. : `TENSORLY_BACKEND=paddle pytest  ...`) I get:
```
UserWarning: TENSORLY_BACKEND should be one of 'numpy''pytorch''tensorflow''cupy''jax', got paddle. Defaulting to numpy'
```
I am sure I am missing something trivial here, any suggestion/help would be welcome :)

cchatzis:
Hi, are there any updates on this?

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -488,36 +488,35 @@ def test_update_imputed():
...
 
     # Check that _update_imputed works correctly when imputing values according to nanmean of mode-2 slices
 
-    tensor = np.asarray(tensor)
-
-    tensor[mask == 0] == np.nan
-
-    for slice in tensor:
-        slice[slice == 0] = np.nanmean(slice)
+    slices = np.array(slices)
```
yngvem, line 508: Hmmm, maybe we want to use `tl.stack` here instead? Just to be sure that we're using TensorLy tensors?

```suggestion
    slices = tl.stack(slices, axis=0)
```

yngvem (changes requested):
Almost at the finish line now! I just have a small comment :)

Regarding the paddle trouble, I think that will sort itself out if you merge main into your feature branch (run `git merge main` from your feature branch)

tensorly/decomposition/tests/test_parafac2.py
cchatzis: Simply omitting this line seems to do the trick regardless of the backend ¯\_(ツ)_/¯.

yngvem: Fabolous!

JeanKossaifi:
@cchatzis thanks for the changes, feel free to mark the changes requested as done when it's ready!

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -13,6 +13,7 @@
     parafac2,
```
aarmey, line 13: In changes on main, we changed the imports within the test to importing tensorly as a whole, rather than the backend as T. There are a few minor adjustments below as a result.

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -451,3 +452,117 @@ def test_parafac2_to_tensor():
...
+        rank=3,
+        random_state=rng,
+    )
+
+    # Check that _update_imputed works correctly when imputing values according to the reconstructed tensor,
+    # i.e. mode='factors'
+
+    slices = parafac2_to_slices(random_parafac2_tensor)
+    slices_masks = [
+        tl.tensor(rng.binomial(1, 0.25, size=T.shape(slice)), dtype=tl.float64)
```
aarmey, line 472:
```suggestion
        tl.tensor(rng.binomial(1, 0.25, size=tl.shape(slice)), dtype=tl.float64)
```

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -451,3 +452,117 @@ def test_parafac2_to_tensor():
...
+
+    C_corr = congruence_coefficient(random_parafac2_tensor.factors[2], rec.factors[2])[
+        0
+    ]
+    assert_(C_corr > 0.98)
+
+    for i, (true_proj, rec_proj) in enumerate(
+        zip(random_parafac2_tensor.projections, rec.projections)
+    ):
+        true_Bi = T.dot(true_proj, random_parafac2_tensor.factors[1]) * A_sign[i]
```
aarmey, line 565:
```suggestion
        true_Bi = tl.dot(true_proj, random_parafac2_tensor.factors[1]) * A_sign[i]
```

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -451,3 +452,117 @@ def test_parafac2_to_tensor():
...
+        rec_Bi = T.dot(rec_proj, rec.factors[1]) * rec_A_sign[i]
```
aarmey, line 566:
```suggestion
        rec_Bi = tl.dot(rec_proj, rec.factors[1]) * rec_A_sign[i]
```

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -451,3 +452,117 @@ def test_parafac2_to_tensor():
...
+        rank,
+        random_state=rng,
+        n_iter_max=1000,
+        linesearch=linesearch,
+        mask=slices_masks,
+    )
+
+    # assert FMS > 0.98
+    A_sign = T.sign(random_parafac2_tensor.factors[0])
+    rec_A_sign = T.sign(rec.factors[0])
```
aarmey, line 551:
```suggestion
    rec_A_sign = tl.sign(rec.factors[0])
```

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -451,3 +452,117 @@ def test_parafac2_to_tensor():
...
+        slices,
+        rank,
+        random_state=rng,
+        n_iter_max=1000,
+        linesearch=linesearch,
+        mask=slices_masks,
+    )
+
+    # assert FMS > 0.98
+    A_sign = T.sign(random_parafac2_tensor.factors[0])
```
aarmey, line 550:
```suggestion
    A_sign = tl.sign(random_parafac2_tensor.factors[0])
```

tensorly/decomposition/tests/test_parafac2.py
```diff
@@ -451,3 +452,117 @@ def test_parafac2_to_tensor():
...
+    )
+
+    random_parafac2_tensor.factors[0] = random_parafac2_tensor.factors[0] + 0.01
+
+    slices = parafac2_to_slices(random_parafac2_tensor)
+
+    # Form the full data and a mask with ~10% missing values
+
+    slices_masks = [
+        tl.tensor(rng.binomial(1, 0.9, size=T.shape(slice)), dtype=tl.float64)
```
aarmey, line 527:
```suggestion
        tl.tensor(rng.binomial(1, 0.9, size=tl.shape(slice)), dtype=tl.float64)
```

aarmey (changes requested):
Thanks for putting this together, @cchatzis! As this has been sitting, there are a few minor adjustments to make it consistent with code changes on main. Please also run the `black` formatter so that test passes. Everything else looks good to me, so I can merge it if you ping me after these edits.

cchatzis:
Thank you for looking into this @aarmey! I added your suggestions and also made a small change for the paddle backend. Something in the lines of `tl.nanmean` is not available to compare the results with - I used the numpy version in the `test_impute_results()` test. Would that be acceptable?

cchatzis:
The test results with black don't seem to make sense, two-line comments are marked as problematic:
```
--- /home/runner/work/tensorly/tensorly/tensorly/tenalg/__init__.py	2025-05-05 19:11:01.364542+00:00
+++ /home/runner/work/tensorly/tensorly/tensorly/tenalg/__init__.py	2025-05-05 19:11:08.018732+00:00
@@ -1,7 +1,7 @@
-""" 
would reformat /home/runner/work/tensorly/tensorly/tensorly/tenalg/__init__.py
-The :mod:`tensorly.tenalg` module contains utilities for Tensor Algebra 
+"""
+The :mod:`tensorly.tenalg` module contains utilities for Tensor Algebra
 operations such as khatri-rao or kronecker product, n-mode product, etc.
 """
```
[...]
```
  from .... import backend as tl
 
 import pytest
 import numpy as np
--- /home/runner/work/tensorly/tensorly/tensorly/decomposition/__init__.py	2025-05-05 19:11:01.360542+00:00
+++ /home/runner/work/tensorly/tensorly/tensorly/decomposition/__init__.py	2025-05-05 19:11:07.181622+00:00
@@ -1,8 +1,8 @@
 """
 The :mod:`tensorly.decomposition` module includes utilities for performing
-tensor decomposition such as CANDECOMP-PARAFAC and Tucker.                                                                                               
+tensor decomposition such as CANDECOMP-PARAFAC and Tucker.
 """
```

aarmey:
We can ignore those errors, since they don't involve the code here. Also, yes, use of `np.nanmean` is reasonable within the tests. This all looks in order so I'll merge. Thanks again for implementing this!

Issue #481, opened 2023-01-20
Unified and consistent error/convergence tracking 


aarmey:
Related to #470, I have been wondering if we should make the convergence handling consistent and reusable across methods. Currently, methods handle convergence checks in slightly different ways—some have an absolute threshold (PARAFAC2), or a relative threshold set according to the absolute difference in error (CP) or fractional difference in error (PARAFAC2). Finally, there is repeated code for storing a history of the error, checking these thresholds, printing the error at each iteration, etc.

One possibility would be to have an error tracker class, that is instantiated by any method. This class takes the tolerances on instantiation and handles storing a history of the errors. It has one method that is called on each iteration of a method, and returns a boolean as to whether the method should continue. Printing the per-iteration error is handled by this method (with a verbose flag).

My impression is that this would clean up a lot of code and force methods to have consistent convergence criteria. Note that the method itself would still have to handle calculating the absolute error on each iteration (so efficiencies like reuse of MTTKRP in CP can still occur).

JeanKossaifi:
Thanks @aarmey. I am a little conflicted by such refactoring. Overall I like to have each algorithm being clearly readable from source and not spread out important bits in different files. Convergence can be pretty method specific as you mentioned, so I think it makes sense to do it in each file. Plus, the code to store error, and check threshold is pretty trivial so I don't think there is a need to centralize that. I often use scikit-learn as reference, they manage really well the balance features/complexity despite the growing codebase.

I almost find the callback option already overkill in some ways but if we really want to refactor I guess we could use a callback to let users store the error history and not do that ourselves in the also.

In any case, we could probably use some harmonization of the error metrics across the codebase as you're saying.

aarmey:
Indeed, I was worried this might be too clever and impact readability. Agreed that we can do some work to unify the convergence options, but leave it at that.

JeanKossaifi:
Wanted to check in - what do you think @aarmey ?

aarmey:
I would still like to go back through methods and make the tolerance checks defined in a consistent way. This shouldn't be too difficult, so I'll hopefully have a PR in a couple weeks.

aarmey:
@JeanKossaifi @cohenjer what is the purpose of `cvg_criterion`? If the error is strictly decreasing, shouldn't both options lead to the same behavior?

https://github.com/tensorly/tensorly/blob/59c1126251b248d85d5333477d91c50eae3b2864/tensorly/decomposition/_cp.py#L285

[...]

PR #564, opened 2024-07-23
Cleanup PARAFAC2 tolerances

aarmey:
Part of #481, this defines the tolerance in PARAFAC2 identically to the PARAFAC methods.

I still need to review the docstrings and explain a few more points justifying this change here.


aarmey:
This is ready to review.

tensorly/decomposition/_parafac2.py
```diff
@@ -573,11 +560,7 @@ def parafac_updates(X, w, f):
                         f"PARAFAC2 reconstruction error={rec_errors[-1]}, variation={rec_errors[-2] - rec_errors[-1]}."
                     )
 
-                if (
-                    tl.abs(rec_errors[-2] ** 2 - rec_errors[-1] ** 2)
-                    < (tol * rec_errors[-2] ** 2)
-                    or rec_errors[-1] ** 2 < absolute_tol
-                ):
+                if rec_errors[-2] - rec_errors[-1] < tol:
```
JeanKossaifi, line 563: Should there be an absolute value here?

tensorly/tenalg/svd.py
```diff
@@ -357,6 +358,7 @@ def randomized_svd(
 
 
 SVD_FUNS = ["truncated_svd", "symeig_svd", "randomized_svd"]
+SVD_TYPES = Literal["truncated_svd", "symeig_svd", "randomized_svd"]
```
JeanKossaifi, line 361: Should it just be Literal[f for f in SVD_FUNS] ?

JeanKossaifi (commented):
Thanks @aarmey sorry just saw you marked it as ready for review.

It looks good, just left some comments - I somehow find type hints more difficult to parse than just signature + doc, especially when it gets nested like Optional[Union[..]...] but I guess it can be nice to have for type check.

tensorly/tenalg/svd.py
yngvem: Unfortunately, `Literal` [doesn't allow arbitrary values](https://typing.readthedocs.io/en/latest/spec/literal.html#illegal-parameters-for-literal-at-type-check-time). A hacky workaround is

```
SVD_TYPES = Literal["truncated_svd", "symeig_svd", "randomized_svd"]
SVD_FUNS = list(SVD_TYPES.__args__)
```

but I'm not sure if that's any better than just duplicating the string, which makes it absolutely clear for the reader what's happening with the code.

yngvem:
Great job @aarmey! I haven't done a super thorough review, but it looks great at first glance!

I have one question. I see you're using `Union`- and `Optional`-types instead of the new `|`- and ` | None`-syntax. I assume this is to support Python 3.9, which is [EOL next year](https://devguide.python.org/versions/)?

A workaround if we want to use the new syntax is adding a `from __future__ import annotations`, which turns all type annotations into strings. The upside is that it makes type annotations forward compatible, and a downside is that arbitrary syntax is allowed so typos might not be found. However IDEs are usually good at spotting typos, so I think it's worth it for the increased readibility.

Also, the previous default tolerance of `1e-8` was selected based on @MarieRoald's experience with PARAFAC2, so if we change the tolerance function, it might be worthwhile to also check that the default value is sensible (and works for non-simulated data also), but you might have done that already?

aarmey:
> Thanks @aarmey sorry just saw you marked it as ready for review.
> 
> It looks good, just left some comments - I somehow find type hints more difficult to parse than just signature + doc, especially when it gets nested like Optional[Union[..]...] but I guess it can be nice to have for type check.

I agree that the type annotations can become unwieldy, @JeanKossaifi. I have found them to still be helpful, on balance, because they describe the argument types in a very exact manner:

1. Occasionally, the documentation is a little vague because arguments such as `nn_mode` are quite complicated. A more advanced user can inspect these to know exactly what can be passed in.
2. These seem to help with autocomplete tools, probably because they are machine-readable.
3. We have a complicated codebase, and I've found automated checkers such as ruff/pyright to be helpful. For instance, quickly skimming the results, `tl.linalg.higher_order_moment()` currently does not work because `batched_outer()` only takes one argument, not two. It's easy to overdo it with automated tools, but I wonder if these plus some type annotation could help to avoid a separate set of errors compared to our tests. (I have little experience with large codebases, so I am genuinely wondering what people think about this.)

aarmey:
Thanks, @JeanKossaifi and @yngvem! I like this plan, so I'll merge this and then follow up with a PR removing Python 3.9 support and simplifying the type annotations.

PR #565, opened 2024-07-24
[Paddle] Add paddle backend

HydrogenSulfate:
Add paddle backend in tensorly. For unitest, please run script below:

``` sh
# other cuda version see: https://www.paddlepaddle.org.cn/en/install/quick?docurl=/documentation/docs/en/develop/install/pip/linux-pip_en.html
python -m pip install --pre paddlepaddle-gpu -i https://www.paddlepaddle.org.cn/packages/nightly/cu118/

cd ./tensorly
TENSORLY_BACKEND='paddle' pytest -v tensorly
```

the output will be 

``` log
...
...
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================================================================================== 425 passed, 4 skipped, 1 xfailed, 17 warnings in 275.20s (0:04:35===========================================================================================
```

This is mainly contain two part of changes, adaption code for paddle([5bf4ccb](https://github.com/tensorly/tensorly/pull/565/commits/5bf4ccbdabdf9e10b91cfaed38987131f1d1f95c)), and changes by `black ./`(as descripted in RAEDME.md#Contributing code, [ae7249f](https://github.com/tensorly/tensorly/pull/565/commits/ae7249f31c53a1223ddbae7144d89b690a8226d6))


HydrogenSulfate:
@aarmey Can you please help to review?

aarmey:
@HydrogenSulfate please first pull the changes from `⁣main` and resolve the conflicts. I removed some of these files from `doc/` that caused many linting changes.

aarmey (commented):
Here are a couple questions to start.

tensorly/tenalg/proximal.py
```diff
@@ -586,17 +586,16 @@ def unimodality_prox(tensor):
         tl.max(sum_inc + tl.flip(sum_dec, axis=0)),
     )
     min_indice = tl.argmin(tl.tensor(difference), axis=0)
-
     for i in range(len(min_indice)):
         tensor_unimodal = tl.index_update(
             tensor_unimodal,
             tl.index[: int(min_indice[i]), i],
-            monotone_increasing[: int(min_indice[i]), i],
+            monotone_increasing[: int(min_indice[i]), i].contiguous(),
```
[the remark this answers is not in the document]
HydrogenSulfate: Oh it's a typo, has been deleted.

tensorly/solvers/tests/test_proximal.py
```diff
@@ -25,6 +25,11 @@
     reason=f"Indexing with list not supported in TensorFlow",
 )
 
+skip_paddle = pytest.mark.skipif(
```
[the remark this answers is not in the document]
HydrogenSulfate: Deleted

tensorly/regression/tests/test_cp_plsr.py
```diff
@@ -16,6 +16,10 @@
     tl.get_backend() in ("tensorflow",),
     reason=f"Operation not supported in {tl.get_backend()}",
 )
+skip_if_paddle = pytest.mark.skipif(
+    tl.get_backend() == "paddle",
+    reason=f"Operation not supported in {tl.get_backend()}",
```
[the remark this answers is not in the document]
HydrogenSulfate: Since paddle do not support `Tensor_x[bool_mask] = Tensor_y`(assigning tensor to a mask slice tensor), I implemented it with combination of apis below:

https://github.com/tensorly/tensorly/pull/565/commits/67ac03a0b8ff61f8e3dab30409be47cf8b8d183b#diff-f495243586a2d2cfc90c4858babef0d80bb35207b402d9b91d9690cdb012da44R134-R145

HydrogenSulfate:
@aarmey I have update the code and deleted those skip_paddle flags.

aarmey:
@HydrogenSulfate please look at the diff associated with this PR. It looks like you are adding back the files in `doc/sphinx_ext/`, when these should remain deleted.

HydrogenSulfate:
> @HydrogenSulfate please look at the diff associated with this PR. It looks like you are adding back the files in `doc/sphinx_ext/`, when these should remain deleted.

Done, and remove the last skip_paddle in test_backend.

tensorly/tenalg/svd.py
```diff
@@ -228,7 +228,7 @@ def truncated_svd(matrix, n_eigenvecs=None, **kwargs):
     """
     n_eigenvecs, min_dim, _ = svd_checks(matrix, n_eigenvecs=n_eigenvecs)
     full_matrices = True if n_eigenvecs > min_dim else False
-
+    # print("matrix", matrix.shape, matrix)
```
aarmey, line 231: This comment can be removed.

tensorly/tests/test_backend.py
```diff
@@ -327,6 +403,13 @@ def test_clips_all_negative_tensor_correctly():
     assert tl.all(clipped_tensor == 0)
 
 
+def test_clips_all_negative_tensor_correctly_paddle():
```
aarmey, line 406: Could you please add a short docstring describing this test? Why does it have paddle in the name if it addresses a pytorch regression?

aarmey (changes requested):
This looks great! I just have a few minor edits here.

We should also enable testing for paddle if it is included as a backend. If you look in the project makefile, there is a `test-all` command that should have paddle added. `.github/workflows/test.yml` will also need to be modified to include paddle installation and testing. If you aren't sure how to make these changes, you can give me push access to your fork and I can add them.

After that, I'd like to have one other maintainer look this over, since we have to support all backends when we make most code changes.

tensorly/tenalg/svd.py
HydrogenSulfate: oops, this should be deleted. 👍

tensorly/tests/test_backend.py
HydrogenSulfate: Sorry for that I just copy and paste at that time, it is more suitable to add if-else for paddle in test_set_backend 👍

HydrogenSulfate:
> This looks great! I just have a few minor edits here.
> 
> We should also enable testing for paddle if it is included as a backend. If you look in the project makefile, there is a `test-all` command that should have paddle added. `.github/workflows/test.yml` will also need to be modified to include paddle installation and testing. If you aren't sure how to make these changes, you can give me push access to your fork and I can add them.
> 
> After that, I'd like to have one other maintainer look this over, since we have to support all backends when we make most code changes.

Thanks for your careful code review. I will fix these problem and commit later.

HydrogenSulfate:
> This looks great! I just have a few minor edits here.
> 
> We should also enable testing for paddle if it is included as a backend. If you look in the project makefile, there is a `test-all` command that should have paddle added. `.github/workflows/test.yml` will also need to be modified to include paddle installation and testing. If you aren't sure how to make these changes, you can give me push access to your fork and I can add them.
> 
> After that, I'd like to have one other maintainer look this over, since we have to support all backends when we make most code changes.

A question is that why not using pre-commit with black instead of execute black manually, which seems a common practice.

yngvem:
Thank you @HydrogenSulfate for this great PR! It's great to see that you want to add Paddle support for TensorLy :) Overall, it's a strong PR that adds a new backend and adds many small improvements (particularly `tl.minimum`, `abs` -> `tl.abs`)

I have a high-level question that I think @JeanKossaifi should chime in on. We've removed the MXNet backend and talked about removing the TensorFlow backend, and I'm wary of the possible increased maintainer work needed by adding official support for a new backend. How do you feel about adding a new backend, @JeanKossaifi?

I was not familiar with Paddle before this PR, but with a quick Google Scholar search it seems to be a fairly commonly used and supporting it might bring new users to TensorLy and tensor methods. This is great! I just want to be sure that we as a team are able to maintain this new feature -- especially considering that some of the tests had to be disabled. Would you, @HydrogenSulfate, for example be willing to help out in the future if we find issues with Paddle support?

There's also an alternative to adding official support for Paddle in TensorLy, where you, @HydrogenSulfate, has ownership of paddle support. If you create a separate package, TensorLy-Paddle, which adds the paddle backend when imported. We could of course add links to this in the readme and documentation to make sure that paddle support gets the attention it deserves from potential TensorLy users.

Technically, I think this should be relatively straightforward: Move the `tensorly.backend.paddle_backend` module to the new package and append `"paddle"` to `tl.backend.available_backend_names` when the paddle_backend is imported. In the CI-CD pipeline for TensorLy-Paddle, TensorLy could be installed from the main branch, thus testing TensorLy-Paddle against the latest TensorLy development version. The trigger could e.g. be a scheduled job that runs weekly. If we get our test suite more stable, then we could even have TensorLy-Paddle automatically create issues when the tests fail.

Finally, I'd like to reiterate that I'm not against adding the paddle backend, and I really appreciate the hard work you've done @HydrogenSulfate (and I definitely want to add your `tl.minimum` and `tl.abs` fixes no matter what!), I just want @JeanKossaifi to be involved in the discussion before we add a new backend.

HydrogenSulfate:
> Thank you @HydrogenSulfate for this great PR! It's great to see that you want to add Paddle support for TensorLy :) Overall, it's a strong PR that adds a new backend and adds many small improvements (particularly `tl.minimum`, `abs` -> `tl.abs`)
> 
> I have a high-level question that I think @JeanKossaifi should chime in on. We've removed the MXNet backend and talked about removing the TensorFlow backend, and I'm wary of the possible increased maintainer work needed by adding official support for a new backend. How do you feel about adding a new backend, @JeanKossaifi?
> 
> I was not familiar with Paddle before this PR, but with a quick Google Scholar search it seems to be a fairly commonly used and supporting it might bring new users to TensorLy and tensor methods. This is great! I just want to be sure that we as a team are able to maintain this new feature -- especially considering that some of the tests had to be disabled. Would you, @HydrogenSulfate, for example be willing to help out in the future if we find issues with Paddle support?
> 
> There's also an alternative to adding official support for Paddle in TensorLy, where you, @HydrogenSulfate, has ownership of paddle support. If you create a separate package, TensorLy-Paddle, which adds the paddle backend when imported. We could of course add links to this in the readme and documentation to make sure that paddle support gets the attention it deserves from potential TensorLy users.
> 
> Technically, I think this should be relatively straightforward: Move the `tensorly.backend.paddle_backend` module to the new package and append `"paddle"` to `tl.backend.available_backend_names` when the paddle_backend is imported. In the CI-CD pipeline for TensorLy-Paddle, TensorLy could be installed from the main branch, thus testing TensorLy-Paddle against the latest TensorLy development version. The trigger could e.g. be a scheduled job that runs weekly. If we get our test suite more stable, then we could even have TensorLy-Paddle automatically create issues when the tests fail.
> 
> Finally, I'd like to reiterate that I'm not against adding the paddle backend, and I really appreciate the hard work you've done @HydrogenSulfate (and I definitely want to add your `tl.minimum` and `tl.abs` fixes no matter what!), I just want @JeanKossaifi to be involved in the discussion before we add a new backend.

Many thanks for your review and patience. 

1. abs --> tl.abs: I found paddle do not support `__abs__` which is will be called by python builtin `abs()`, so I replace it with `tl.abs` since it only use `XXBackend.abs` which is more feasible. The important thing that we should and will support the magic method `__abs__` for `paddle.Tensor` in near future.
2. We are warmly welcome anyone raise problems or issues when using PaddlePaddle, that's one of the reason I made this pull request, **only more users can make the code more robust**. We can provide CI/CD resources to test dayily/weekly.
3. This should be a great development. I will discuss it with my colleagues as soon as possible. We might refer to a similar PyTorch process and add this as a separate installation package.
4. Yes, I also found it very strange at the time why there was a maximum but no minimum :laughing: .

HydrogenSulfate:
> There's also an alternative to adding official support for Paddle in TensorLy, where you, @HydrogenSulfate, has ownership of paddle support. If you create a separate package, TensorLy-Paddle, which adds the paddle backend when imported. We could of course add links to this in the readme and documentation to make sure that paddle support gets the attention it deserves from potential TensorLy users.
> 
> Technically, I think this should be relatively straightforward: Move the `tensorly.backend.paddle_backend` module to the new package and append `"paddle"` to `tl.backend.available_backend_names` when the paddle_backend is imported. In the CI-CD pipeline for TensorLy-Paddle, TensorLy could be installed from the main branch, thus testing TensorLy-Paddle against the latest TensorLy development version. The trigger could e.g. be a scheduled job that runs weekly. If we get our test suite more stable, then we could even have TensorLy-Paddle automatically create issues when the tests fail.

In fact, we more eagerly hope that the TensorLy library can directly support Paddle in its code. Currently, many of our users rely on TensorLy, but they have found that TensorLy does not support Paddle well. We hope to support Paddle in the TensorLy main branch through a PR, so they do not need to download additional packages and can seamlessly switch between PyTorch, JAX, and Paddle backends.

aarmey:
Thanks, @yngvem, for an excellent summary of the options here.

While I was initially hesitant to add another backend, our decision to remove MXNet and (maybe) TensorFlow has primarily been motivated by differences in behavior that then require workarounds. Paddle seems to compare well in this regard.

HydrogenSulfate:
> Hi @HydrogenSulfate, @aarmey and @yngvem, thank you all for the great discussion - it's great to come back from ICML to this PR and discussion very much on topic! :)
> 
> To summarize, our main reasons to deprecate MXNet was the lack of support, and for TensorFlow, both the decreasing support but also the large burden of supporting it due to very different API. It forced us to make some algorithms less efficient just to support it. As we are seeing a harmonization of API across array frameworks, on the other hand, things were getting easier for NumPy/JAX/PyTorch/CuPy.
> 
> I am not very familiar with Paddle @HydrogenSulfate, could you perhaps give us a little bit of context on expected support of the package going forward and advantages/reasons to use it over other frameworks?
> 
> In parallel, it makes me think that it may be useful to have somewhere some **backend-specific** examples, e.g. using PyTorch with TensorLy and leveraging autograd @aarmey @yngvem ? When building the doc, we could also loop through the available backends and have a short documentation for each?

Let me briefly introduce it: [PaddlePaddle](https://www.paddlepaddle.org.cn/en), similar to PyTorch, is a deep learning framework (both in terms of functionality and API design). From the user's perspective, they would use [dynamic graph mode for algorithm development](https://www.paddlepaddle.org.cn/documentation/docs/en/guides/beginner/tensor_en.html), and then utilize our provided tools for [model exporting](https://www.paddlepaddle.org.cn/documentation/docs/en/guides/infer/paddleslim/paddle_slim_en.html), model compression, and inference libraries. We have also deeply integrated with inference libraries like ONNX and TensorRT.

From a global framework perspective, we offer corresponding support tools at the chip layer, framework layer, model layer, and application layer. This includes multi-hardware adaptation (CPU, GPU, DCU, XPU), deep framework + AI compiler (PaddlePaddle + CINN compiler), model layer ([PaddleOCR](https://github.com/PaddlePaddle/PaddleOCR/tree/release/2.7), [PaddleDetection](https://github.com/PaddlePaddle/PaddleDetection)...), and application layer ([AIStudio](https://aistudio.baidu.com/projectdetail/8141482), [Fastdeploy](https://github.com/PaddlePaddle/FastDeploy)).

**The CINN compiler, mentioned earlier, is a newly added feature that we have been rapidly iterating on. It is similar to `torch.compile`, and we have achieved support for higher-order differentiation scenarios (as far as I know, even torch 2.4 does not yet support compiling for higher-order differentiation scenarios). On the [NVIDIA Modulus-sym](https://github.com/PaddlePaddle/modulus-sym) (a scientific computing development suite), we achieved a performance of 71% more instances per second (IPS) than PyTorch across 46 models in single A100 40GB GPU. Considering the extensive use of TensorLy in AI4S, we believe that the adaptation work for TensorLy is an important part of the framework's future development in the scientific computing field.**

image


Besides, we are glad to provide simple tutorial document about how to intergrate tensorly with PaddlePaddle if accepted as backend.

HydrogenSulfate:
@aarmey Hi, are there any further plans or progress regarding this PR?

tensorly/backend/paddle_backend.py
```diff
@@ -0,0 +1,366 @@
...
+    def eigh(tensor: paddle.Tensor):
+        return paddle.linalg.eigh(tensor)
+
+    @staticmethod
+    def sign(tensor: paddle.Tensor):
+        """paddle.sign does not support complex numbers."""
+        return paddle.sgn(tensor)
+
+    @staticmethod
+    def logsumexp(tensor: paddle.Tensor, axis=0):
```
aarmey, line 334: This function can just be imported (like the code below line 338), since the arguments aren't being changed.

tensorly/backend/paddle_backend.py
```diff
@@ -0,0 +1,366 @@
...
+            return paddle.flip(tensor, axis=[i for i in range(tensor.ndim)])
+        else:
+            return paddle.flip(tensor, axis=axis)
+
+    @staticmethod
+    def concatenate(tensors: paddle.Tensor, axis=0):
+        return paddle.concat(tensors, axis=axis)
+
+    @staticmethod
+    def argmin(input: paddle.Tensor, axis=None):
```
aarmey, line 267: `argmin`, `argmax`, `argsort`, and `stack` can just be imported using the code at the bottom, since these functions are not altering the arguments. These are handled specially in the torch backend because they use `dim` instead of `axis`.

tensorly/backend/paddle_backend.py
```diff
@@ -0,0 +1,366 @@
...
+        if not a.ndim or not b.ndim:
+            return a * b
+        return paddle.matmul(a, b)
+
+    @staticmethod
+    def tensordot(a: paddle.Tensor, b: paddle.Tensor, axes=2, **kwargs):
+        return paddle.tensordot(a, b, axes=axes, **kwargs)
+
+    @staticmethod
+    def mean(tensor: paddle.Tensor, axis=None):
```
aarmey, line 233: Can you confirm that special handling is needed for `mean`, `sum`, `max`, and `flip`? If their behavior matches NumPy, they can just be imported at the bottom.

tensorly/backend/paddle_backend.py
```diff
@@ -0,0 +1,366 @@
...
+            axes = list(axes)
+        return tensor.transpose(axes)
+
+    @staticmethod
+    def copy(tensor: paddle.Tensor):
+        return tensor.clone()
+
+    @staticmethod
+    def norm(tensor: paddle.Tensor, order=None, axis=None):
+        # paddle does not accept `None` for any keyword arguments. additionally,
```
aarmey, line 207: This comment is [not consistent with the paddle documentation](https://www.paddlepaddle.org.cn/documentation/docs/en/2.2/api/paddle/norm_en.html#norm). Can you confirm that `norm()` needs to be handled as a special case like this? It looks like you may be able to just import it at the bottom with `eigh()`, etc.

.github/workflows/test.yml
```diff
@@ -52,6 +52,10 @@ jobs:
         elif [[ "${{matrix.BACKEND}}" == "jax" ]]; then
           echo "Installing JAX";
           pip install jax jaxlib;
+        elif [[ "${{matrix.BACKEND}}" == "paddle" ]]; then
+          echo "Installing PaddlePaddle";
+          pip install --pre paddlepaddle -i https://www.paddlepaddle.org.cn/packages/nightly/cpu/;
```
aarmey, line 57: Would it be possible to test against a released version? This would help to ensure TensorLy works for users.

tensorly/backend/paddle_backend.py
```diff
@@ -0,0 +1,366 @@
...
+            sol = sol.squeeze(-1)
+        return sol, res, rank, single_value
+
+    @staticmethod
+    def eigh(tensor: paddle.Tensor):
+        return paddle.linalg.eigh(tensor)
+
+    @staticmethod
+    def sign(tensor: paddle.Tensor):
+        """paddle.sign does not support complex numbers."""
```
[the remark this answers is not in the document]
aarmey: Indeed, I'm not seeing [`paddle.sgn`](https://www.paddlepaddle.org.cn/en/searchall?q=paddle.sgn&language=en&version=2.2) in the documentation.

tensorly/backend/paddle_backend.py
```diff
@@ -0,0 +1,366 @@
...
+    @staticmethod
+    def shape(tensor: paddle.Tensor):
+        return tuple(tensor.shape)
+
+    @staticmethod
+    def ndim(tensor: paddle.Tensor):
+        return tensor.ndim
+
+    @staticmethod
+    def arange(start, stop=None, step=1.0, *args, **kwargs):
```
aarmey, line 173: Is this handled as a special case to avoid a typing error? From the documentation, it looks like you may be able to just import straight from paddle.

aarmey (changes requested):
Thank you for all the great information about paddle, @HydrogenSulfate! Bear with us, as we are just a small group of volunteers, so development of TensorLy is intermittent.

I have a few more questions after looking through the paddle documentation. It is helpful to us to keep the backend code as simple as possible, since it interacts with all the other code in sometimes complex ways.

tensorly/backend/paddle_backend.py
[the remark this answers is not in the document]
HydrogenSulfate: 1. The english document will be fixed as soon as possible, https://www.paddlepaddle.org.cn/documentation/docs/en/api/paddle/sgn_en.html
2. `paddle.sgn has` same behavior and degisning motivation as pytorch, which returns `x/abs(x)` rather than `-1/1`

tensorly/backend/paddle_backend.py
```diff
@@ -0,0 +1,366 @@
...
+                    compute_residuals = all([r.item() == n for r in rank])
+            if not compute_residuals:
+                res = paddle.empty([0], dtype=a.dtype)
+
+        if vector_case:
+            sol = sol.squeeze(-1)
+        return sol, res, rank, single_value
+
+    @staticmethod
+    def eigh(tensor: paddle.Tensor):
```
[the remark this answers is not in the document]
HydrogenSulfate: Removed.

tensorly/backend/paddle_backend.py
HydrogenSulfate: Simplified to register code

tensorly/backend/paddle_backend.py
HydrogenSulfate: Simplified to registering code

tensorly/backend/paddle_backend.py
HydrogenSulfate: You are right, only `order` be changed to float when order is `inf`/`-inf`.

tensorly/backend/paddle_backend.py
HydrogenSulfate: `mean`, `sum`, `max` can be removed to registering code, but `flip` is remained to handle `axis` when it is None.

.github/workflows/test.yml
HydrogenSulfate: Updated to `3.0.0b1`

tensorly/backend/paddle_backend.py
HydrogenSulfate: Simplified to registering code

aarmey (approved):
Great. @JeanKossaifi @yngvem, I'm in favor of merging as it looks like there will be minimal support overhead. Please let me know what you think.

I agree that better documentation would be helpful around integrating TensorLy with backend-specific features such as autodiff. In fact, I'm not sure what is the expected behavior. Should one be able to autodiff through each of the basic constructs such as CPTensor, as well as the metric functions? This could be helpful to test; some backends are finicky about which operations are allowed during tracing. We probably have some conversions to/from NumPy hiding around, too.

yngvem:
Considering @HydrogenSulfate's positive feedback to our questions, I'm also in favour of merging :)

Also 100% agree that we should take a stance regarding what operations are allowed during tracing. Ideally everything would be traceable, but that would make it impossible to e.g. use the SVD, so that's obviously out of the picture. My gut feeling is that fitting models doesn't need to be traceable, but assembling tensors from decompositions should be? Anyways, that's probably a discussion best had elsewhere!

Issue #547, opened 2024-03-18
Remove MXNet from doc


JeanKossaifi:
The website of tensorly still promises MxNet.
<img width="905" alt="Screenshot 2024-01-31 at 13 43 20" src="https://github.com/tensorly/tensorly/assets/68953985/29304ad6-6966-4741-b98a-3686541c8719">

_Originally posted by @bakhtos in https://github.com/tensorly/tensorly/issues/533#issuecomment-1918944246_

cohenjer:
Closing since this has been handled.

PR #567, opened 2024-07-25
Fix the documentation and remove last reference to mxnet

aarmey:
- The documentation is now built on each push, but only deployed on main.
- matplotlib was failing due to incompatibility with numpy 2.0
- Removed code that is already just installed from a package


aarmey:
This should fix #547.

doc/_static/tensorly-pyramid.png
JeanKossaifi, line 1: Not sure what happened here? Was the file changed at all?

.github/workflows/build_doc.yml
```diff
@@ -37,7 +34,7 @@ jobs:
           make html
           cd ..
       - name: Push doc to website
-        if: success()
+        if: success() && github.ref == 'refs/heads/main'
```
JeanKossaifi, line 37: This is a great change, thank you!

JeanKossaifi (commented):
This all looks great to me, thank you @aarmey !

doc/_static/tensorly-pyramid.png
aarmey: Yes, I removed mxnet.

PR #569, opened 2024-07-26
moving solvers related code to solvers/* directory, improving fista and nnls_hals (reopened)

cohenjer:
Reopening #550, I removed the tenalg/proximal.py which has now moved to solvers/*

You can see changes to proximal.py at this commit from #550  
https://github.com/tensorly/tensorly/pull/550/files#diff-1e631b99f3adf79834a04bcbc48dce54f1c7aa7f25f092853fe169a32bd8231d


JeanKossaifi:
Thanks @cohenjer, looks good to me!Tthe only thing is I wonder whether it would make sense to have (some of) the proximal operators in tonal to enable different implementations using either core operations or einsum when users switch tenalg.backend. What do you think @cohenjer @aarmey ?

cohenjer:
@JeanKossaifi My impression is that at the moment, the proximal operators we have implemented are designed to work on factor matrices rather than tensors, or on columns of tensors, therefore the contraction strategy is not an issue. This could change in the future.

For now, I would find it awkward to put them in tenalg when their implementation (at the moment) is tenalg-independent?

JeanKossaifi:
I see your point, in practice most of the tensor algorithms still use proximal operators on unfolding -- e.g. the [robust tensor PCA](https://github.com/tensorly/tensorly/blob/main/tensorly/decomposition/robust_decomposition.py) just maintains a list of Lagrangian for each unfolding.

I do have a preference to have tenalg encompass both matrix and tensors (e.g. I wouldn't split it into linalg and tenalg, matrices *are* second order tensors and it's ok if all our functions do not support any arbitrary order). 

By multiple implementations I was also thinking of a more general support of "backend", e.g. if some library supports efficient implementation of some of these, we could use that directly through the tenalg backend. Otherwise to accelerate any of them we'd have to define a third backend (currently we have the core backend and tenalg backend). Going forward I'd like to explore plugging in newer libraries, e.g. to try to JIT some of the functions or accelerate them. 

As a side note, by supporting einsum, my hope it not to just find better contraction path but also enable faster (ideally inplace) tensor contraction, e.g. through cuTensor and cuQuantum.

cohenjer:
The rationale for having a solvers/ directory is to have readily available a few algorithms that are often encountered in tensor/matrix factorization such as ADMM and Nonnegative Least Squares in a dedicated and easily identifiable location in the code. We could also decide that these algorithms may be provided by the backend (eg. numpy has an implementation of nnls which only accepts vectorized inputs) but as far as I am concerned, these solvers are an intermediate layer between decomposition algorithms like ALS and the backend and need to be stored in a dedicated folder. In this regard, prox operators are indeed closer to elementary operations found in the backend/ directory, but they are mostly used inside algorithms stored in the solvers/ directory or by higher level algorithms such as robust PCA.

@Jean I am only half-convinced that we will have different implementations of proximity operators depending on the back-end, but I might be wrong and there is no strong reason to put proximity operators in solvers/ rather than clarity. So at the end of the day I am happy to move them in tenalg if you think the trade-off is worth it.

Should I move all of them or only some? on what criteria should we decide which prox to put in tenalg? Splitting the proximal.py file in two would make it harder for new contributors to know where to implement proxs operators, so we could put all of them in tenalg.

cohenjer:
Any update on this ? The PR is almost ready to merge and I want to move forward with improving the solvers !

aarmey:
@cohenjer, for what it's worth, I favor proceeding with your current structure. @JeanKossaifi brings up exciting possibilities, like JIT and more efficient contractions, but right now, we aren't considering existing alternative implementations. It wouldn't be a major refactoring to move these to provide dispatching if one day we decide to introduce this. I'm happy to look over everything if you want a second pair of eyes before merging.

JeanKossaifi:
Apologies, the response got lost in all the Github notifications, I need to find a better way to filter those where I'm directly tagged. 

Ok, let's keep the current structure - for me the main point would be to have tenalg encompass both linear and tensor algebra: are you fine just leaving the proximal in tenalg? @aarmey @cohenjer

JeanKossaifi (approved):
Looks good to me, thanks a lot @cohenjer!

JeanKossaifi:
Just fixed the merge conflict, I'll wait for the CI

JeanKossaifi:
Looks good @cohenjer but it [breaks the doc](https://github.com/cohenjer/tensorly/actions/runs/11599394099/job/32297425700), I think a matter of updating it with the new organization.

cohenjer:
> Looks good @cohenjer but it [breaks the doc](https://github.com/cohenjer/tensorly/actions/runs/11599394099/job/32297425700), I think a matter of updating it with the new organization.

I forgot to report the rollback of proximal.py in tenalg in the api doc ! Will do asap.

cohenjer:
I fixed the api doc, also there is a case of tucker decomposition in the doc not covered in the tests which failed ^^

I think we could eventually:
- improve the documentation for proximal operators
- should have a better test for `nonnegative_tucker_hals` using active_set (or remove this option).

I can open two issues regarding this, but anyway we should be able to merge at this point, the doc built on my side :)

JeanKossaifi:
Super, merci beaucoup @cohenjer! Merging.

PR #572, opened 2024-08-05
Virtually all tests working across backends

aarmey:
It turns out that we weren't fully using the experimental NumPy interface within Tensorflow. With that enabled, and a couple other minor changes, virtually all the tests can be enabled across backends. As far as I can tell, the JAX and CuPy tests that were skipped already worked.

`test_tensor_train_cross_3()` still doesn't work, but I believe it is a problem with the method. A tensor is being indexed with an empty mask.

`active_set_nnls()` also still doesn't work. I think this is an issue with our approach for enabling item assignment. It looks like the `X.assign()` approach is deprecated, and we should be doing something with "gather" methods.

### DIFF
--- a/tensorly/__init__.py
+++ b/tensorly/__init__.py
@@ -126,6 +126,8 @@ from .backend import (
     pi,
     inf,
     nan,
+    int32,
+    int64,
 )
 
 from . import backend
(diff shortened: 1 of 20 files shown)


JeanKossaifi:
This is looking great to me @aarmey !

PR #533, opened 2023-10-19
Remove mxnet

aarmey:
This removes mxnet as a backend, since development has halted.

PR #573, opened 2024-09-04
[Paddle] Update index.html and pyramid png

HydrogenSulfate:
1. Add paddle url in index.html(It seems the source file of tensorly homepage)
2. Add Paddle logo in pyramid png, and remove mxnet logo(related to #547, @JeanKossaifi , #533 )

I found the homepage of tensorly is out of date and not consistent with index.html, will it be rebuild @aarmey ?


doc/index.html
```diff
@@ -1,34 +1,35 @@
...
+                <li> Open Source, BSD Licensed</li>
                 <li> Pure Python, Tested & Optimized </li>
                 <li> Flexible Backends for:
                         <a target="_blank" href="https://numpy.org/"> NumPy</a>,
                         <a target="_blank" href="https://pytorch.org/"> PyTorch</a>,
                         <a target="_blank" href="https://www.tensorflow.org/"> TensorFlow</a>,
                         <a target="_blank" href="https://jax.readthedocs.io/en/latest/"> JAX</a>,
-                        and <a target="_blank" href="https://cupy.dev/"> CuPy </a> </li>
+                        <a target="_blank" href="https://cupy.dev/"> CuPy </a>,
+                        and <a target="_blank" href="https://www.paddlepaddle.org.cn/en/install/quick?docurl=/documentation/docs/en/develop/install/pip/linux-pip_en.html"> PaddlePaddle</a> </li>
```
aarmey, line 32: Can we link to `https://www.paddlepaddle.org.cn` to be consistent with the other tools?

aarmey (changes requested):
Thanks, @HydrogenSulfate. @JeanKossaifi, I think I'll need your help to identify why the website is not updating.

JeanKossaifi:
Thanks @HydrogenSulfate,

 How did you make the jpg? If you worked directly on the existing image the result most likely would be lower res?
 If you link me to the Paddle logo, I can update the original pyramid.
 
 We probably should add a page to highlight the support of the various backends - what do you think @aarmey .
 
 Re. website, is it not building? It should be deployed each time there is a new push on main? 
 Are you maybe looking at the stable version? 
 The [dev version](http://tensorly.org/dev/modules/api.html) gets updated until we make a new release (we should probably make one soon).

doc/index.html
HydrogenSulfate: Done

HydrogenSulfate:
> Thanks @HydrogenSulfate,
> 
> How did you make the jpg? If you worked directly on the existing image the result most likely would be lower res? If you link me to the Paddle logo, I can update the original pyramid.
> 
> We probably should add a page to highlight the support of the various backends - what do you think @aarmey .
> 
> Re. website, is it not building? It should be deployed each time there is a new push on main? Are you maybe looking at the stable version? The [dev version](http://tensorly.org/dev/modules/api.html) gets updated until we make a new release (we should probably make one soon).

1. First I save the png in <https://tensorly.org/stable/index.html>, than I replace mxnet logo with a rectangle with same color as background and finally put Paddle logo. Here is the logo pic:
飞桨英文标识RGB, you can use this logo in your original pyramid, thanks.

2. I found the stable version of tensorly homepage is displayed in searching result, so MXNet is still here.
image
and I couldn't find a switch where I can change the homepage version fron stable to dev

aarmey:
@JeanKossaifi you're right, the dev version is updated from my earlier changes. It looks like we just need to issue a new version.

aarmey (approved):

aarmey:
Thanks @HydrogenSulfate!

PR #575, opened 2024-09-13
Add dummy keyword arguments to Backend svd to throw a more informative error

characat0:
In some cases, the Backend.svd method will be called (ie. when using SparseBackend that does not implement it).
In those cases, the keyword argument `full_matrices` is present, but not expected.

This will still throw an error, but the `NotImplementedError` error is more informative than 
```
TypeError: Backend.svd() got an unexpected keyword argument 'full_matrices'
```


JeanKossaifi:
That looks good to me, thanks @characat0!

PR #576, opened 2024-10-09
Upgrade Python version

aarmey:
This PR:

- Removes Python v3.8 from testing.
- Adds Python v3.13 to testing.
- Resolves pyupgrade checks for >= Python v3.9


JeanKossaifi:
Should we also stop testing on 3.9 as it's reaching EOL soon, so we can use the newer features (e.g. type hints)?

PR #583, opened 2024-12-07
[Paddle] Fix default device setting and allow lstsq fallback

HydrogenSulfate:
1. Switch the default device from CPU to current global device automatically detected by paddle
2. Allow lstsq fallback to CPU implementation when default device is not CPU, with propriate warning given, or the `test_tensor_ring_als` will fail when GPU is available.
3. Relax Paddle version check to allow nightly build version which a lot of paddle users has been installed

test result:

``` sh
TENSORLY_BACKEND='paddle' pytest -s -v -x tensorly
``` 

B70AAC32F9C348F366604742BF5525F9

### DIFF
--- a/tensorly/solvers/nnls.py
+++ b/tensorly/solvers/nnls.py
@@ -15,7 +15,7 @@ def hals_nnls(
     epsilon=0.0,
     callback=None,
 ):
-    """
+    r"""
     Non Negative Least Squares (NNLS)
 
     Computes an approximate solution of a nonnegative least
--- a/tensorly/solvers/nnls.py
+++ b/tensorly/solvers/nnls.py
@@ -15,7 +15,7 @@ def hals_nnls(
     epsilon=0.0,
     callback=None,
 ):
-    r"""
+    """
     Non Negative Least Squares (NNLS)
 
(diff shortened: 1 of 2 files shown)


tensorly/solvers/nnls.py
```diff
@@ -15,7 +15,7 @@ def hals_nnls(
     epsilon=0.0,
     callback=None,
 ):
-    """
+    r"""
```
yngvem, line 18: If we make this into a raw-string, then we also need to remove all the double backslashes in the docstring. Alternatively, we can set all single backslashes into double backslashes and skip the r.

Also, it's probably better to use an uppercase R instead of a lowercase one since the lowercase R will trigger regex code highlighting in some editors.

yngvem (changes requested):
Looks good to me, just a tiny comment regarding the docstring change.

tensorly/solvers/nnls.py
HydrogenSulfate: Thanks for comments, it would be better to remove the 'r'. 😄

yngvem (approved):
LGTM :)

Issue #585, opened 2024-12-12
Bug in ```_validate_contraction_modes```: modes seem to be wrongfully assigned.


ax-le:
#### Describe the bug
As far as I understand tensor contractions, it seems to me that ```modes1``` and ```modes2``` are wrongfully set in the function ```_validate_contraction_modes``` in [tensorly.tenalg.tenalg_utils.py](https://github.com/tensorly/tensorly/blob/main/tensorly/tenalg/tenalg_utils.py) when the number of ```modes``` is equal to 2.

#### Steps or Code to Reproduce

The bug appears in the following code block, when ```modes``` is of length 2 (e.g. modes = [0,1]).

```python
modes1, modes2 = modes
```

The code comes from line 22 of [tensorly.tenalg.tenalg_utils.py](https://github.com/tensorly/tensorly/blob/main/tensorly/tenalg/tenalg_utils.py).

#### Expected behavior

 ```modes1``` and ```modes2``` should both take the value of ```modes``` (in my example,  modes1 = [0,1] and modes2 = [0,1]).

#### Actual result

```modes1``` takes the first value and ```modes2``` the second one (in my example, modes1 = 0 and modes2 = 1).

### Potential solution

A potential solution should be easy to implement, by just deleting lines 21/22/23 (the try/except), and just letting lines 24 and 25 
```python
modes1 = modes
modes2 = modes
```

#### Versions
Linux-6.8.0-49-generic-x86_64-with-glibc2.39
Python 3.12.3 (main, Nov  6 2024, 18:32:19) [GCC 13.2.0]
NumPy 2.0.2
SciPy 1.14.1
TensorLy 0.9.0

Thanks in advance! :)

JeanKossaifi:
Hi @ax-le, thank you for the detailed context - I remember hesitating on this edge case when I wrote that function.

I chose the current behavior so that it defaults to a tensor contraction between mode1 and mode2: `tensordot(t1, t2, [m1, m2])` then contracts mode `m1` of `t1` with mode `m2` of tensor `t2`. Otherwise the syntax becomes very convoluted to do that :  `tensordot(t1, t2, [(m1, ), (m2, )])`

Either way if you got confused this clearly means the behavior should be better documented!

ax-le:
Hi @JeanKossaifi thanks for your answer!

I understand your viewpoint, which was indeed very different from mine. I understand that, as it is, my pull request prevents your behaviour from being easily implemented, and I will hence probably retract it.

[...]

PR #587, opened 2024-12-20
Fixing a bug (issue #585)

ax-le:
Hi,
It seems to me that the a bug was introduced in the function ```_validate_contraction_modes``` in [tensorly.tenalg.tenalg_utils.py](https://github.com/tensorly/tensorly/blob/main/tensorly/tenalg/tenalg_utils.py), see details in issue #585.
This pull request aims at fixing this bug, using the suggested trick in issue #585 (i.e. deleting the try/except which generated the bug).
Let me know if you have any comment.
Thanks!


ax-le:
Closed because it interferes with the desired behaviour of the check, see Issue #585.

PR #589, opened 2024-12-23
FIX partial_unfold + add test

### LINKED ISSUE #588: Dimensions in partial_tensor_to_vec are sometimes flipped
#### Describe the bug
The expected tensor dimensions in ``partial_tensor_to_vec`` appear to be flipped in some situations. I don't know if this is the expected result, but looking at the [API reference](https://tensorly.org/stable/modules/generated/tensorly.base.partial_tensor_to_vec.html#tensorly.base.partial_tensor_to_vec), I wasn't expecting that.  

For instance:
```python
import tensorly as tl
T = tl.randn((2, 3, 4, 5))

print(tl.partial_tensor_to_vec(T, skip_begin=0, skip_end=2).shape)
```
returns ```(6, 5, 4)``` while one could expect ```(6, 4, 5)``` (according to the documentation, the two last dimensions should be left untouched in this case. Is that the expected behavior?

### version
macOS-14.5-arm64-arm-64bit
Python 3.11.1 (main, Dec 21 2022, 18:33:23) [Clang 14.0.0 (clang-1400.0.29.202)]
NumPy 2.0.0
SciPy 1.14.1
TensorLy 0.9.0

JeanKossaifi:
Fixes  #588

### DIFF
--- a/tensorly/base.py
+++ b/tensorly/base.py
@@ -113,7 +113,7 @@ def partial_unfold(tensor, mode=0, skip_begin=1, skip_end=0, ravel_tensors=False
         new_shape = [tensor.shape[i] for i in range(skip_begin)] + new_shape
 
     if skip_end:
-        new_shape += [tensor.shape[-i] for i in range(1, 1 + skip_end)]
+        new_shape += [tensor.shape[-i] for i in range(skip_end, 0, -1)]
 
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
Fixes #588

PR #593, opened 2025-01-22
Allow tucker_to_tensor() to work with partial_tucker() outputs with the modes parameter

Mithrillion:
`tucker_to_tensor()` is already calling `multi_mode_dot()` for reassembling decomposition, and `multi_mode_dot()` can already take a `modes` parameter to act on partially decomposed tucker tensors. We only have to pass the `modes` value to `multi_mode_dot()`.

This will facilitate situations using a decomposition where the batch dimension is kept (e.g. multiple instances with tensor-valued features).


aarmey:
Thanks, @Mithrillion! This is great.

JeanKossaifi:
@Mithrillion @aarmey I was thinking more about this, that is why I left it in multi_mode_dot. As you mentioned, it's a single call either way. My worry with the current PR is that we break the API consistency: if we want to support this, we should support it in *all* the Tucker functions, or at least all the major ones: we should at least be able to go back and forth.

PR #595, opened 2025-01-27
Share reconstruction error between CP and NN CP to fix masking

### LINKED ISSUE #578: Inaccurate reconstruction error for PARAFAC with missing data
When running (non-negative) parafac on a tensor with missing data (i.e. using ```mask```), in the current version the reconstruction error at every iteration is calculated based on the norm of the original data tensor. However, the masked part of the tensor is imputated at every iteration, so the norm of the tensor is not constant. Hence the values of the reconstruction error are inaccurate in this case, and the error is not strictly decreasing at every iteration.

Recalculating the norm of the tensor at every iteration appears to restore the expected behavior.

aarmey:
Fixed #578

There was also an argument mismatch in the class interface, where `rank` was missing.

### DIFF
--- a/tensorly/decomposition/_nn_cp.py
+++ b/tensorly/decomposition/_nn_cp.py
@@ -1,7 +1,7 @@
 import warnings
 import tensorly as tl
 from ._base_decomposition import DecompositionMixin
-from ._cp import initialize_cp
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
This looks good to me, thanks @aarmey !

PR #596, opened 2025-02-26
Added support for factors-only initialization in tucker decomposition

sebulo:
## Motivation
The `initialize_tucker` function currently requires `init` to be `"svd"`, `"random"`, or a `(core, factors)` tuple. For warm-starting Tucker decomposition, users may only have factors without a core. This PR lets `init` accept a list of factors and recomputes the core internally.

## Changes
- In `initialize_tucker`:
  - Added support for `init` as a list of factors under `elif isinstance(init, (tuple, list)):`
  - Validates factor count and shapes, then recomputes core with `multi_mode_dot`
  - Updated error message to include the new option
  
- In `mode_dot`:
  - Added handling of Complex32 floats to support low-precision complex tensors.

### DIFF
--- a/tensorly/decomposition/_tucker.py
+++ b/tensorly/decomposition/_tucker.py
@@ -22,6 +22,13 @@ from ..tenalg.svd import svd_interface
 # License: BSD 3 clause
 
 
+def _is_tucker_factors_init(init):
+    return (
+        isinstance(init, Iterable)
+        and not isinstance(init, (str, bytes, tuple, TuckerTensor))
+    )
+
+
 def initialize_tucker(
     tensor,
     rank,
@@ -92,8 +99,34 @@ def initialize_tucker(
             for index, mode in enumerate(modes)
         ]
 
-    else:
+    elif isinstance(init, tuple):
+        if len(init) != 2 or not isinstance(init[1], Iterable):
+            raise ValueError(
+                "If `init` is a tuple, it must be of the form (core, factors) "
+                "with `factors` an iterable of factor matrices."
+            )
         (core, factors) = init
+    elif isinstance(init, TuckerTensor):
+        core, factors = init
+    elif _is_tucker_factors_init(init):
+        factors = list(init)
+        if len(factors) != len(modes):
+            raise ValueError(f"Expected {len(modes)} factors, got {len(factors)}")
+
+        for i, (factor, mode) in enumerate(zip(factors, modes)):
+            expected_shape = (tensor.shape[mode], rank[i])
+            if tl.shape(factor) != expected_shape:
+                raise ValueError(
+                    f"Factor {i} has shape {tl.shape(factor)} but expected "
+                    f"{expected_shape}"
+                )
+
+        core = multi_mode_dot(tensor, factors, modes=modes, transpose=True)
+    else:
+        raise ValueError(
+            "Got invalid `init`. Expected 'svd', 'random', a Tucker tensor, "
+            "a (core, factors) tuple, or an iterable of factors."
+        )
 
     if non_negative is True:
         factors = [tl.abs(f) for f in factors]
@@ -285,8 +318,18 @@ def tucker(
     """
     if fixed_factors:
         try:
-            (core, factors) = init
-        except:
+            if _is_tucker_factors_init(init):
+                core, factors = initialize_tucker(
+                    tensor,
+                    rank,
+                    modes=list(range(tl.ndim(tensor))),
+                    init=init,
+                    svd=svd,
+                    random_state=random_state,
+                )
+            else:
+                (core, factors) = init
+        except Exception:
             raise ValueError(
                 f'Got fixed_factor={fixed_factors} but no appropriate Tucker tensor was passed for "init".'
             )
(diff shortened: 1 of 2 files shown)


tensorly/decomposition/_tucker.py
```diff
@@ -59,6 +59,9 @@ def initialize_tucker(
               initialized core tensor
     factors : list of factors
     """
+    # Validate rank against tensor shape
+    rank = validate_tucker_rank(tensor.shape, rank)
```
JeanKossaifi, line 63: Is there a reason to add this? We already validate inside the main decomposition function, which would be called less frequently than this one. I think we can assume it is already validated here - and document that in the docstring.

tensorly/decomposition/_tucker.py
```diff
@@ -92,8 +95,24 @@ def initialize_tucker(
             for index, mode in enumerate(modes)
         ]
 
+    elif isinstance(init, (tuple, list)): # either a tuple (core, factors) or a list of factors
+        if isinstance(init, tuple) and len(init) == 2:
```
JeanKossaifi, line 99: We might want to use Iterable here instead of tuple to keep it consistent and more robust, same below instead of list.

JeanKossaifi: Also I think we may want to verify that init[1] is also Iterable.. It's verbose but init may be a weird 2-tuple from a partial-tucker

tensorly/decomposition/_tucker.py
```diff
@@ -285,6 +304,8 @@ def tucker(
     """
     if fixed_factors:
         try:
+            if isinstance(init, list):  # Handle factors-only init
```
JeanKossaifi, line 307: Same, let's use Iterable here

tensorly/decomposition/_tucker.py
```diff
@@ -285,6 +304,8 @@ def tucker(
...
+                init = initialize_tucker(tensor, rank, list(range(tensor.ndim)), random_state, init=init)
```
JeanKossaifi, line 308: no need to create intermediate init variable, we can directly unpack into (core, factors)

tensorly/tenalg/core_tenalg/n_mode_product.py
```diff
@@ -68,9 +69,31 @@ def mode_dot(tensor, matrix_or_vector, mode, transpose=False):
             f"Provided array of dimension {T.ndim(matrix_or_vector)} not in [1, 2]."
         )
 
-    res = T.dot(matrix_or_vector, unfold(tensor, mode))
+    unfolded = unfold(tensor, mode)
+    if tensor.dtype == torch.complex32 and not handle_complex_half_by_upcasting:
```
JeanKossaifi, line 73: Let's leave this for a separate PR, we need to think better on how to incorporate it cleanly. Also we should be using tl.dtype not directly torch.dtype since the latter will break on other backends.

JeanKossaifi (changes requested):
Looks good overall, thanks @sebulo, I left some comments in the code. Generally, it's good to group each big change in its own PR.

tensorly/decomposition/_tucker.py
sebulo: You are right. It is already validated earlier.

tensorly/decomposition/_tucker.py
sebulo: Thanks for the comments! I addressed these concerns in the newest commit: use Iterable, validate factors in tuple init.

tensorly/decomposition/_tucker.py
sebulo: Added in latest commit.

tensorly/decomposition/_tucker.py
sebulo: Addressed in latest commit.

tensorly/tenalg/core_tenalg/n_mode_product.py
cohenjer: I agree with @JeanKossaifi, the support of complex operations in Tensorly, while interesting, should be handled separately.

Also, I believe some backends like Pytorch have built-in support for matrix multiplication with complex numbers, that may be faster than doing the splitting of real and imaginary parts manually, e.g. see [here](https://docs.pytorch.org/docs/stable/complex_numbers.html)

cohenjer:
Thanks for the useful addition @sebulo! Can I also ask if you can add a small unit test for the initialization with only factor matrices? something like checking that the core computed is indeed the optimal one, and that Tucker runs when called after with this initialization (on very small examples).

JeanKossaifi:
@sebulo would be great if you could give this another pass whenever you have time!

sebulo:
Sorry for the long delay. I gave this another pass and pushed a cleaned update focused only on factors-only Tucker initialization.

The change allows init in Tucker decomposition to be provided as only the factor matrices, without an explicit core. In that case, the core is recomputed from the input tensor and the provided factors.

I also removed the unrelated mode_dot complex change from this PR and added tests for:
- checking that the core recomputed from the provided factors is the expected one
- checking that tucker(...) runs correctly when initialized with factors only

aarmey (approved):
Thanks, @sebulo! This is much cleaner and straightforward.

PR #599, opened 2025-04-15
Disambiguating the choices of modes in tensor contraction (Issue #585).

ax-le:
Hi, just adding a comment to try to disambiguate the choice of the modes in tensor contraction, when the number of modes is equal to 2 (see Issue #585 for discussion).


JeanKossaifi:
Thank you @ax-le, this is much clearer! Merging, the Black issue is unrelated, will fix separately.

PR #600, opened 2025-04-22
Escape characters compatible with python >= 3.12

acotino-ignitioncomputing:
Minor correction in docstrings: changed escape characters to prevent 'SyntaxError' in python versions >= 3.12


JeanKossaifi:
Thank you @acotino-ignitioncomputing, looks good to me. Merging since Black failure is unrelated, will fix separately.

PR #608, opened 2025-09-03
Phase multiplier of singular vectors fixed

### LINKED ISSUE #607: Tensor Train fails to approximate complex-valued tensors (error grows with rank)
#### Describe the bug
For a fixed complex-valued tensor, it was observed that the approximation error **increases** when the rank of the Tensor Train decomposition increases.  

The issue can be reproduced, for example, in Google Colab by installing NumPy ≥2.0 (or by using other backends e.g. torch). The following code prints the mean squared error of a complex-valued Tensor Train approximation for different decomposition ranks (from 1 up to 15):

```python
#!pip install numpy==1.26.0 
!pip install numpy==2.3.1
!pip install tensorly

import numpy as np
import tensorly as tl
print(np.__version__)
np.set_printoptions(precision=3, floatmode='fixed')
tl.set_backend('numpy')

tensor_shape = (32, 32, 32)
rank_range = list(range(1, 15))
rng = tl.check_random_state(None)
tensor = tl.tensor(rng.normal(size=tensor_shape)) + 1j*tl.tensor(rng.normal(size=tensor_shape))

mse = []
for r in rank_range:
    tensor_approx = tl.tt_to_tensor(
        tl.decomposition.tensor_train(tensor, rank = r)
        )
    mse.append(tl.mean(tl.abs(tensor_approx - tensor)**2))

print(np.array(mse))
```

#### Expected behavior
For a fixed complex-valued tensor, the reconstruction error is expected to **decrease** monotonically as the decomposition rank increases.
For numpy 1.26.0 backend the output is following:
```[1.995 1.979 1.955 1.928 1.894 1.857 1.818 1.773 1.725 1.675 1.621 1.562 1.502 1.439]```

#### Actual result
The approximation error **increases** when the rank of the Tensor Train decomposition increases.  
The output of provided code for numpy 2.3.1 backend: 
```[2.003 2.038 2.096 2.095 2.133 2.147 2.188 2.216 2.311 2.387 2.423 2.370 2.447 2.497]``` 

#### Versions
```Linux-6.1.123+-x86_64-with-glibc2.35
Python 3.12.11 (main, Jun  4 2025, 08:56:18) [GCC 11.4.0]
NumPy 2.3.1
SciPy 1.16.1
TensorLy 0.9.0
Pytorch 2.8.0+cu126
```

The issue does not appear when using the NumPy backend with version <2.0, but it does appear with NumPy ≥2.0 and with other backends (checked with PyTorch).

Albly:
Fixes #607 
# Introduction
Consider the singular value decomposition (SVD) of a complex-valued matrix $\mathbf{X} \in \mathbb{C}^{M \times N}$ truncated to rank $R$:
$$\mathbf{X} = \mathbf{U}\mathbf{S}\mathbf{V}^H, \quad \mathbf{U}\in\mathbb{C}^{M×R}, \mathbf{S}\in \mathbb{R}^{R×R}, \mathbf{V}^H \in \mathbb{C}^{R×N}$$
It is known that SVD is invariant under multiplication of each pair of singular vectors by a complex exponential $C_r = e^{j\phi(r)}$. In other words, for the $r$-th singular component we can replace: $\mathbf{u}_r \leftarrow  C_r \cdot \mathbf{u}_r$ and $\mathbf{v}_r^H \leftarrow C_r^* \cdot \mathbf{v}_r^H$ without affecting the reconstruction accuracy.

# Problem description
According to the documentation of the ```svd_flip``` function, its purpose is to eliminate the arbitrary phase of singular vectors by aligning them with the element of largest magnitude. In other words, the correction factor for each $r$ is chosen as:
$$C_r = \mathrm{sign}(z)^*,\quad z = \mathbf{u}_r(\mathrm{argmax}(|\mathbf{u}_r|))$$ 
Where $\mathrm{sign}(z) = z/|z|$.  
**But in current version the conjugate operation is missed**. 

# Why it worked for numpy<2.0
Before Numpy 2.0, the definition of the complex sign function was [different](https://numpy.org/doc/stable/reference/generated/numpy.sign.html#numpy.sign): $sign(z) = z/ \sqrt{zz}$ equvalent to (```sign(x.real) + 0j if x.real != 0 else sign(x.imag) + 0j```) which always evaluates to either $1$ or $-1$. Since both the left and right singular vectors were multiplied by the same $\pm 1$ factor, the overall decomposition remained consistent, and the missing conjugation had no visible effect.

### DIFF
--- a/tensorly/tenalg/tests/test_svd.py
+++ b/tensorly/tenalg/tests/test_svd.py
@@ -32,8 +32,11 @@ def test_svd_interface_shape(shape, rank, nn):
 
 @pytest.mark.parametrize("shape", [(10, 5), (10, 10), (5, 10)])
 @pytest.mark.parametrize("rank", [1, 3, 10])
-@pytest.mark.parametrize("is_complex", [False, True])
 @pytest.mark.parametrize("is_u_based_flip_sign", [False, True])
+@pytest.mark.parametrize(
+    "is_complex",
+    [False] if tl.get_backend() is "paddle" else [True, False],
+)  # due to unsupported svd of cplx input in v3.0
 def test_svd_interface_approx(shape, rank, is_complex, is_u_based_flip_sign):
     """Test that SVD interface can approximate input matrix"""
     tol = 1e-6


Albly:
Thank you @JeanKossaifi for the feedback!
Regarding the tests, I actually found two possible ways to improve them and prevent the reported problem:

- There was no check with complex-valued tensors in the `test_tt_decomposition.py` test file. After reviewing the project history, I decided to fix it in the same way as `test_cp.py` does. ([4e2ff5a](https://github.com/tensorly/tensorly/pull/608/commits/4e2ff5a2bc282db6dada166e1f43b3107ed811e1))

- The `svd_interface` did not have a check, but it’s mostly about verifying the `svd_flip` function, since the SVD itself should already be tested in each backend. So, I added one more test for checking the SVD approximation. This might be useful if new functions are added to `svd_interface`. ([ea28781](https://github.com/tensorly/tensorly/pull/608/commits/ea2878118a94aac59b37a25ae7b52af12bbae156))

Both tests fails if described bug exists.

JeanKossaifi:
Apologies for the delay in merging @Albly, thanks a lot for the great PR!

Issue #615, opened 2026-01-13
Random state not passed in randomized svd initialization case


MrBones1102:
#### Describe the bug

`initialize_cp()` in `_cp.py` does not pass `rng` to `svd_interface()` in the `init == "svd"` case, meaning there is no way to fix the random state when using the "randomized_svd" initialization method.

#### Steps or Code to Reproduce

I was using the parafac2 method but anything that calls `initialize_cp()` would be effected.
The lack of fixed state should be apparent by specifying `init="svd"`, `svd="randomized_svd"`, and `random_state=42` when calling the `parafac2()` method.

#### Expected behavior
I would expect resulting factor matrices to be equal when fixed state is specified

#### Actual result

Factor matrices experience run-to-run variation.

#### Versions

Linux-6.17.0-8-generic-x86_64-with-glibc2.42
Python 3.12.6 (main, Sep  9 2024, 22:11:19) [Clang 18.1.8 ]
NumPy 2.3.5
SciPy 1.16.3
TensorLy 0.9.0
CuPy 13.6.0

MrBones1102:
I appears there are quite a few places where `svd_interface()` is called throughout the factorization pipeline and as someone new to this codebase, I may not be the best person to boldly modify every `svd_interface()` call, despite my initial enthusiasm when I opened #616

PR #616, opened 2026-01-13
add random_state to svd_interface()

MrBones1102:
From #615. Pass `rng` to new `random_state` in `svd_interface()` for use as kwargs in subsequent randomized svd initialization

### DIFF
--- a/tensorly/decomposition/_constrained_cp.py
+++ b/tensorly/decomposition/_constrained_cp.py
@@ -105,7 +105,7 @@ def initialize_constrained_parafac(
     elif init == "svd":
         factors = []
         for mode in range(tl.ndim(tensor)):
-            U, S, _ = svd_interface(unfold(tensor, mode), n_eigenvecs=rank, method=svd)
+            U, S, _ = svd_interface(unfold(tensor, mode), n_eigenvecs=rank, method=svd, random_state=rng)
 
             # Put SVD initialization on the same scaling as the tensor in case normalize_factors=False
             if mode == 0:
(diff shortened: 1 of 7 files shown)


MrBones1102:
Having had a look around some more, there are a few more instances of svd_interface() that require passing random_state

aarmey (changes requested):
Thanks @MrBones1102! Could you please fix those other uses of `svd_interface()`? They should pass on `random_state` for the same reason as here.

If a function calls `svd_interface()`, but does not have a random state, note it and I can take a look.

MrBones1102:
Added all other instances where `svd_interface()` is called, including situations where a `random_state` needed to be specified. I added `random_state` to any enclosing function that also takes the `svd` argument and which the documentation claims can have any value present in `tensorly.SVD_FUNS`, which includes `randomized_svd` potentially requiring `random_state`.

As a result, the following enclosing functions were modified to take `random_state` with default value set to `None`:
 - `_compute_projections()` in `_parafac2.py`
 - `tensor_ring()` in `_tr_svd.py`
 - `tensor_train()` in `_tt.py`
 - `svd_compress_tensor_slices()` in `preprocessing.py`

Additionally, I modified the `svd_interface()` calls in `test_backend.py` as one of those tests explicitly checks for `svd=="randomized_svd"` without passing a `random_state`, but perhaps this was intentional?

aarmey:
Take a look at the test errors here. I'm happy to discuss them if it's not clear what broke.

PR #617, opened 2026-03-05
PARAFAC2: hybridize SVD init for high rank and add warning/tests

MiklosKralik:
I've run into a use-case of PARAFAC2 where I would like to initialize the decomposition with svd features (for a sensible initialization) but would like to allow for the decomposition to fit a higher rank than the dimension of the data. I've added some logic to concatenate svd initialized rankd with randomly initialized ranks for this use case. 

I've added a UserWarning as this use case is rare and the user should be intentional when specifying a high rank. I've also added a unittest of rebroducibility for this init mode.


aarmey:
Closing and reopening to trigger tests.

aarmey (approved):
Thanks, @MiklosKralik! This is an issue I've had before, too.

